PR #2, opened 2021-06-24
build: enable building & running tests from a subdir

xnox:
During dpkg build, in a subdir, it is currently not possible to run
tests. Building testsuite/modules due to non-existance of the
testsuite directory under the build dir. Thus create it, when it is
not there.

Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@canonical.com>


lucasdemarchi:
Applied, thanks. Sorry for the delay.

PR #6, opened 2021-09-30
libkmod: fix CHECK_ERR_AND_FINISH

wangyuntao:


marcosfrm:
The macro is gone since 82972710196442c503cc1068a05651c354a865fb

PR #9, opened 2021-11-19
modprobe: Write error messages to syslog if stderr is unavailable

pqarmitage:
The man page modprobe(8) states for the --syslog option:
"This is also automatically enabled when stderr is unavailable."
but it wasn't happening.

This commit now makes modprobe write to syslog if stderr is closed.


lucasdemarchi:
Applied, thanks. Sorry for the delay.

PR #19, opened 2022-09-25
libkmod: do not crash on unknown signature algorithm

mikhailnov:
Example kernel module:

https://file-store.rosalinux.ru/download/7281f97e0c04c0f818ad3f936706f4a407e8dc7e
(/lib/modules/5.15.67-generic-1rosa2021.1-x86_64/kernel/drivers/usb/host/xhci-pci.ko.zst)
It is signed with Streebog 512.

libkmod v30 crashed in libkmod-module.c:2413 in this code:

```c
n = kmod_module_info_append(list,
	"sig_hashalgo", strlen("sig_hashalgo"),
	sig_info.hash_algo, strlen(sig_info.hash_algo));
```

because strlen() got null.

### DIFF
--- a/libkmod/libkmod-signature.c
+++ b/libkmod/libkmod-signature.c
@@ -278,6 +278,9 @@ static bool fill_pkcs7(const char *mem, off_t size,
 	X509_ALGOR_get0(&o, NULL, NULL, dig_alg);
 
 	sig_info->hash_algo = pkey_hash_algo[obj_to_hash_algo(o)];
+	// hash algo has not been recognized
+	if (sig_info->hash_algo == NULL)
+		goto err3;
 	sig_info->id_type = pkey_id_type[modsig->id_type];
 
 	pvt = malloc(sizeof(*pvt));


libkmod/libkmod-signature.c
```diff
@@ -278,6 +278,9 @@ static bool fill_pkcs7(const char *mem, off_t size,
 	X509_ALGOR_get0(&o, NULL, NULL, dig_alg);
 
 	sig_info->hash_algo = pkey_hash_algo[obj_to_hash_algo(o)];
+	// hash algo has not been recognized
+	if (sig_info->hash_algo == NULL)
```
lucasdemarchi, line 282: It looks like we should rather control the accesses to pkey_hash_algo[]. We don't have holes pkey_hash_algo, so this should never return NULL - it will return garbage if obj_to_hash_algo(o) returns something outside the array range.
In fact, obj_to_hash_algo(o) returns -1 and we shouldn't just pass that to the array access.

mikhailnov:
I have not understood well what you mean.
It was NULL with the exanple module from the patch description. Such check helped to avoid the crash.
Maybe it may be fixed better, I made just an obvious check/fix.

lucasdemarchi:
Yeah... it's just reading random memory that may or may not be zero. Proper fix would be to check
obj_to_hash_algo(o) >= 0

evelikov:
Sent a proper fix to the ML https://lore.kernel.org/linux-modules/20230206143259.585180-1-emil.l.velikov@gmail.com/T/#u

evelikov:
The original commit in this PR has landed a while back. The extra validation was merged with https://github.com/kmod-project/kmod/commit/b9605c63b859adfffc0b4b9420d720aa323b90e9

Seems like this PR can be closed.

PR #21, opened 2022-10-30
libkmod: use memory-mapped access to modules.builtin.modinfo

o2cheretnyi:
Sometimes the read access to huge modules.builtin.modinfo file via tons of pread() calls can cause speed degradation during 'depmod' command's execution, so the accessing of memory-mapped file's area is faster.

Signed-off-by: Oleksandr Ocheretnyi <oocheret@cisco.com>

### DIFF
--- a/libkmod/libkmod-builtin.c
+++ b/libkmod/libkmod-builtin.c
@@ -19,6 +19,7 @@
 
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <sys/mman.h>
 
 #include <unistd.h>
 #include <stdio.h>
@@ -36,6 +37,9 @@ struct kmod_builtin_iter {
 
 	// The file descriptor.
 	int file;
+
+	// The memory-mapped file address.
+	char *mmf_addr;
 
 	// The total size in bytes.
 	ssize_t size;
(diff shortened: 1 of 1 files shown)


evelikov:
@o2cheretnyi some notes from a first-time contributor to kmod:

Have you measured the impact of using mmap? Quantifying with concrete numbers in the commit message is always appreciated.

Would it make sense to `flock` for the duration of the mmap? Some projects do so, although I'm not sure if it'll be useful here.

Last but not least - consider sending the patch to the ML listed in the README. HTH o/

o2cheretnyi:
@evelikov , I have some comparison table based on Yocto's core-image-full-cmdline image run on QEMU. However I need to note that all measurements were done with kmod's version 29 where I'd observed this drawback first time.
 ```
'depmod' execution time during Yocto's core-image-full-cmdline image run on QEMU
+-----------------------------+-------------------------------+-------------------------------+
 |                             |          pread() calls        |          mmap() call          |
+-----------------------------+-------------------------------+-------------------------------+
 |        time, seconds        |         ~ 45-120              |            ~ 1-7              |
+-----------------------------+-------------------------------+-------------------------------+
```
Conclusion: these changes do not play much role when we are using kmod with version 30 what resolves this issue with https://github.com/kmod-project/kmod/commit/0246e06340df292b5dda4bc00e24cc9ae894e881

lucasdemarchi:
depmod 30 has changes explicitly to handle this issue. See https://github.com/kmod-project/kmod/blob/master/NEWS#L8-L11

could you check if upgrading to kmod 30 helps?

lucasdemarchi:
Oh... missed your last phrase in the comment. Yeah, I don't think we really need to move to mmap as just avoiding the open/read/close multiple times is sufficient.

PR #25, opened 2023-03-02
configure.ac: fix link with -llzma

ffontaine:
Add `liblzma_LIBS` to `LIBS` to avoid the following build failure when building with a static-only `liblzma.a`:

```
/home/autobuild/autobuild/instance-5/output-1/host/lib/gcc/xtensa-buildroot-linux-uclibc/11.3.0/../../../../xtensa-buildroot-linux-uclibc/bin/ld: libkmod/.libs/libkmod-internal.a(libkmod-file.o):(.text.xz_uncompress+0x10): undefined reference to `lzma_code'
```

Fixes:
 - http://autobuild.buildroot.org/results/83a4a7ecc77f39639d3e5bc8554bd01a62a3ede0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,7 +88,7 @@ AC_ARG_WITH([zstd],
 	AS_HELP_STRING([--with-zstd], [handle Zstandard-compressed modules @<:@default=disabled@:>@]),
 	[], [with_zstd=no])
 AS_IF([test "x$with_zstd" != "xno"], [
-	PKG_CHECK_MODULES([libzstd], [libzstd >= 1.4.4])
+	PKG_CHECK_MODULES([libzstd], [libzstd >= 1.4.4], [LIBS="$LIBS $libzstd_LIBS"])
 	AC_DEFINE([ENABLE_ZSTD], [1], [Enable Zstandard for modules.])
 ], [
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Do you have any additional downstream patch? Currently static builds are disabled:
https://github.com/kmod-project/kmod/pull/25/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810L21

ffontaine:
Even if static build is disabled with kmod, it doesn't mean that all reverse dependencies of kmod such as xz/lzma will provide a shared version of their library. This build failure is raised when building with buildroot in the `BR2_SHARED_STATIC `configuration. In this configuration, we're telling kmod to build dynamically, however the other packages are free to provide shared and/or static version of their library (i.e. --enable-static and --enable-shared are passed). In this case, xz is providing a static-only version of its library.

lucasdemarchi:
Right. I think it's not a common thing to do. It looks like we are missing a similar change on the other libs too. Can we have the same fix you did extended to the others?

ffontaine:
Sure, I updated the PR.

lucasdemarchi:
Appiled, thanks

Issue #29, opened 2023-08-23
cannot compile


Qwer-TeX:
First used `./autogen.sh`

Then Configured with:
``
./configure CFLAGS='-g -O2' --with-zlib --with-zstd --with-xz --prefix=/usr --libdir=/lib --sysconfdir=/etcc
``
Then:
`
make -j2
`

And I get this error
```
libkmod/libkmod-module.c: In function ‘do_finit_module’:
libkmod/libkmod-module.c:884:33: error: ‘MODULE_INIT_COMPRESSED_FILE’ undeclared (first use in this function)
  884 |                 kernel_flags |= MODULE_INIT_COMPRESSED_FILE;
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~
libkmod/libkmod-module.c:884:33: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [Makefile:1765: libkmod/libkmod-module.lo] Error 1
make[1]: *** [Makefile:2179: all-recursive] Error 1
make: *** [Makefile:1224: all] Error 2
```

PR #30, opened 2023-09-27
libkmod: add fallback MODULE_INIT_COMPRESSED_FILE define

evelikov:
The symbol was somewhat recently introduced by the kernel and not all distributions may be have available.

The number is part of the ABI, so we can add a local fallback define.

Closes: https://github.com/kmod-project/kmod/issues/29

@lucasdemarchi not sure if you're keeping an eye here or I should send it over to the ML.

### DIFF
--- a/shared/missing.h
+++ b/shared/missing.h
@@ -15,6 +15,10 @@
 # define MODULE_INIT_IGNORE_VERMAGIC 2
 #endif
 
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
LGTM. I will merge it later today

lucasdemarchi:
Applied, thanks

PR #32, opened 2023-12-04
Use portable implementation for basename API

kraj:
musl has removed the non-prototype declaration of basename from string.h [1] which now results in build errors with clang-17+ compiler

include libgen.h for using the posix declaration of the funciton.

Fixes
../git/tools/kmod.c:71:19: error: call to undeclared function 'basename'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   71 |                         "Commands:\n", basename(argv[0]));
      |                                        ^

[1] https://git.musl-libc.org/cgit/musl/commit/?id=725e17ed6dff4d0cd22487bb64470881e86a92e7
Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -794,7 +794,7 @@ static int conf_files_insert_sorted(struct kmod_ctx *ctx,
 	bool is_single = false;
 
 	if (name == NULL) {
-		name = basename(path);
+		name = gnu_basename(path);
 		is_single = true;
 	}
 
--- a/shared/util.c
+++ b/shared/util.c
@@ -172,9 +172,9 @@ char *modname_normalize(const char *modname, char buf[static PATH_MAX], size_t *
 
 char *path_to_modname(const char *path, char buf[static PATH_MAX], size_t *len)
 {
-	char *modname;
+	const char *modname;
 
-	modname = basename(path);
+	modname = gnu_basename(path);
 	if (modname == NULL || modname[0] == '\0')
 		return NULL;
 
--- a/shared/util.h
+++ b/shared/util.h
@@ -5,6 +5,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <string.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <time.h>
@@ -75,6 +76,12 @@ do {						\
 	} *__p = (typeof(__p)) (ptr);		\
 	__p->__v = (val);			\
 } while(0)
+
+static _always_inline_ const char *gnu_basename(const char *s)
+{
+  const char *p = strrchr(s, '/');
+  return p ? p+1 : s;
+}
 
 static _always_inline_ unsigned int ALIGN_POWER2(unsigned int u)
 {
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -70,7 +70,7 @@ static void help(void)
 
 	printf("Usage:\n"
 	       "\t%s [options] <test>\n"
-	       "Options:\n", basename(progname));
+	       "Options:\n", gnu_basename(progname));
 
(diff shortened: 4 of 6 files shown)


kraj:
> But they are not the same. See VERSIONS @ https://man7.org/linux/man-pages/man3/basename.3.html
> 
> We pass in a const pointer and expect it not to be modified by basename()

ok, I thought it was not the case, however if we want to emulate the glibc behaviour we will need to create a temporary string and use it as input to basename(). This will let it use posix basename. Will that be ok ?

richfelker:
If the GNU basename behavior is really what's wanted, that can be achieved with basically a one-line function calling `strrchr`, something like:

```
static const char *gnu_basename(const char *s) {
    const char *p = strrchr(s, '/');
    return p ? p+1 : s;
}
```

lucasdemarchi:
Thanks @richfelker. Yeah, using that as alternative sounds good to me.

libkmod/libkmod-config.c
```diff
@@ -794,7 +794,7 @@ static int conf_files_insert_sorted(struct kmod_ctx *ctx,
 	bool is_single = false;
 
 	if (name == NULL) {
-		name = basename(path);
```
lucasdemarchi, line 797: general approach in kmod is to consider the gnu-behavior as the desired one and add missing stuff for other libc's where appropriate. That way we don't have to keep patching kmod when the next .c uses basename() rather than gnu_basename().

skarnet: I understand that the GNU basename() behaviour is more desirable (and agree that it is), but the unfortunate fact is that the POSIX behaviour exists and is not going away; it's a bad move on GNU's part to have defined a symbol with a different prototype and a different behaviour from the POSIX one.

If I don't want the POSIX behaviour, I don't use `basename()`. I think that every project should do the same. I suggest that you go with `gnu_basename()`, and keep patching kmod because using `basename()` is suboptimal instead of keeping patching kmod because it breaks on some systems.

lucasdemarchi: My proposal is we do this: https://github.com/kmod-project/kmod/commits/tip-basename/

877280d fixup! fixup! Use portable implementation for basename API
3325c9d fixup! Use portable implementation for basename API
81580c1 Use portable implementation for basename API

I think we can just squash my commits onto yours if you agree.

lucasdemarchi: I should probably split the missing.h, but that can be done on top.

skarnet: That will work as long as the project *never* includes `libgen.h`. Which, true, it doesn't have a reason to.

kdave:
I fixed the basename semantics in btrfs-progs and got pointed to this PR for a fix. You may need to add handling of "/" as special case, this is mentioned in the manual page that basename is supposed to return that as "/", i.e. the strrchr() would return empty string.

lucasdemarchi:
@kdave  thanks for reaching out. However this is not what I get. :

```
# glibc
/foo/bar/: 
/foo/bar: bar
/foo: foo
/: 
:

# musl  with special / handling
/foo/bar/: 
/foo/bar: bar
/foo: foo
/: 
:
```

Checking the man page, it explicitly calls out that in case of '/' it returns the empty string. Emphasis mine:

> The GNU version never modifies its argument, and returns the **empty string** when path has a trailing slash, and in particular **also when it is "/"**.

libkmod/libkmod-config.c
kraj: @lucasdemarchi are you happy with this patch or do you have comments that should be addressed.

lucasdemarchi: Yep, applying it now. Thanks

lucasdemarchi:
Applied with the fixups.

PR #34, opened 2024-02-27
libkmod: add user soft dependecies

jtornosm:
It has been seen that for some network mac drivers (i.e. lan78xx) the related module for the phy is loaded dynamically depending on the current hardware. In this case, the associated phy is read using mdio bus and then the associated phy module is loaded during runtime (kernel function phy_request_driver_module). However, no software dependency is defined, so the user tools will no be able to get this dependency. For example, if dracut is used and the hardware is present, lan78xx will be included but no phy module will be added, and in the next restart the device will not work from boot because no related phy will be found during initramfs stage.

In order to solve this, we could define a normal 'pre' software dependency in lan78xx module with all the possible phy modules (there may be some), but proceeding in that way, all the possible phy modules would be loaded while only one is necessary.

The idea is to add a new attribute when the software dependency is defined, apart from the normal ones 'pre' and 'post', I have called it 'user', to be used only by the user tools that need to detect this situation. In that way, for example, dracut could check the 'user' attribute of the modules in order to install these software dependencies in initramfs too. That is, for the  commented lan78xx module, defining the 'user' attribute to the software dependency with the possible phy modules list, only the necessary phy would be loaded on demand keeping the same behavior but all the possible phy modules would be available from initramfs.

A new function 'kmod_module_get_user_softdeps' in libkmod will be added for this to avoid breaking the API and maintain backward compatibility. This general procedure could be useful for other similar cases (not only for dynamic phy loading).

### DIFF
--- a/libkmod/docs/libkmod-sections.txt
+++ b/libkmod/docs/libkmod-sections.txt
@@ -62,6 +62,7 @@ kmod_module_remove_module
 kmod_module_get_module
 kmod_module_get_dependencies
 kmod_module_get_softdeps
+kmod_module_get_user_softdeps
 kmod_module_apply_filter
 kmod_module_get_filtered_blacklist
 kmod_module_get_install_commands
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -54,8 +54,10 @@ struct kmod_softdep {
 	char *name;
 	const char **pre;
 	const char **post;
+	const char **user;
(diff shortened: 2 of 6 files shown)


jtornosm:
Also sent to kernel-modules mailing list:
https://lore.kernel.org/linux-modules/20240318161517.58550-1-jtornosm@redhat.com/T/#u

lucasdemarchi:
review being handled through mailing list.

PR #36, opened 2024-03-08
Install kmod.pc in ${datadir}/pkgconfig

floppym:
The data in this file isn't related to installed libraries, so put it in an abi-neutral location.

pkg.m4 provides macros that also allow the user to override the location with configure switches.

Bug: https://bugs.gentoo.org/926431


lucasdemarchi:
Applied, thanks

PR #42, opened 2024-06-07
lsmod, libkmod: handle module changes during execution

### LINKED ISSUE #41: Module may be unloaded before checking holders
The module may be loaded or unloaded between the read of `/proc/modules` and `/sys/module/*/holders`.

```
sneki@openstack-blade:~$ sudo lsmod | grep nvidia
sneki@openstack-blade:~$ sudo lsmod | grep nvidia
libkmod: ERROR ../libkmod/libkmod-module.c:1941 kmod_module_get_holders: could not open '/sys/module/nvidia/holders': No such file or directory
nvidia                    -2  -2
drm                   765952  8 drm_kms_helper,drm_display_helper,nvidia,drm_buddy,i915,ttm
sneki@openstack-blade:~$ sudo lsmod | grep nvidia
sneki@openstack-blade:~$ sudo lsmod | grep nvidia
libkmod: ERROR ../libkmod/libkmod-module.c:1941 kmod_module_get_holders: could not open '/sys/module/nvidia/holders': No such file or directory
nvidia                    -2  -2
```

Not sure if the kernel provides a way to inspect atomically, but `lsmod` should not print out inconsistent states in any case.

SuibianP:
Kernel modules may be unloaded within the time window of checking `/proc/modules` and `/sys/module/*/*`. As a result,

* `-errno` is printed in place of size or reference count
* `EXIT_SUCCESS` is reported despite failure

Add checks for better error reporting.

Fixes #41.

### DIFF
--- a/tools/lsmod.c
+++ b/tools/lsmod.c
@@ -61,11 +61,19 @@ static int do_lsmod(int argc, char *argv[])
 		const char *name = kmod_module_get_name(mod);
 		int use_count = kmod_module_get_refcnt(mod);
 		long size = kmod_module_get_size(mod);
-		struct kmod_list *holders, *hitr;
+		struct kmod_list *holders = kmod_module_get_holders(mod), *hitr;
 		int first = 1;
 
+		if (holders == NULL || use_count < 0 || size < 0) {
+			fprintf(stderr, "Error: could not query module %s! Was the module unloaded?\n", name);
+			kmod_module_unref_list(holders);
+			kmod_module_unref(mod);
+			kmod_module_unref_list(list);
+			kmod_unref(ctx);
+			return EXIT_FAILURE;
+		}
+
 		printf("%-19s %8ld  %d", name, size, use_count);
-		holders = kmod_module_get_holders(mod);
 		kmod_list_foreach(hitr, holders) {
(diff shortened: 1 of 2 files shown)


tools/lsmod.c
```diff
@@ -61,11 +61,19 @@ static int do_lsmod(int argc, char *argv[])
 		const char *name = kmod_module_get_name(mod);
 		int use_count = kmod_module_get_refcnt(mod);
 		long size = kmod_module_get_size(mod);
-		struct kmod_list *holders, *hitr;
+		struct kmod_list *holders = kmod_module_get_holders(mod), *hitr;
 		int first = 1;
 
+		if (holders == NULL || use_count < 0 || size < 0) {
```
lucasdemarchi, line 67: holders == NULL  is a normal thing to have, e.g:

```
serio_raw              20480  0
```

`use_count < 0` allows us to use modules in the "going" state, i.e. in the process of being unloaded.

`size < 0` I believe is the only one that we should really do something.... maybe setting it to 0.  Rather than exiting here, I think we should just handle the races as a normal situation and continue printing the rest of the list.

lucasdemarchi: @SuibianP can you rebase taking this into account?

PR #43, opened 2024-06-17
libkmod: improve realloc behavior for zstd outbuffer

q66:
This more than doubles performance of depmod (and likely modinfo etc.) on my machine.

### DIFF
--- a/libkmod/libkmod-file.c
+++ b/libkmod/libkmod-file.c
@@ -89,7 +89,11 @@ static int zstd_ensure_outbuffer_space(ZSTD_outBuffer *buffer, size_t min_free)
 	if (buffer->size - buffer->pos >= min_free)
 		return 0;
 
-	buffer->size += min_free;
+	if (buffer->size < min_free)
+		buffer->size = min_free;
+	else
+		buffer->size *= 2;
+
 	buffer->dst = realloc(buffer->dst, buffer->size);
 	if (buffer->dst == NULL) {
 		ret = -errno;


lucasdemarchi:
For a modinfo bench I didn't have such a big diff:


```
# Before:
✔ ~/p/kmod/build $ hyperfine -w 1 -r 5 ./bench.sh
Benchmark 1: ./bench.sh
  Time (mean ± σ):     13.838 s ±  2.058 s    [User: 7.485 s, System: 5.642 s]
  Range (min … max):   12.329 s … 17.353 s    5 runs
 
# After:
✔ ~/p/kmod/build $ hyperfine -w 1 -r 5 ./bench.sh
Benchmark 1: ./bench.sh
  Time (mean ± σ):     12.890 s ±  2.374 s    [User: 7.418 s, System: 4.679 s]
  Range (min … max):   11.214 s … 17.065 s    5 runs

✔ ~/p/kmod/build $ cat bench.sh 
#!/bin/bash
set -e

while read f; do ./tools/modinfo $f > /dev/null; done < <(find /lib/modules/$(uname -r) -type f -name *.zst)
```

For depmod I've got 1.031 s ±  0.021 s vs 953.1 ms ±  27.5 ms....

Definitely an improvement, but for me it's more in the ~7% ball park for a real workload.

q66:
That's due to a specific behavior of the glibc allocator (where directly subsequent reallocs with no allocations inbetween won't always actually allocate, however this comes at a cost elsewhere), but on libcs other than glibc (or with different allocators) it's going to make a much bigger difference

nekopsykose:
on my machine:  

```
# before:
$ hyperfine -N 'bash bench.sh'
Benchmark 1: bash bench.sh
  Time (mean ± σ):      5.203 s ±  0.022 s    [User: 2.607 s, System: 2.787 s]
  Range (min … max):    5.174 s …  5.244 s    10 runs
$ doas hyperfine -N './tools/depmod 6.9.8-0-generic'
Benchmark 1: ./tools/depmod 6.9.8-0-generic
  Time (mean ± σ):      2.571 s ±  0.013 s    [User: 1.145 s, System: 1.408 s]
  Range (min … max):    2.547 s …  2.591 s    10 runs

# after:
$ hyperfine -N 'bash bench.sh'
Benchmark 1: bash bench.sh
  Time (mean ± σ):      4.103 s ±  0.012 s    [User: 2.364 s, System: 1.934 s]
  Range (min … max):    4.088 s …  4.119 s    10 runs

$ doas hyperfine -N './tools/depmod 6.9.8-0-generic'
Benchmark 1: ./tools/depmod 6.9.8-0-generic
  Time (mean ± σ):      1.481 s ±  0.009 s    [User: 0.887 s, System: 0.576 s]
  Range (min … max):    1.468 s …  1.494 s    10 runs
```

the reason you see much lower values is probably because the allocator in your case is the glibc one, which has a particular behaviour around these reallocs where it gets to skip the whole memcpy in some cases. with other allocators (in our case it's llvm scudo, could also test the stock musl malloc-ng one) this behaviour is not present so you get the following behaviour:

https://share.firefox.dev/4bA5glJ

lucasdemarchi:
> That's due to a specific behavior of the glibc allocator (where directly subsequent reallocs with no allocations inbetween won't always actually allocate, however this comes at a cost elsewhere), but on libcs other than glibc (or with different allocators) it's going to make a much bigger difference

so for both of you it's llvm scudo... right?  Can you add these bench notes to the commit message?

q66:
it's not particular to scudo, likely to happen with anything not specifically glibc allocator (having the same quirk as glibc comes with excess memory fragmentation)

but sure, i can add something

q66:
updated commit message, sorry i forgot :)

lucasdemarchi:
Rebased the branch and force-pushed here so it runs CI checks. LGTM, thanks.

lucasdemarchi:
Applied, thanks.

Issue #47, opened 2024-07-17
Isolate (de)compression libraries


lucasdemarchi:
Isolate decompression libraries routines in their own files so it's easier to eventually load them ondemand

PR #58, opened 2024-07-19
Move decompression functions to separate files

lucasdemarchi:
Closes: #47


lucasdemarchi:
> Can we use `if ENABLE_XZ` within the C file(s) instead?

I don't see a point in having the compiler handle a completely empty file. IMO it also makes it harder to understand how something is disabled by having to open each individual file.

It's not how it's usually done in projects like kernel (with files being included according to kconfig) or systemd (with meson).

lucasdemarchi:
Applied, thanks.

PR #67, opened 2024-08-13
Fix typos

stoeckmann:
Typos found with codespell.


lucasdemarchi:
The shoemaker's son always goes barefoot :)

Thanks for the fixes, applied.

PR #75, opened 2024-08-16
testsuite: Rename test-user

lucasdemarchi:
First version of the patches implementing weakdep called them "user dep", hence the name of the test. That doesn't make sense anymore after the rename.  Rename the test accordingly.


lucasdemarchi:
Applied, thanks.

PR #76, opened 2024-08-16
Codespell fixes, CI action and couple of action updates

evelikov:
This builds up on the previous codespell typo fixes by addressing the remaining ones - note the github action flagged a few more, that a local run did not.

Then we add codespell CI action, alongside a `.editorconfig` so the former is properly formatted. Followed by dependabot.yml to ensure the actions are up-to date (NOTE: repo config is needed, see inline comment), apply some security tweaks (use hashes for actions, restrict permissions) and finally we enable schedule-based runs.

Technically this could be split in multiple PRs - 3 or so - although it  seemed like an overkill, considering how tiny things are. That said, if it makes things clearer, I don't mind splitting.

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,0 +1,2 @@
+caf
+parm
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,27 @@
+# To use this config on you editor, follow the instructions at:
+# http://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+tab_width = 8
+
+[*.{c,h}]
+indent_style = tab
+max_line_length = 80
+
+[{Makefile*,*.mk}]
+indent_style = tab
+
+[*.yml]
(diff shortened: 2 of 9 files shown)


.editorconfig
```diff
@@ -0,0 +1,27 @@
...
+
+[{Makefile*,*.mk}]
+indent_style = tab
+
+[*.yml]
+indent_style = space
+indent_size = 2
+
+[*.patch]
+trim_trailing_whitespace = false
```
lucasdemarchi, line 23: .patch is irrelevant, no?

.codespellignore
```diff
@@ -0,0 +1,2 @@
+caf
```
lucasdemarchi, line 1: bonus point if you merge a patch in tar to allow using `calf` as an alias to `caf` :)

lucasdemarchi (approved):

lucasdemarchi:
Applied, thanks.

PR #82, opened 2024-08-21
libkmod: Fix memory leak on error path

stoeckmann:
If realloc fails, do not override the still valid pointer with NULL. Otherwise freeing the iterator won't free the previously allocated buffer.


lucasdemarchi:
Applied, thanks.

PR #83, opened 2024-08-21
libkmod: Properly treat erroneous modules.builtin.modinfo files

stoeckmann:
The file modules.builtin.modinfo is supposed to consist of NUL terminated strings. If the file contains a string which is not properly NUL terminated, treat it as error. Also make sure that errors never end up with errno being 0. These code paths negate the errno value to get a negative value to propagate the fact that an error occurred. -0 would be 0 again so errors are not detected.

Proof of Concept:

1. Create a copy of module files of currently running system
```
IMGDIR=$(mktemp -d)
KDIR=/lib/modules/$(uname -r)
mkdir -p $IMGDIR/$KDIR
cp $KDIR/modules.* $IMGDIR/$KDIR
```

2. Create a modules.builtin.modinfo file without terminating NUL
```
dd if=/dev/zero bs=8192 count=1 | tr '\0' 'A' > $IMGDIR/$KDIR/modules.builtin.modinfo
```

3. Run modinfo for an existing built-in module (rapl in my case)
```
modinfo -b $IMGDIR rapl
```

The modinfo process crashes due to an invalid free later in codepath. To be precise here due to this line:
https://github.com/kmod-project/kmod/blob/9fc3aa152950dc0cbb2d8705d732089634a3795f/libkmod/libkmod-module.c#L2523
The `strings` variable is not initialized to NULL, which is okay because a negative return value of `kmod_builtin_get_modinfo` is supposed to indicate an error in which case `strings` is never set. But due to missing `errno` assignment, the return value is 0, implying that everything is okay.

### DIFF
--- a/libkmod/libkmod-builtin.c
+++ b/libkmod/libkmod-builtin.c
@@ -210,8 +210,10 @@ static bool kmod_builtin_iter_get_modname(struct kmod_builtin_iter *iter,
 	size_t linesz, len;
 	off_t offset;
 
-	if (iter->pos == iter->size)
-		return false;
+	if (iter->pos == iter->size) {
+		sv_errno = EINVAL;
+		goto fail;
+	}
 
 	line = NULL;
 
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
> Do you see value in having the PoC in-tree and triggered on `make check`? The existing test infra allows for easy spawning depmod/friends instance and has some handy helpers.

I think that's a good idea as it's easy to regress on corner cases. But as you said, non-blocking. Same for me.

Applied, thanks.

PR #86, opened 2024-08-22
Prep and add meson build

evelikov:
With this PR, we do some prep work (and a bunch of fixes) on top of adding meson build to the project.

The series is not yet complete - missing gtk-doc, CI and wrapper like autogen.sh. Although considering the size and notable churn, I think it makes sense to post early.

Happy to split and send to ML - your input is appreciated here as well as (re)naming all the meson toggles.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,6 +59,19 @@ jobs:
           # doesn't quite work. Perhaps our double/sparse checkout is to blame?
           git config --global --add safe.directory '*'
 
+          for moddir in /usr/lib/modules /lib/modules; do
+            if [ -e "$moddir" ]; then
+              kernel=$(find "$moddir" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
+              break
+            fi
+          done
+
+          # There should be one entry - the kernel we installed
+          if [ $(echo "$kernel" | wc -l) -ne 1 ]; then
+            echo >&2 "Error: exactly one kernel must be installed"
+          fi
+          echo "KDIR=$moddir/$kernel/build" >> "$GITHUB_ENV"
+
       - name: configure
         run: |
           mkdir build
@@ -74,7 +87,7 @@ jobs:
         if: ${{ matrix.test == 'yes' }}
         run: |
           cd build
-          make KDIR=/usr/lib/modules/*/build -j$(nproc) check
+          make -j$(nproc) check
 
       - name: install
         run: |
@@ -85,4 +98,4 @@ jobs:
         if: ${{ matrix.test == 'yes' }}
         run: |
           cd build
-          make KDIR=/usr/lib/modules/*/build distcheck
+          make distcheck
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          ../autogen.sh c ${{ matrix.configure }}
+          ../autogen.sh c
 
       - name: build
         run: |
@@ -73,4 +73,4 @@ jobs:
         if: ${{ matrix.test == 'yes' }}
         run: |
           cd build
-          DISTCHECK_CONFIGURE_FLAGS=${{ matrix.configure }} make KDIR=/usr/lib/modules/*/build distcheck
+          make KDIR=/usr/lib/modules/*/build distcheck
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,10 @@ jobs:
       image: ${{ matrix.container }}
 
     steps:
-      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: Sparse checkout the local actions
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          sparse-checkout: .github
 
       - uses: ./.github/actions/setup-ubuntu
         if: ${{ startsWith(matrix.container, 'ubuntu') }}
@@ -46,6 +49,15 @@ jobs:
         if: ${{ startsWith(matrix.container, 'alpine') }}
       - uses: ./.github/actions/setup-debian
         if: ${{ startsWith(matrix.container, 'debian') }}
+
+      - name: Checkout the whole project
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+
+      - name: Set the environment
+        run: |
+          # The second checkout above claims to set safe.directory, yet it
+          # doesn't quite work. Perhaps our double/sparse checkout is to blame?
+          git config --global --add safe.directory '*'
 
       - name: configure
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,22 +16,23 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        include:
-          - container: 'ubuntu:22.04'
-            test: 'yes'
-          - container: 'ubuntu:24.04'
-            test: 'yes'
-          - container: 'archlinux:base-devel'
-            test: 'yes'
-          - container: 'fedora:latest'
-            test: 'yes'
-          - container: 'alpine:latest'
-            test: 'no'
-          - container: 'debian:unstable'
-            test: 'yes'
+        build: ['meson', 'autotools']
+        container:
+          - name: 'ubuntu:22.04'
+            test: 'true'
+          - name: 'ubuntu:24.04'
+            test: 'true'
+          - name: 'archlinux:base-devel'
+            test: 'true'
+          - name: 'fedora:latest'
+            test: 'true'
+          - name: 'alpine:latest'
+            test: 'false'
+          - name: 'debian:unstable'
+            test: 'true'
 
     container:
-      image: ${{ matrix.container }}
+      image: ${{ matrix.container.name }}
 
     steps:
       - name: Sparse checkout the local actions
@@ -40,15 +41,15 @@ jobs:
           sparse-checkout: .github
 
       - uses: ./.github/actions/setup-ubuntu
-        if: ${{ startsWith(matrix.container, 'ubuntu') }}
+        if: ${{ startsWith(matrix.container.name, 'ubuntu') }}
       - uses: ./.github/actions/setup-archlinux
-        if: ${{ startsWith(matrix.container, 'archlinux') }}
+        if: ${{ startsWith(matrix.container.name, 'archlinux') }}
       - uses: ./.github/actions/setup-fedora
-        if: ${{ startsWith(matrix.container, 'fedora') }}
+        if: ${{ startsWith(matrix.container.name, 'fedora') }}
       - uses: ./.github/actions/setup-alpine
-        if: ${{ startsWith(matrix.container, 'alpine') }}
+        if: ${{ startsWith(matrix.container.name, 'alpine') }}
       - uses: ./.github/actions/setup-debian
-        if: ${{ startsWith(matrix.container, 'debian') }}
+        if: ${{ startsWith(matrix.container.name, 'debian') }}
 
       - name: Checkout the whole project
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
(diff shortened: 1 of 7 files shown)


Makefile.am
```diff
@@ -196,7 +196,6 @@ EXTRA_DIST += \
 	testsuite/module-playground/dummy.sha1 \
 	testsuite/module-playground/dummy.sha256 \
 	testsuite/module-playground/Makefile \
-	testsuite/module-playground/Makefile.arch \
```
lucasdemarchi, line 213: `testsuite: remove cross compilation test and binaries`

I'm not entirely sure about this. The commit message doesn't convey the real reason: it's expected the libkmod (and more specifically modinfo) to work correctly with a foreign arch, even bigendian ones or different bit width. I *think* to cover that we would need to add support to cross-compile the kernel modules for a few archs.

I'm not really against the removal, but we need to fix the commit message (required) and hopefully add coverage in another way for what we are dropping here.

lucasdemarchi: > Err not sure how I can reshuffle the commit message - concrete examples will be greatly appreciated.

when building the kernel, it's expected that this will work:

```
make ARCH=mips CROSS_COMPILER=mips-sde-elf- O=build-mips
modinfo build-mips/any/path/to/module.ko
```

> Wrt the coverage there are paid or self-hosted runners - is that something you're interested in funding or settings up?

that wouldn't solve it. I don't want to build in another arch (at least not for this issue). I want modinfo for x86-64 to work for random arch, like it does  today, without needing modinfo from that random arch:

```
$ modinfo ./testsuite/module-playground/mod-simple-sparc64.ko
filename:       /home/ldmartin/p/kmod/./testsuite/module-playground/mod-simple-sparc64.ko
license:        LGPL
author:         Lucas De Marchi <lucas.demarchi@intel.com>
srcversion:     84C5336E000C5A4C187E5B4
depends:        
vermagic:       3.19.0 SMP mod_unload modversions
```
and no, I'm not running on a sparc machine and of course not using a 3.19 kernel. Yet, my modinfo from today works for that module.

> An alternative is to pull a cross gcc and use that - Arch has a few. Haven't really thought about how it will be integrated - generally meson and makefiles don't play too well.

yep, that could do it.  cross-compiling a kernel module is pretty easy as long as you have the toolchain. See above.

shared/missing.h
```diff
@@ -20,6 +20,7 @@
 #endif
 
 #ifndef __NR_finit_module
+# warning __NR_finit_module missing - kmod might not work correctly
```
lucasdemarchi, line 19: `missing.h: warn on missing __NR_finit_module`

> Another, perhaps better option, is to outright fail the build, since 
> if this triggers we have semi-broken kmod. Namely: rmmod will work, yet
> insmod would fail.

did we break the fallback to init_module() when finit_module() is missing?  By now I think we can start warning since it has been a decade since finit_module() was introduced, but I don't like outright breaking the build if we don't have a good reason to.

shared/util.c
```diff
@@ -387,9 +387,13 @@ static inline int is_dir(const char *path)
...
 {
-	char *start, *end;
+	_cleanup_free_ char *start = strdup(path);
+	char *end;
 
-	start = strndupa(path, len);
+	if (start == NULL)
+		return -ENOMEM;
+
+	start[len] = '\0';
```
lucasdemarchi, line 396: so now you depend on len being >= strlen(path), otherwise you are writing past the allocation. I don't think this change is actually required and we'd better leave this to its own PR.

Anyway, better alternative here would be do: 

```
start = memdup(path, len + 1);
if (!start)
        return -ENOMEM;

start[len] = '\0';
```

shared/missing.h
```diff
@@ -3,21 +3,17 @@
...
+ * Macros pulled from linux/module.h, to avoid pulling the header.
+ *
+ * In practise very few people have it installed and distros do not install the
+ * relevant package during their build.
+ *
+ * Values are UAPI, so they cannot change.
+ */
+#define MODULE_INIT_IGNORE_MODVERSIONS 1
+#define MODULE_INIT_IGNORE_VERMAGIC 2
+#define MODULE_INIT_COMPRESSED_FILE 4
```
lucasdemarchi, line 16: grepping from this of the PR:

```
$ pacman -Ql linux-api-headers | cut -d' ' -f2 | grep -v '.*/$' | sed '/^[[:space:]]*$/d' |  sed 's@/usr/include/@@' | while read f; do git grep  $f; done
shared/missing.h: * Macros pulled from linux/module.h, to avoid pulling the header.
testsuite/module-playground/mod-fake-cciss.c:#include <linux/module.h>
testsuite/module-playground/mod-fake-hpsa.c:#include <linux/module.h>
testsuite/module-playground/mod-fake-scsi-mod.c:#include <linux/module.h>
testsuite/module-playground/mod-foo-a.c:#include <linux/module.h>
testsuite/module-playground/mod-foo-b.c:#include <linux/module.h>
testsuite/module-playground/mod-foo-c.c:#include <linux/module.h>
testsuite/module-playground/mod-foo.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-a.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-b.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-c.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-d.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-e.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-f.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-g.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-h.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-i.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-j.c:#include <linux/module.h>
testsuite/module-playground/mod-loop-k.c:#include <linux/module.h>
testsuite/module-playground/mod-simple.c:#include <linux/module.h>
testsuite/module-playground/mod-weakdep.c:#include <linux/module.h>
```

ok, the remaining ones are the ones from linux-headers. Only caveat is that if we end up including a header and get linux/module.h indirectly include, we break the build. Unlikely though. Maybe just copy the header and adjust the include path to handle it:

shared/missing/linux/module.h

In igt we do a similar thing with the uapi headers: we just pull them in to avoid depending on them (since there we are usually ahead of the any released kernel anyway)

lucasdemarchi:
@evelikov I went through the commits, until the part it actually adds meson. I picked most of them, but left out some controversial ones or that needs some rework.  I pushed to the ci-test branch until it passed... just need to flip the flag in CI to make it pass in Debian.  I rebased the remaining commits on that branch and they all apply cleanly, so re-ordering shouldn't cause you any problem. Are you ok with me pushing those to master branch? Thanks.

Makefile.am
lucasdemarchi: ack

shared/util.c
lucasdemarchi: I'm ok with dropping it as indeed no call to mkdir_p is in a hot path and this already caused too many problems. If we need it in the future, we can copy strndupa_safe() from systemd and forget getting that from libc.

Let's just used memdup instead for now.

lucasdemarchi:
Thanks. Just merged those patches and rebased this PR.

lucasdemarchi:
```
./bin/build-scdoc.sh
./bin/kmod-symlink.sh
./testsuite/setup-rootfs.sh
./testsuite/setup-modules.sh
```

not happy with ./bin and not using a single dir for all of them. Can we always use a ./scripts dir and move everything there?

lucasdemarchi:
> I'm wondering if splitting the monolithic meson.build to per-directory - man/meson.build shared/meson.build etc - would be cleaner? I imagine that the single Makefile.am currently used was because recursive makefiles are a PITA?

I think splitting it would be good. At least man pages and testsuite would benefit from a separate meson.build. I'm not so sure about shared, libkmod and tools. Maybe leave them together.

shared/util.c
lucasdemarchi: shared/util.h:void *memdup(const void *p, size_t n) __attribute__((nonnull(1)));

lucasdemarchi:
> The latter is missing from CI - is that something of interest?

I don't think Ubuntu 20.04 will change the kmod version.  What's the minimum version we can support without going out of our way with compat/old-way-of-doing-things? If we can support 0.53.2, great, otherwise, don't bother. Looking at systemd as reference:

```
commit 35c035a1a76cd7aa633c483f0094f03395a9c28f
Author: Yu Watanabe <watanabe.yu+github@gmail.com>
Date:   Mon May 16 04:08:21 2022 +0900

    meson: bump required version to 0.60.0
    
    To suppress warning about install_tag argument.

commit ff5becf585eeab3f47fe532180c682363fe4c3b3
Author: Yu Watanabe <watanabe.yu+github@gmail.com>
Date:   Mon May 16 03:53:01 2022 +0900

    meson: bump required version to 0.59.0
    
    fs.stemp() and fs.name() can take File object since 0.59.0.

commit 80dc9ad98c29c708a8cffbcade44ebf7f3f355e5
Author: Yu Watanabe <watanabe.yu+github@gmail.com>
Date:   Mon May 16 03:36:34 2022 +0900

    meson: bump required version to 0.56.0

commit 573c0dc13406be6575cb9177d8fe818a45426ba8
Author: Yu Watanabe <watanabe.yu+github@gmail.com>
Date:   Mon May 16 03:22:33 2022 +0900

    meson: bump required version to 0.55.0

commit daf4e78e4841a527ba62da48fd00e2e6b8805193
Author: Yu Watanabe <watanabe.yu+github@gmail.com>
Date:   Mon May 16 03:21:16 2022 +0900

    meson: bump required version to 0.54.0

commit 9247df5a8852d11b64a6309d994e3d2afe275250
Author: Zbigniew Jędrzejewski-Szmek <zbyszek@in.waw.pl>
Date:   Tue Jul 27 16:16:07 2021 +0200

    meson: require 0.53.2 and drop some workarounds for old meson
    
    Ubuntu Bionic 18.04 has 0.45, so it was below the previously required
    minimum version already. Focal 20.04 has 0.53.2. Let's require that
    and use various features that are available.
```

I believe our use of meson would be simpler than theirs, so if possible support older versions  as apparently they didn't have a very specific reason for changing. Looking at mesa, I also see their bumped versions not being necessarily a show stopper for kmod to support 0.53.

> 
> Edit: alternatively, we can grab newer meson off pip

that would make it harder for distros to package kmod.

rfc1036:
> I don't think Ubuntu 20.04 will change the kmod version.
Of course not.

If you also want to support backports to Debian/Ubuntu stable then a 1.0.1 baseline is enough.

> that would make it harder for distros to package kmod.
Almost make it impossible really.

rfc1036:
A stable update to a new version is not acceptable unless there are significant bugs, and even in that case it will usually be more appropriate to just backport the fixes.

I do not think that there is any point in running kmod CI on any stable Debian or Ubuntu release, since the distributions will never update kmod there, and I also doubt that anybody needs installing it manually. Targeting Debian unstable (and maybe whatever is the new Ubuntu release in progress) should be enough.

I try to frequently package new snapshots for unstable because the releases have been infrequent and this way I can give more exposure to the new code, but if you believe this to be a bad idea then I will stop.

lucasdemarchi:
on the other side, requiring everyone to run bleeding edge means you won't have many developers. As I said, let's try to support a reasonable minimum version and raise the dependency once we require it. If it's 1.0.1, then fine.  *developers* may very well develop/test by installing from pypi. Let's just not depend on "latest-from-pypi", making it harder for distros to support a new kmod version.

lucasdemarchi (approved):

scripts/setup-modules.sh
```diff
@@ -0,0 +1,18 @@
...
+export MODULE_DIRECTORY=$4
+
+# TODO: meson allows only out of tree builds
+if test "$SRCDIR" != "$BUILDDIR"; then
+	rm -rf "$MODULE_PLAYGROUND"
+	mkdir -p "$(dirname "$MODULE_PLAYGROUND")"
+	cp -r "$SRCDIR/$MODULE_PLAYGROUND" "$MODULE_PLAYGROUND"
+fi
+
+"${MAKE-make}" -C "$PWD/$MODULE_PLAYGROUND" modules
```
lucasdemarchi, line 18: when called from the autotools build, this will honor the parallelism set via -j, whereas it won't for meson. Simple test with meson changing this to include `-j$(nproc)` speeds up the build by a lot on my aged TGL laptop:

command: time ninja -C build-meson test
Before: real    0m13.444s
After: real    0m5.201s

Maybe we should set -j$(nproc) in MAKEFLAGS when calling from meson?

lucasdemarchi:
I need to think more about some of the recently included commits... It seems they are over-complicating it

lucasdemarchi:
Applied most of the patches, leaving out the recently introduced ones.

meson.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
lucasdemarchi, line 1: do we actually need a meson.sh?  I think an update to the README.md should be sufficient for the "commonly used args".

.github/workflows/main.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         run: |
           mkdir build
           cd build
-          ../autogen.sh c ${{ matrix.configure }}
+          ../autogen.sh c
```
lucasdemarchi: but isn't matrix.configure a good way to support different build configurations?

.github/workflows/main.yml
```diff
@@ -16,25 +16,24 @@ jobs:
...
+        container:
+          - 'ubuntu:22.04'
+          - 'ubuntu:24.04'
+          - 'archlinux:base-devel'
+          - 'fedora:latest'
+          - 'alpine:latest'
+          - 'debian:unstable'
+        build:
+          - 'meson'
+          - 'autotools'
```
lucasdemarchi, line 28: in the previous version we had a flat list, which IMO is cleaner. Can we keep 'test' as is and just add meson/autotools permutations?

.github/workflows/main.yml
```diff
@@ -16,25 +16,24 @@ jobs:
...
 
     container:
       image: ${{ matrix.container }}
 
     steps:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          sparse-checkout: .github
```
lucasdemarchi, line 36: humn... apparently 2 different things in this commit:  one is about building with meson, another is about using git to clone the repo.

lucasdemarchi (changes requested):

meson.sh
rfc1036: I say make this simple: if you want to switch to meson then remove the autotools stuff at the same time. Or else there will be no reason to switch it for distributions.

.github/workflows/main.yml
lucasdemarchi: do we need it as prep commit? is it faster?

.github/workflows/main.yml
lucasdemarchi: See branch ci-test: https://github.com/kmod-project/kmod/commit/4b2a79776461fa923a8d148f92537b59652075fd and execution: https://github.com/kmod-project/kmod/actions/runs/10687966057

IMO nice thing that the configuration for each run is all inside the matrix definition, and not spread in the tasks.  If you are fine with that I can squash it in your last commit.

PR #92, opened 2024-09-02
shared: use memdup over stdndupa

evelikov:
The stdndupa has a couple of gotchas:
 - allocates memory on stack via alloca(3)... where we pass it a user-provided string in at least one instance
 - it's a GNU extension missing on musl and bionic

The mkdir_p() function is not a hot path, so using heap allocation is perfectly fine. Swap the stdndupa for our local helper memdup.


lucasdemarchi:
Applied, thanks

Issue #65, opened 2024-07-24
Fix testsuite with musl


lucasdemarchi:
Currently the build on Alpine doesn't run the testsuite because the build fails with musl. This should be fixed and start testing with it in CI.

PR #97, opened 2024-09-03
testsuite/path: keep 64 variants for _GLIBC_

evelikov:
Since musl v1.2.4 and Alpine 3.18, circa May 2023, the LFS64 variants
are no longer exported by libc.so.

In addition musl headers provide relevant "#define foo64 foo" macros,
when _LARGEFILE64_SOURCE is defined and symbol resolution is deferred to
the runtime linker. See musl commit 246f1c811448 ("remove LFS64 symbol
aliases; replace with dynamic linker remapping") more details.

Earlier versions of musl had the symbols exposed by libc, although the
^^ macros were also available when _GNU_SOURCE is set, as our build
does.

Most importantly, we do not explicitly use the 64 variants and musl
doesn't implicitly pull them either. So in practise building with musl,
the binaries do not have any 64 references, thus there is no need to
wrap them.

AFAICT there is no obvious way of detecting musl, apart from the lack of
of __GLIBC__, so pull features.h (which defines it) and check on that.

There is a small chance the __GLIBC__ check will miss-fire on other
implementations like bionic, dietlibc or ulibc(-ng). We don't run the
tests on those platforms (although we could) so it should be fine.

With that we can install tar (for make distcheck) and enable the tests
in CI \o/

Closes: https://github.com/kmod-project/kmod/issues/65

---

Another alternative to `__GLIBC__` is to check each function individually, which means a lot more build system magic. Unless we decide to slim down the list a bit. At a glance, we don't use `lstat` so that can go? In addition the `__x` variants are defined in LSB as ABI yet it seems like glibc has stopped forwarding to them, as they dropped their declarations from the public headers.

Personally doing the cleanup && individual check sound like the better option. Although I don't have strong preferences.

### DIFF
--- a/testsuite/path.c
+++ b/testsuite/path.c
@@ -184,11 +184,15 @@ WRAP_1ARG(DIR*, NULL, opendir);
 WRAP_1ARG(int, -1, chdir);
 
 WRAP_2ARGS(FILE*, NULL, fopen, const char*);
-WRAP_2ARGS(FILE*, NULL, fopen64, const char*);
 WRAP_2ARGS(int, -1, mkdir, mode_t);
 WRAP_2ARGS(int, -1, access, int);
 WRAP_2ARGS(int, -1, stat, struct stat*);
 WRAP_2ARGS(int, -1, lstat, struct stat*);
+
+WRAP_OPEN();
+
+#if HAVE_DECL___GLIBC__
+WRAP_2ARGS(FILE*, NULL, fopen64, const char*);
 WRAP_2ARGS(int, -1, stat64, struct stat64*);
 WRAP_2ARGS(int, -1, lstat64, struct stat64*);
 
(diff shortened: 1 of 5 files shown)


testsuite/path.c
```diff
@@ -184,11 +184,15 @@ WRAP_1ARG(DIR*, NULL, opendir);
...
 WRAP_2ARGS(FILE*, NULL, fopen, const char*);
-WRAP_2ARGS(FILE*, NULL, fopen64, const char*);
 WRAP_2ARGS(int, -1, mkdir, mode_t);
 WRAP_2ARGS(int, -1, access, int);
 WRAP_2ARGS(int, -1, stat, struct stat*);
 WRAP_2ARGS(int, -1, lstat, struct stat*);
+
+WRAP_OPEN();
+
+#if HAVE_DECL___GLIBC__
```
lucasdemarchi, line 194: not sure we can rely on __GLIBC__ and it'd be better to check case by case. Can't we simply

```
#ifndef fopen64
WRAP_2ARGS(FILE*, NULL, fopen64, const char*);
#endif
```

and so on? Because if it's coming from the header as a define, then we shouldn't wrap it (as we obviously can't).

lucasdemarchi:
I don't like the blanket check on __GLIBC__, but I can live with it. Let's merge this. Thanks.

PR #101, opened 2024-09-04
shared: use size_t for strbuf

stoeckmann:
The unsigned datatype could overflow on 64 bit systems with sufficiently large index files. Switch to size_t just to be safe.

Since the API of strbuf is not exported, this has no side-effect for libkmod users.


stoeckmann:
A proper test suite would require at least 2 GB (and a bit more) of free RAM, which makes it rather difficult to write a reliable test.

lucasdemarchi:
Applied, thanks.

PR #102, opened 2024-09-04
libkmod: check for NULL in kmod_get_dirname

stoeckmann:
The other KMOD_EXPORT'ed functions protect against ctx being NULL. Add this to kmod_get_dirname purely for consistency reasons.


lucasdemarchi:
Applied, thanks

PR #103, opened 2024-09-04
libkmod: check fstatat return value

stoeckmann:
If fstatat fails, st is not set, which leads to undefined results of subsequent S_ISDIR call.


lucasdemarchi:
Applied, thanks.

PR #104, opened 2024-09-06
meson: Add build-tests to build-dev.ini

lucasdemarchi:
Developers are supposed to run tests to test their code.  CI continues to pass it in the command line as it may need to disable it.


lucasdemarchi:
long long time ago I was porting something to meson and remember I used a stamp file for that. I wouldn't know how to do it again without looking up the docs though.

lucasdemarchi:
Applied.

PR #109, opened 2024-09-06
ci: Publish libkmod docs

lucasdemarchi:
Publish docs with Github Actions to https:://kmod-project.github.io. The action here is a mix of our own build instructions and Github's recommendation for a jekyll-based pipeline.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,62 @@
+name: Publish docs
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["master"]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Actually, it seems it publishes under /kmod. We'd need a separate repo for a org site. End result is this:
https://kmod-project.github.io/kmod/

I think creating a separate repo for the pages would be ok, and maybe figure our how to split it so we can have multiple kmod versions + development one.

lucasdemarchi:
@evelikov thoughts?

lucasdemarchi:
I think I will put this on hold. I'm not sure I like to have it rendered under a /kmod dir. Looking at alternatives, I think I will create a separate repo kmod-project to save the pages. According to https://docs.github.com/en/pages/getting-started-with-github-pages/about-github-pages#types-of-github-pages-sites, that should be a repo named `kmod-project.github.io`. Then we can do:

- front page
- /<release-number>/man
- /<release-number>/libkmod

Thoughts?

Issue #108, opened 2024-09-06
Add 32bit permutation to the CI


evelikov:
Bunch of the truncation issues spotted by stoeckmann are specific to 32bit builds. Add a variant to the CI so we get extra coverage.

PR #110, opened 2024-09-06
ci: add multilib (Arch, Debian, Ubuntu) and enable werror=true

evelikov:
This adds a few simple 32bit permutation in the matrix and promotes all warnings to errors.

At first I though about enabling openssl + compression, although I don't think the extra churn (both downloads and yaml)  is worth it. These three should offer somewhat wide enough coverage as a start.

AFAICT there's no equivalent for Alpine, while Fedora should have one yet I have zero experience there.

Closes: https://github.com/kmod-project/kmod/issues/108

### DIFF
--- a/.github/actions/setup-archlinux/action.yml
+++ b/.github/actions/setup-archlinux/action.yml
@@ -6,6 +6,8 @@ runs:
     - name: Install dependencies
       shell: bash
       run: |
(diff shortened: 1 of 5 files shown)


lucasdemarchi (approved):

lucasdemarchi:
Applied, thanks.

PR #111, opened 2024-09-06
libkmod: smoke test all the API

evelikov:
Aka make sure the library does not crash if we give it a NULL as input argument.

All entry-points but two look to be safe, update the final instances.

---
Follow up on https://github.com/kmod-project/kmod/pull/102#pullrequestreview-2282713826

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -1317,6 +1317,7 @@ KMOD_EXPORT bool kmod_config_iter_next(struct kmod_config_iter *iter)
 
 KMOD_EXPORT void kmod_config_iter_free_iter(struct kmod_config_iter *iter)
 {
-	free(iter->data);
+	if (iter)
+		free(iter->data);
 	free(iter);
 }


libkmod/libkmod-module.c
```diff
@@ -511,8 +511,10 @@ KMOD_EXPORT int kmod_module_new_from_name_lookup(struct kmod_ctx *ctx,
 
 KMOD_EXPORT int kmod_module_unref_list(struct kmod_list *list)
 {
-	for (; list != NULL; list = kmod_list_remove(list))
```
lucasdemarchi, line 514: humn... what's the problem with this line? if list is NULL it will simply return, like the change you did

libkmod/libkmod-config.c
```diff
@@ -1317,6 +1317,7 @@ KMOD_EXPORT bool kmod_config_iter_next(struct kmod_config_iter *iter)
 
 KMOD_EXPORT void kmod_config_iter_free_iter(struct kmod_config_iter *iter)
 {
-	free(iter->data);
+	if (iter)
```
lucasdemarchi, line 1320: this one looks sane

lucasdemarchi (changes requested):

lucasdemarchi:
Applied, thanks.

PR #112, opened 2024-09-06
Remove the "Ugly assignment + check"

evelikov:
Small prep commit, apart from the actual removal... In case the latter is tad much, the former should be less controversial.

### DIFF
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -754,9 +754,9 @@ KMOD_EXPORT int kmod_module_insert_module(struct kmod_module *mod,
 	return err;
 }
 
-static bool module_is_blacklisted(struct kmod_module *mod)
+static bool module_is_blacklisted(const struct kmod_module *mod)
 {
-	struct kmod_ctx *ctx = mod->ctx;
+	const struct kmod_ctx *ctx = mod->ctx;
 	const struct kmod_config *config = kmod_get_config(ctx);
 	const struct kmod_list *bl = config->blacklists;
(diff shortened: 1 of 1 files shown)


libkmod/libkmod-module.c
```diff
@@ -1072,17 +1072,15 @@ KMOD_EXPORT int kmod_module_probe_insert_module(struct kmod_module *mod,
...
-	 * blacklist and also return the reason why we failed.
-	 * KMOD_PROBE_APPLY_BLACKLIST_ALIAS_ONLY will take effect only if the
-	 * module is an alias, so we also need to check it
-	 */
-	if ((mod->alias != NULL && ((err = flags & KMOD_PROBE_APPLY_BLACKLIST_ALIAS_ONLY)))
-			|| (err = flags & KMOD_PROBE_APPLY_BLACKLIST_ALL)
-			|| (err = flags & KMOD_PROBE_APPLY_BLACKLIST)) {
-		if (module_is_blacklisted(mod))
-			return err;
+	if (module_is_blacklisted(mod)) {
```
lucasdemarchi, line 1075: I'd OR all flags together and check that before checking if module is blacklisted.

lucasdemarchi: Reconsidered: one of these flags are actually **likely** to be set so it doesn't matter checking for the flag before checking if the module is blacklisted. Example: src/shared/module-util.c in systemd.

lucasdemarchi:
Applied, thanks

PR #113, opened 2024-09-06
libkmod: rework index_files::prefix

evelikov:
Swap the empty or alias strings for a is_alias bool. Slightly smaller binary, simpler code and fewer pointers being dereferenced.

### DIFF
--- a/libkmod/libkmod-index.h
+++ b/libkmod/libkmod-index.h
@@ -19,7 +19,7 @@ struct index_file;
 struct index_file *index_file_open(const char *filename);
 void index_file_close(struct index_file *idx);
 char *index_search(struct index_file *idx, const char *key);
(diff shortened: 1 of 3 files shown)


libkmod/libkmod-index.h
```diff
@@ -19,7 +19,7 @@ struct index_file;
 struct index_file *index_file_open(const char *filename);
 void index_file_close(struct index_file *idx);
 char *index_search(struct index_file *idx, const char *key);
-void index_dump(struct index_file *in, int fd, const char *prefix);
+void index_dump(struct index_file *in, int fd, bool is_alias);
```
lucasdemarchi, line 22: I think bool as parameters in functions makes up for really bad and error-prone APIs. In the end the caller is doing foo(bla, true, true, false, true). I actually prefer keeping the prefix

PR #114, opened 2024-09-07
Some clang-format inspired cleanups

evelikov:
Forking some of the cleanups/ideas from https://github.com/kmod-project/kmod/pull/91

Namely:
 - `s/;;//` in a handful of places, from Lucas
 - add trailing commas for multi-line arrays/enums, so clang-format doesn't foobar them
 - remove unused `_noreturn` macro
 - remove _must_check and -Wno-unused-result - see commit for details
 - remove open-coded `_print_format_`
 - add `_nonnull_*` variants, place them on the left hand side of the function return type
 
I threw the series together, although I'm more than happy to split it. Let me know if there's any preferences.

### DIFF
--- a/libkmod/libkmod-internal.h
+++ b/libkmod/libkmod-internal.h
@@ -48,9 +48,9 @@ static _always_inline_ _printf_format_(2, 3) void
 #  endif
 #endif
 
-void kmod_log(const struct kmod_ctx *ctx,
+_printf_format_(6, 7) void kmod_log(const struct kmod_ctx *ctx,
 		int priority, const char *file, int line, const char *fn,
-		const char *format, ...) __attribute__((format(printf, 6, 7))) __attribute__((nonnull(1, 3, 5)));
+		const char *format, ...) __attribute__((nonnull(1, 3, 5)));
 
 struct list_node {
 	struct list_node *next, *prev;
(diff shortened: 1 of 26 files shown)


libkmod/libkmod-internal.h
```diff
@@ -68,16 +68,16 @@ enum kmod_file_compression_type {
 	KMOD_FILE_COMPRESSION_ZLIB,
 };
 
-struct kmod_list *kmod_list_append(struct kmod_list *list, const void *data) _must_check_ __attribute__((nonnull(2)));
-struct kmod_list *kmod_list_prepend(struct kmod_list *list, const void *data) _must_check_ __attribute__((nonnull(2)));
```
lucasdemarchi, line 72: I don't understand why you think this one is ok. The _must_check_ is serving its purpose that nobody should call this function without checking the return.

lucasdemarchi (changes requested):
The removal of _must_check_ seems wrong, rest lgtm

libkmod/libkmod-internal.h
lucasdemarchi: My proposal is that we drop that patch, just replacing with one that moves it to the left. See branch ci-test - 
 up to 0c8f1271eda6 ("Remove -Wno-unused-result")

lucasdemarchi:
Applied, thanks.

PR #115, opened 2024-09-07
testsuite: use size_t with strbuf

evelikov:
We recently updated the API, but forgot to update the tests.

Fixes: 38943b2 ("shared: use size_t for strbuf")

---

Follow-up from https://github.com/kmod-project/kmod/pull/101#issuecomment-2334634793


lucasdemarchi:
Applied, thanks

PR #117, opened 2024-09-08
libkmod: remove internal kmod_list iterators

evelikov:
These were introduced to workaround since the compiler cannot properly optimise things, since the symbols have external linkage.

These days this is not so much an issue. For example comparing a gcc 14 build, with -O2 shows 24 bytes growth with this reverted.

Although in practise anyone looking for extreme optimisations - be that speed or size - should be using LTO and/or PGO.

### DIFF
--- a/libkmod/libkmod-internal.h
+++ b/libkmod/libkmod-internal.h
@@ -79,20 +79,6 @@ _nonnull_(2) struct kmod_list *kmod_list_insert_after(struct kmod_list *list, co
 _nonnull_(2) struct kmod_list *kmod_list_insert_before(struct kmod_list *list, const void *data);
 _must_check_ struct kmod_list *kmod_list_append_list(struct kmod_list *list1, struct kmod_list *list2);
 
-#undef kmod_list_foreach
-#define kmod_list_foreach(list_entry, first_entry) \
-	for (list_entry = ((first_entry) == NULL) ? NULL : (first_entry); \
-		list_entry != NULL; \
-		list_entry = (list_entry->node.next == &((first_entry)->node)) ? NULL : \
-		container_of(list_entry->node.next, struct kmod_list, node))
-
-#undef kmod_list_foreach_reverse
-#define kmod_list_foreach_reverse(list_entry, first_entry) \
-	for (list_entry = (((first_entry) == NULL) ? NULL : container_of(first_entry->node.prev, struct kmod_list, node)); \
-		list_entry != NULL; \
-		list_entry = ((list_entry == first_entry) ? NULL :	\
-		container_of(list_entry->node.prev, struct kmod_list, node)))
-
 /* libkmod.c */
 _nonnull_all_ int kmod_lookup_alias_from_config(struct kmod_ctx *ctx, const char *name, struct kmod_list **list);
 _nonnull_all_ int kmod_lookup_alias_from_symbols_file(struct kmod_ctx *ctx, const char *name, struct kmod_list **list);


lucasdemarchi:
> These days this is not so much an issue. For example comparing a gcc 14 build, with -O2 shows 24 bytes growth with this reverted.

size is not a good measure for this... we'd need to look at the disassembled code and I doubt it will really be any different than it was before.

lucasdemarchi:
> While doing this patch, I compared 100k `insmod already-loaded-module` runs with/without this patch. There was basically zero change, although I admit it's not the best tool for the job. Happy to try others though.

That's not a good test since we statically link our own tools. The problem is for tools linking to libkmod.so, or calls to that symbol inside the .so itself. In this case, for every single kmod_list_foreach we will have a call to kmod_list_next() going through the PLT:

```
0000000000004250 <kmod_list_next@plt>:
    4250:       ff 25 ba 8e 01 00       jmp    *0x18eba(%rip)        # 1d110 <kmod_list_next@@LIBKMOD_5+0xe7f0>
    4256:       68 22 00 00 00          push   $0x22
    425b:       e9 c0 fd ff ff          jmp    4020 <_init+0x20>
```

Right now we have in e.g. kmod_lookup_alias_from_kernel_builtin_file:

```
        kmod_list_foreach(l, *list) {
    5f2d:       48 8b 5d 00             mov    0x0(%rbp),%rbx
        ret = kmod_lookup_alias_from_alias_bin(ctx,
    5f31:       41 89 c4                mov    %eax,%r12d
        kmod_list_foreach(l, *list) {
    5f34:       66 66 2e 0f 1f 84 00    data16 cs nopw 0x0(%rax,%rax,1)
    5f3b:       00 00 00 00 
    5f3f:       90                      nop
    5f40:       48 85 db                test   %rbx,%rbx
    5f43:       74 17                   je     5f5c <kmod_lookup_alias_from_kernel_builtin_file+0x4c>
                struct kmod_module *mod = l->data;
                kmod_module_set_builtin(mod, true);
    5f45:       48 8b 7b 10             mov    0x10(%rbx),%rdi
    5f49:       be 01 00 00 00          mov    $0x1,%esi
    5f4e:       e8 2d 90 00 00          call   ef80 <kmod_module_set_builtin>
        kmod_list_foreach(l, *list) {
    5f53:       48 8b 1b                mov    (%rbx),%rbx
    5f56:       48 3b 5d 00             cmp    0x0(%rbp),%rbx
    5f5a:       75 e4                   jne    5f40 <kmod_lookup_alias_from_kernel_builtin_file+0x30>
```

And after this patch we have:
```
        kmod_list_foreach(l, *list) {
    5f2d:       48 8b 5d 00             mov    0x0(%rbp),%rbx
        ret = kmod_lookup_alias_from_alias_bin(ctx,
    5f31:       41 89 c4                mov    %eax,%r12d
        kmod_list_foreach(l, *list) {
    5f34:       48 85 db                test   %rbx,%rbx
    5f37:       74 29                   je     5f62 <kmod_lookup_alias_from_kernel_builtin_file+0x52>
    5f39:       0f 1f 80 00 00 00 00    nopl   0x0(%rax)
                struct kmod_module *mod = l->data;
                kmod_module_set_builtin(mod, true);
    5f40:       48 8b 7b 10             mov    0x10(%rbx),%rdi
    5f44:       be 01 00 00 00          mov    $0x1,%esi
    5f49:       e8 d2 8f 00 00          call   ef20 <kmod_module_set_builtin>
        kmod_list_foreach(l, *list) {
    5f4e:       48 8b 7d 00             mov    0x0(%rbp),%rdi
    5f52:       48 89 de                mov    %rbx,%rsi
    5f55:       e8 f6 e2 ff ff          call   4250 <kmod_list_next@plt>
    5f5a:       48 89 c3                mov    %rax,%rbx
    5f5d:       48 85 c0                test   %rax,%rax
    5f60:       75 de                   jne    5f40 <kmod_lookup_alias_from_kernel_builtin_file+0x30>
        }

        return ret;
```

However you are correct in the sense that LTO will indeed remove the call and turn that into:

```
       48 85 f6                test   %rsi,%rsi
       74 22                   je     c0ab <kmod_lookup_alias_from_kernel_builtin_file+0x4b>
       48 89 f7                mov    %rsi,%rdi
       eb 05                   jmp    c093 <kmod_lookup_alias_from_kernel_builtin_file+0x33>
       66 90                   xchg   %ax,%ax
       48 8b 3b                mov    (%rbx),%rdi
mod->builtin =
       48 8b 56 10             mov    0x10(%rsi),%rdx
       c7 42 5c 02 00 00 00    movl   $0x2,0x5c(%rdx)
       e8 ad 81 ff ff          call   4250 <kmod_list_next@plt>
       48 89 c6                mov    %rax,%rsi
       48 85 c0                test   %rax,%rax
       75 de                   jne    5f40 <kmod_lookup_alias_from_kernel_builtin_file+0x30>
       75 e5                   jne    c090 <kmod_lookup_alias_from_kernel_builtin_file+0x30>
```

Cool. Still on the fence about applying it, but this looks much better. Are distros enabling LTO nowadays?

lucasdemarchi:
Ok, let's apply this. Done. Thanks.

Issue #85, opened 2024-08-22
depmod fallouts from module-directory


evelikov:
[Initially reported](https://lore.kernel.org/all/CAK7LNARYK-xjBS8puEM9xFtmjBNW6KJ2Qd6f7diZkdEEbUgVHA@mail.gmail.com/) by Masahiro Yamada on the ML, although it seemingly got lost in the noise.

Tl:Dr: People currently use their native kmod when cross-building kernels. So if you use `/usr/lib/modules` for your kmod, while the target platform uses `/lib/modules`, depmod fails.

One option is to require cross-compile (aware) kmod and another to add a new argument to depmod (basedir is not sufficient) Former is more idealistic, although not too practical since a few people have hit this and more to follow.

@hramrach what's your take? Can you lend a hand to fix this bug/regression?

lucasdemarchi:
See my reply: https://lore.kernel.org/all/5gx6vt4tzgk4zvboxrrahexr4ja6zm6fisjshdvnlfihsysqzb@quhp42ydtvh2/

I think we need the -m|--module-dir to be future proof. Kernel should default it to the output of `pkg-config --variable=module_directory kmod`.  For compatibility and the common case (host on /lib/modules, target on /usr/lib/modules) it's possible to do without that by hacking the path passed to depmod (e.g. append the /usr in $INSTALL_MOD_PATH), but more complex cases in which target or host are in random dirs, I don't think so.

rfc1036:
I do not think that Debian needs to care about cross-compiling to targets which do not implement merged-/usr, since it is a more and more unsupported configuration, but please let me know if I am missing anything.

lucasdemarchi:
@rfc1036 the problem is actually cross-compiling to a target with a different --with-module-directory, or when mixing new (on a host with --with-module-directory=<something>) and old (on the target, that is hardcoding /lib/modules before that became configurable)

hramrach:
Since  the kernel support for --with-module-directory= was not merged merely compiling a kernel when you used --with-module-directory= triggers a problem. However, multiple patches to implement that in the kernel were rejected, and it's not clear what is acceptable for Masahiro Yamada to merge.

Of course, you can apply one of the historical versions of the patch locally.

rfc1036:
But how can this relevant **if** `/lib/modules/` and `/usr/lib/modules/` are the same directory, thanks to a symlink?

lucasdemarchi:
[...]

PR #118, opened 2024-09-08
Prep and enable clang-format

evelikov:
This series builds on the initial work in https://github.com/kmod-project/kmod/pull/91

 - reformat and clang-format off various blocks of existing code
 - imports the .clang-format from the kernel and tweaks it
 - runs ^^ and commits the changes in 4 patches, one per top-level folder - do we want the mass conversion?
 - adds CI action which runs on PRs - do we want one for pushes as well?

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,0 +1,18 @@
+name: Check code style with clang-format
+
+on:
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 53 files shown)


lucasdemarchi:
Pushed up to the mass conversion commits. I'd like to postpone those for when we have the #85 fixed to avoid issues with backporting.

lucasdemarchi:
@evelikov can you regenerate the last commits?

lucasdemarchi:
Applied.

PR #120, opened 2024-09-09
meson: Move symlinks up

lucasdemarchi:
Currently it's only possible to test the tools by using the symlink. However, differently from autotools, the symlink is created inside the testsuite dir, and only if build-tests=true. Move it to the same directory that kmod is so it's possible to test without the testsuite.


lucasdemarchi:
Applied, thanks

Issue #119, opened 2024-09-09
meson: build vs run tests vs cache build


lucasdemarchi:
We need to do something to improve meson's behavior wrt testsuite. Current issues comparing to the autotools setup:

1) if build_tests=true, it always rebuilds all the modules from playground
2) if build_tests=false, it runs normally and reports 1 test passing (gtk-doc)
3) it builds the testsuite outside of the `test` target. In autotools the testsuite is built and executed when we invoke `make check`. I think it would not be so bad if we were tracking the build status and only rebuilding when needed... but maybe we can leave the build of anything related to testsuite only when one calls `ninja -C test`?

PR #121, opened 2024-09-09
meson: Stop rebuilding modules over and over

lucasdemarchi:
Instead of removing the sources every time and rebuilding, just use rsync to preserve the timestamps and allow Make to do its job of rebuilding if it changed.

There's a bug in meson that keeps building the testsuite even outside of `ninja test`, but if the build result is cached, we can greatly minimize the impact for developers.

Closes: https://github.com/kmod-project/kmod/issues/119

### DIFF
--- a/.github/actions/setup-alpine/action.yml
+++ b/.github/actions/setup-alpine/action.yml
@@ -18,6 +18,7 @@ runs:
           linux-edge-dev \
           meson \
           openssl-dev \
+          rsync \
           scdoc \
           tar \
           xz-dev \
--- a/.github/actions/setup-archlinux/action.yml
+++ b/.github/actions/setup-archlinux/action.yml
@@ -20,6 +20,7 @@ runs:
           multilib-devel \
           linux-headers \
           meson \
+          rsync \
           scdoc \
           git \
           gtk-doc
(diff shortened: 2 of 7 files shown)


lucasdemarchi:
Oops, that would add a dependency on rsync so we'd need to add that to our containers. I can add it once we agree on approach.

lucasdemarchi:
I tried a few options that I normally use for in-tree kernel modules (like setting KBUILD_OUTPUT or O=), but they don't seem to work for external modules. I will change the needed parts to use rsync, but we may migrate to something simpler if we find a way.

lucasdemarchi:
Applied, thanks

lucasdemarchi:
> I tried a few options that I normally use for in-tree kernel modules (like setting KBUILD_OUTPUT or O=), but they don't seem to work for external modules. I will change the needed parts to use rsync, but we may migrate to something simpler if we find a way.

so... it turns out out-of-tree module builds with separate build dir was never officially supported. It happened to work until 6.9 with some variable overriding. I chatted with Masahiro and he ended up adding the official way:
https://lore.kernel.org/linux-kbuild/20240917141725.466514-1-masahiroy@kernel.org/

However it's not even in 6.12.  Once we have something like that we can drop the additional copy of the binary and just build it in place

PR #124, opened 2024-09-10
kmod: Avoid negative exit codes

stoeckmann:
Exit codes shall be non-negative to comply with standards.

How to reproduce:

```
ln -s $(which kmod) something
./something
echo $?
```

On my system, the exit code is `254`, which falsely indicates that some signal has been used to terminate the process.


lucasdemarchi (approved):

lucasdemarchi:
Applied, thanks

PR #125, opened 2024-09-11
build: Fix KDIR again

lucasdemarchi:
KDIR is not related to the what we configure in kmod's build. It's only used in kmod to locate where the distro's kernel source/headers is, which may be different from what we are configuring the (under development) kmod with.

Remove the setting from meson/autotools and figure it out inside the module-playground Makefile what should be used. For advanced use cases, KDIR= can be passed to override the location.

For our own tests, which includes testing with a different module_directory, scripts/setup-rootfs.sh will copy the module to the desired location according to the map defined in the script.

Fixes: 27ff72732666 ("testsuite: correct the default KDIR")

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -178,7 +178,7 @@ EXTRA_DIST += scripts/setup-modules.sh
 EXTRA_DIST += scripts/setup-rootfs.sh
 
 MODULE_PLAYGROUND = testsuite/module-playground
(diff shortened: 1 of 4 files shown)


lucasdemarchi:
> Looking at the original patch - it seems like it was also borked.

yep, it was always broken and we never got it because we were not testing it in this mixed scenario (and just now distros started to change it on their side).

lucasdemarchi:
@evelikov  can you check the comment if it's good enough?

lucasdemarchi:
@evelikov it looks like I merged this while merging #122 , so if there's a typo in the comment we will need a fix on top :)

PR #127, opened 2024-09-13
Add depmod -M option to replace kmod.pc

hramrach:
Move ligkmod features from kmod.pc to libkmod.pc
    
    sysconfdir, distconfdir, module_compressions, and module_signatures are
    libkmod features, these should be in libkmod.pc, not kmod.pc.
    
    module_directory is special: While libkmod has a default for when the
    directory is not specified depmod does spefcify it, and has its own
    embetted value.
    
    Technically it's possible to have kmod and libkmod with mismatched
    values due to them being built separately. Also on a binary OS
    distribution the libkmod.pc would be installed only with libkmod
    development headers, not the tool.

depmod: Add -M parameter to print the value of MODULE_DIRECTORY
    
    This is the only value left in kmod.pc.
    
    It is an overkill to create a pkgconfig file for one value.
    
    Also this file was initially introduced, then moved to a different
    loaction because its purpose is misunderstood. Using it is awkward
    because the binary search path and pkgconfig path must be matched in
    case there is a possibility that multiple builds of kmod exist.
    
    As to why multiple build of kmod might exist on a system
    
     - custom test build
     - build of kbuild that matches some target system which has different
       settings for sysconfdit, distconfdir, or module_directory
     - per-architecture tools on 'multiarch' system that is set up for
       building binaries/boot images for arbitrary architectures
       - when these use a filesystem scheme in which per-architecture
         libraries and modules are stored in per-architecture directories
         Linux sticks out like a sore thumb with all modules in /lib
    
    While in the past it was fine to have One Kmod To Rule Them All because
    everything was hardcoded to a fixed value (and all compression methods
    and signature options could be built at the same time) this is no longer
    the case.
    
    Values for sysconfdir, distconfdir, and module_directory may differ
    between systems, and thus a build can be hosted on a system with one set
    of values but targeted for a system with different set of values.
    
    The -m option is provided to make it possible to override
    module_directory.
    
    However, for the modules to be installed in the correct location by
    kernel build the kernel build needs a way to determine the default value
    of module_directory, which this -M option provides.
    
    With a corresponding change to the Linux kbuild system installing
    modules should just work for both host with arbitrary module_directory
    and target for which correctly configured target-depmod is available and
    used during build.
    
    There is always the option to pass arbitrary value as argument to the -m
    option to build modules for a target system that uses different
    module_directory, or to move the installed modules after the fact.
    However, that requires somehow examining the target system out-of-band
    to determine the correct value, and hardcode such value in a build
    script. When working with multiple different targets regularly using a
    target-specific depmod is a way to manage the depmod build-time settings
    in a standardized way that works equally for other tools.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -121,7 +121,6 @@ libkmod_libkmod_internal_la_DEPENDENCIES  = $(libkmod_libkmod_la_DEPENDENCIES)
 libkmod_libkmod_internal_la_LIBADD = $(libkmod_libkmod_la_LIBADD)
 
 pkgconfig_DATA = libkmod/libkmod.pc
-noarch_pkgconfig_DATA = tools/kmod.pc
 
 bashcompletiondir=@bashcompletiondir@
 dist_bashcompletion_DATA = \
--- a/configure.ac
+++ b/configure.ac
@@ -312,7 +312,6 @@ AC_CONFIG_FILES([
 	libkmod/docs/Makefile
 	libkmod/docs/version.xml
 	libkmod/libkmod.pc
-	tools/kmod.pc
 ])
 
 
--- a/libkmod/libkmod.pc.in
+++ b/libkmod/libkmod.pc.in
@@ -2,6 +2,11 @@ prefix=@prefix@
 exec_prefix=@exec_prefix@
 libdir=@libdir@
 includedir=@includedir@
+sysconfdir=@sysconfdir@
+distconfdir=@distconfdir@
+module_directory=@module_directory@
+module_signatures=@module_signatures@
+module_compressions=@module_compressions@
 
 Name: libkmod
 Description: Library to deal with kernel modules
--- a/man/depmod.8.scd
+++ b/man/depmod.8.scd
@@ -12,6 +12,8 @@ depmod - Generate modules.dep and map files.
 *depmod* [*-e*] [*-E* _Module.symvers_] [*-F* _System.map_] [*-n*] [*-v*] [*-P* _prefix_]
 \ \ \ \ \ \ \ \[*-w*] [_version_] [_filename_...]
 
+*depmod* *-M*
+
 # DESCRIPTION
 
 Linux kernel modules can provide services (called "symbols") for other modules
@@ -79,6 +81,10 @@ rather than the current kernel version (as returned by *uname -r*).
 	This expects all input files under
 	_/tmp/build/kernel-modules/$(uname -r)_ and generates index files under
 	that same directory.
+
+*-M*
+*--moduledir-defualt*
+	Print the moduledir prefix @MODULE_DIRECTORY@ set at compile time.
 
 *-o* _outdir_
 *--outdir* _outdir_
--- a/meson.build
+++ b/meson.build
@@ -363,11 +363,25 @@ libkmod = shared_library(
   install : true,
 )
 
+_libkmod_variables = [
+  'sysconfdir=' + sysconfdir,
+  'distconfdir=' + distconfdir,
+  'module_directory=' + moduledir,
+  'module_signatures=' + module_signatures,
+]
+
+# XXX: Support for empty variables was added in meson v1.4.0.
+# pkgconf behaves identically on missing and empty variable.
+if module_compressions != ''
+  _libkmod_variables += ['module_compressions=' + module_compressions]
+endif
+
 pkg.generate(
   name : 'libkmod',
   description : 'Library to deal with kernel modules',
   libraries : libkmod,
   requires_private : libkmod_deps,
+  variables: _libkmod_variables
 )
 
 libkmod_internal = static_library(
@@ -401,17 +415,8 @@ kmod = executable(
 )
 
 _kmod_variables = [
-  'sysconfdir=' + sysconfdir,
-  'distconfdir=' + distconfdir,
   'module_directory=' + moduledir,
-  'module_signatures=' + module_signatures,
 ]
-
-# XXX: Support for empty variables was added in meson v1.4.0.
-# pkgconf behaves identically on missing and empty variable.
-if module_compressions != ''
-  _kmod_variables += ['module_compressions=' + module_compressions]
-endif
 
 pkg.generate(
   name : 'kmod',
--- a/meson.build
+++ b/meson.build
@@ -414,17 +414,6 @@ kmod = executable(
   install : get_option('tools'),
 )
 
-_kmod_variables = [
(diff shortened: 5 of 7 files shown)


hramrach:
People depending on kmod.pc would be mainly people using a Linux kernel patch that was not merged. For compatibility with that the patch that actually removes kmod.pc is separate so that they can not apply/revert the patch and get the new feature at the same time.

If there is a problem with the option name feel free to suggest a better one.

There are 'getter' options, at least version and help.

Additional references would be #100 and https://bugs.gentoo.org/926431 (I did not find corresponding one on github).

hramrach:
Yes, you can probably detect some of these by compiling C tests, that's not necessarily a good interface for knowing which features are compiled into libkmod.

hramrach:
Yes, the libkmod being linked statically is a good point, while kmod links to some libraries dynamically libkmod is not one of them so all of these values are independent between kmod and libkmod.

hramrach:
Nonetheless, since e4130a46dd315310eed37c6bb59ca047e849dd43 kmod.pc is not fully usable.

Since this change was accepted, and the revert of it was not included I think the file should be dropped, anyway.

lucasdemarchi:
Let's leave this open as an option if the kernel patch doesn't work for just using pkg-config.

hramrach:
At this point kmod.pc is broken because it's installed in the wrong location.

Either the move should be reverted or the file removed.

Given that the move was submitted, merged, the revert not merged my impression is that the kmod.pc file is too difficult to understand and should be replaced.

hramrach:
> Out of curiosity: can anyone link the ML discussion, I looked at linux-modules + linux-kbuild yet couldn't find it.

Which discussion?

hramrach:
Any decision on how to resolve the kmod.pc bad location?

lucasdemarchi:
@hramrach  I have a patch to the kernel that I have to clean up and send to actually use it where it is and don't see a use case in which this is important. Could you clarify with a practical example?

lucasdemarchi:
> Technically it's possible to have kmod and libkmod with mismatched
> values due to them being built separately. Also on a binary OS
> distribution the libkmod.pc would be installed only with libkmod
> development headers, not the tool.

this can actually be solved in any of these ways:  a) copy the value in libkmod.pc or b) install kmod.pc regardless since it also contains other "project information". No need to burn the castle down.

hramrach:
> Also this file was initially introduced, then moved to a different
> loaction because its purpose is misunderstood. Using it is awkward
> because the binary search path and pkgconfig path must be matched in
> case there is a possibility that multiple builds of kmod exist.

> As to why multiple build of kmod might exist on a system
> 
>  - custom test build
>  - build of kbuild that matches some target system which has different
>    settings for sysconfdit, distconfdir, or module_directory
>  - per-architecture tools on 'multiarch' system that is set up for
>    building binaries/boot images for arbitrary architectures
>    - when these use a filesystem scheme in which per-architecture
>      libraries and modules are stored in per-architecture directories
>      Linux sticks out like a sore thumb with all modules in /lib
>

lucasdemarchi:
in all those cases, you should have a pkg-config that matches the kmod/libkmod you want to use. If you are passing `pkg-config ... kmod` or `pkg-config ... libkmod` is not very important, but as I said we we can have the same info in both places (although possible, it'd be very wrong to have a config for depmod not matching the lib one).

hramrach:
The complaint about jq was that it's an additional tool needed for building the kernel.

That would be the same with libkmod, it was not need until now, only depmod was needed.

Using pkgconfig means that the pkgconfig path and binary path can be mismatched pointing to a different build of kmod. That becomes even worse when libkmod.pc and kmod.pc are in different locations.

Also alternative implementations might have hard time providing a pkgconfig file while program arguments are something that they have to implement for their version of depmod to work.

Using jq or pkgconfig in the first place is the result of overenginnering this feature. Since using mismatched libkmod and kmod is considered 'very wrong' any additional informational variables can be added to libkmod.pc. With that only one variable remains that is _really_ needed for making use of depmod, and that can be handled by an additional program argument as done in this PR.

It has been pointed out that as implemented the short option name is easy to confuse. AFAICT the option parsing logic does need a short option name, and it cannot be set to something like  ℳ to make it unlikely to be typed by the user. If you have better alternatives for short name please share. Also note that -m and -M take different number of arguments so while easy to confuse they are not easy to replace in the depmod invocation.

lucasdemarchi:
Let me try to be clear this time: I'm still very against this option. If you have a mismatching pkg-config you have to fix that problem.

Until now we never required depmod from the target and will not require it because of this path.

hramrach:
It's not required to have target-specific kmod. However, as already said, it is a way to manage target-specific settings when building for many targets aligned with how other target-specific tools work. If we don't make it possible to have target-specific kmod then kmod is that one special tool that needs to be handled differently, and with current situation of kmod.pc being 'platform independent data' that is the case.

hramrach:
> If you are providing it for building separate distros/configuration/arch, then you should point it to the pkg-config of that distro/configuration/arch rather than pointing to the depmod of that distro/configuration/arch..... PKG_CONFIG_SYSROOT_DIR / PKG_CONFIG_PATH / PKG_CONFIG_LIBDIR are your friends for this use case.

but kmod.pc is 'architecture independent data', can't do.

And exactly because people can't get that I want to get rid of it.

It's not worth explaining 100 times that it's broken.

lucasdemarchi:
> but kmod.pc is 'architecture independent data', can't do.

I already said replicating that information in libkmod.pc is fine.

> And exactly because people can't get that I want to get rid of it.
> 
> It's not worth explaining 100 times that it's broken.

Maybe because you keep explaining what was not asked for. I still don't know what distro / tools / concrete-example you are trying to support even if I asked for it multiple times.

hramrach:
> I already said replicating that information in libkmod.pc is fine.

It's not fine. libkmod.pc was not required before.

> I still don't know what distro / tools / concrete-example you are trying to support even if I asked for it multiple times.

Suppose a situation that you have libraries/modules installed in /usr/lib-$target, and binaries /usr/bin/$target-$tool.

With kmod.pc as 'architecture independent data' there is no way to support this.

hramrach:
Also if you rely on libkmod for this information then if busybox wants to support this it needs to provide libkmod.pc when it does not provide a libkmod.

The kernel would need to pick which file it uses even if the MODULE_DIRECTORY is provided by both kmod.pc and libkmod.pc.

hramrach:
ping?

PR #129, opened 2024-09-13
man: add missing trailing slash in outdir example

evelikov:
Fixes: e2536ab ("man: Provide examples for paths")

### DIFF
--- a/man/depmod.8.scd
+++ b/man/depmod.8.scd
@@ -63,7 +63,7 @@ rather than the current kernel version (as returned by *uname -r*).
 		depmod -b /my/build/staging/dir/
 
 	This expects all input files under
-	_/my/build/staging/dir/@MODULE_DIRECTORY@/$(uname -r)_ and generates
+	_/my/build/staging/dir@MODULE_DIRECTORY@/$(uname -r)_ and generates
 	index files under that same directory.
 
 *-m* _moduledir_


man/depmod.8.scd
```diff
@@ -93,7 +93,7 @@ rather than the current kernel version (as returned by *uname -r*).
 
 	This expects all input files under
 	_@MODULE_DIRECTORY@/$(uname -r)_ and generates index files under
-	_/my/build/staging/dir@MODULE_DIRECTORY@/$(uname -r)_.
+	_/my/build/staging/dir/@MODULE_DIRECTORY@/$(uname -r)_.
```
lucasdemarchi, line 96: no, otherwise you'd have a double // here by default. In fact, it's the first example that is wrong, not this one.

```
-b basedir --basedir basedir
...
This expects all input files under /my/build/staging/dir//lib/modules/$(uname -r) and generates index files under that same directory.
...
-m moduledir --moduledir moduledir
...
This expects all input files under /tmp/build/kernel-modules/$(uname -r) and generates index files under that same directory.
...
-o outdir --outdir outdir
...
This expects all input files under /lib/modules/$(uname -r) and generates index files under /my/build/staging/dir/lib/modules/$(uname -r).
```

lucasdemarchi (changes requested):

lucasdemarchi:
Applied, thanks.

PR #130, opened 2024-09-13
depmod: Remove NOFAIL usage

stoeckmann:
Split off from https://github.com/kmod-project/kmod/pull/123

Removes the unimplemented NOFAIL macro from depmod and adds another strdup NULL check.


lucasdemarchi:
Not a fan of overriding the libc functions like that. Let's go with this version. Thanks

PR #131, opened 2024-09-14
Remove LD_PRELOAD wrappers, properly guard the rest

evelikov:
Namely: *lstat* and access - neither of which was ever used by kmod/libkmod. The latter was used solely in one sub-test - access_and_stat - where stat alone is sufficient.

In addition, we have proper checks in the build systems for the other symbols (effectively removing the `__GLIBC__` hack). See the individual commits for details.

Mind you: commit messages might be a bit much, so any specific suggestions are greatly appreciated.

Edit: add the post removal bits

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,6 +40,7 @@ AC_PROG_CC_C99
 #####################################################################
 
 AC_CHECK_FUNCS_ONCE(__xstat)
+AC_CHECK_FUNCS_ONCE([open64 stat64 fopen64])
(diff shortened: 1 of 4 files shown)


lucasdemarchi:
what happens if the libc implementation of stat is a macro that redirects to lstat? Wouldn't it be a reasonable implementation giving its differences?

lucasdemarchi:
Applied, thanks.

PR #132, opened 2024-09-14
Prep and add CodeQL action

evelikov:
Minor prep + bugfix alongside adding the action itself. It flags some ~90 issues atm, which we'll go through as time permits.

At a glance there is nothing too critical in there, but who knows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,18 +60,7 @@ jobs:
           # doesn't quite work. Perhaps our double/sparse checkout is to blame?
           git config --global --add safe.directory '*'
 
-          for moddir in /usr/lib/modules /lib/modules; do
-            if [ -e "$moddir" ]; then
-              kernel=$(find "$moddir" -maxdepth 1 -mindepth 1 -type d -exec basename {} \;)
-              break
-            fi
-          done
-
-          # There should be one entry - the kernel we installed
-          if [ $(echo "$kernel" | wc -l) -ne 1 ]; then
-            echo >&2 "Error: exactly one kernel must be installed"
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -60,18 +60,7 @@ jobs:
...
-              break
-            fi
-          done
-
-          # There should be one entry - the kernel we installed
-          if [ $(echo "$kernel" | wc -l) -ne 1 ]; then
-            echo >&2 "Error: exactly one kernel must be installed"
-          fi
-          echo "KDIR=$moddir/$kernel/build" >> "$GITHUB_ENV"
+          scripts/print-kdir.sh >> "$GITHUB_ENV"
```
lucasdemarchi, line 63: like the ./.github/actions/setup-ubuntu ... I  think we should keep this in .github/ as I don't want `print-kdir.sh` being used outside of this scope... `./.github/print-kdir.sh` would be ok

Makefile.am
```diff
@@ -23,6 +23,7 @@ EXTRA_DIST += \
 	meson_options.txt \
 	testsuite/meson.build \
 	scripts/build-scdoc.sh \
+	scripts/print-kdir.sh \
```
lucasdemarchi, line 26: not needed

lucasdemarchi (changes requested):

.github/workflows/main.yml
lucasdemarchi: That is a simplification for a container environment that relies on having just one kernel installed.

lucasdemarchi:
Applied, thanks

PR #137, opened 2024-09-17
libkmod: Fix memory leaks on error paths

stoeckmann:
Free previously allocated memory if yet another memory allocation fails.

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -135,8 +135,10 @@ static int kmod_config_add_command(struct kmod_config *config,
 	memcpy(cmd->command, command, commandlen);
 
 	l = kmod_list_append(*list, cmd);
-	if (!l)
+	if (!l) {
+		free(cmd);
 		return -ENOMEM;
+	}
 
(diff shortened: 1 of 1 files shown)


libkmod/libkmod-config.c
```diff
@@ -135,8 +135,10 @@ static int kmod_config_add_command(struct kmod_config *config,
 	memcpy(cmd->command, command, commandlen);
 
 	l = kmod_list_append(*list, cmd);
-	if (!l)
+	if (!l) {
+		free(cmd);
```
lucasdemarchi, line 139: nops, this would be a double free. See the use of _cleanup_free_ above:

```
	_cleanup_free_ struct kmod_command *cmd;
```

lucasdemarchi (commented):
All of those seem to already be correct and adding the free() call would actually turn them into double frees

libkmod/libkmod-config.c
stoeckmann: Oops, thanks for checking!

PR #138, opened 2024-09-17
{ls,ins,rm}mod: update help, man and {bash,fish,zsh} shell completion

evelikov:
As the title says, this series focuses on lsmod/rmmod/insmod.

In general, we've ensured they're consistent among other another - syslog/verbose/version/help for all - aligning the help and manual pages.

Some _potentially_ picky bits:
 - removed unused and undocumented since day 1 - `rmmod` `-p` and `-s`
 - added longopt and documentation for `insmod` `-f`
 
Input welcome:
 - the rmmod/insmod force messages (in help and shell completion) should be trimmed and made scarier IMHO

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -18,6 +18,18 @@ option(
   description : 'Bash completions directory. Use "no" to disable.',
 )
 
+option(
+  'fishcompletiondir',
+  type : 'string',
+  description : 'Fish completions directory. Use "no" to disable.',
+)
+
+option(
+  'zshcompletiondir',
+  type : 'string',
+  description : 'Zsh completions directory. Use "no" to disable.',
+)
+
 # Compression options
 option(
   'zstd',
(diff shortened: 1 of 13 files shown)


lucasdemarchi:
> Some _potentially_ picky bits:
> 
> * removed unused and undocumented since day 1 - `rmmod` `-p` and `-s`
> * added longopt and documentation for `insmod` `-f`

agree

> 
> Input welcome:
> 
> * the rmmod/insmod force messages (in help and shell completion) should be trimmed and made scarier IMHO

agree

lucasdemarchi:
> Would love to hear on the fiddly bits mentioned above and generally if this is something we'd want before polishing it.

yes... this all looks good to me. Thanks

meson_options.txt
```diff
@@ -18,6 +18,18 @@ option(
...
+  type : 'string',
+  description : 'Fish completions directory. Use "no" to disable.',
+)
+
+option(
+  'zshcompletiondir',
+  type : 'string',
+  description : 'Zsh completions directory. Use "no" to disable.',
+)
+
```
lucasdemarchi, line 32: next release will still include autotools. Shouldn't we also add these changes there?

lucasdemarchi:
I pushed up the patches that add shell completion. I think we also need to integrate with autotools before merging the rest.

lucasdemarchi:
Applied, thanks.

PR #140, opened 2024-09-18
libkmod: Avoid misaligned memory accesses

stoeckmann:
It cannot be guaranteed that `mem + size` is always propery aligned, because these values are retrieved from a file.

If the architecture needs proper alignments, this leads to issues.

Proof of Concept:

1. Compile kmod with UBSAN
2. Run test-modinfo testsuite (probably requires ASAN_OPTIONS)
```
ASAN_OPTIONS=verify_asan_link_order=0 ./testsuite/test-modinfo
```

You can see lines like these:
```
../libkmod/libkmod-signature.c:287:41: runtime error: member access within misaligned address 0x7fb6ea48a149 for type 'const struct module_signature', which requires 4 byte alignment
0x7fb6ea48a149: note: pointer points here
 91 0d 1d  ad 00 00 02 00 00 00 00  00 00 00 02 a9 7e 4d 6f  64 75 6c 65 20 73 69 67  6e 61 74 75 72
              ^
```

### DIFF
--- a/libkmod/libkmod-signature.c
+++ b/libkmod/libkmod-signature.c
@@ -307,7 +307,7 @@ bool kmod_module_signature_info(const struct kmod_file *file, struct kmod_signat
 {
 	const char *mem;
 	off_t size;
-	const struct module_signature *modsig;
+	struct module_signature modsig;
 	size_t sig_len;
 
 	size = kmod_file_get_size(file);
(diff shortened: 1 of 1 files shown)


stoeckmann:
> P.S. Can I interest you in adding (address/ub/memory/leak) sanitizers to the CI... Or is it still very noisy?

Thanks for already addressing more parts. I knew about the list memory leaks, but didn't address them yet. So, to answer you anyway: It's on my list, especially figuring out why ASAN is unhappy currently with library link ordering.

If this is fixed, and your changes are in, ASAN should be ready.

libkmod/libkmod-signature.c
```diff
@@ -321,21 +321,21 @@ bool kmod_module_signature_info(const struct kmod_file *file, struct kmod_signat
...
-	if (modsig->algo >= PKEY_ALGO__LAST ||
-			modsig->hash >= PKEY_HASH__LAST ||
-			modsig->id_type >= PKEY_ID_TYPE__LAST)
+	memcpy(&modsig, mem + size, sizeof(struct module_signature));
+	if (modsig.algo >= PKEY_ALGO__LAST ||
+			modsig.hash >= PKEY_HASH__LAST ||
+			modsig.id_type >= PKEY_ID_TYPE__LAST)
 		return false;
-	sig_len = be32toh(get_unaligned(&modsig->sig_len));
+	sig_len = be32toh(get_unaligned(&modsig.sig_len));
```
lucasdemarchi, line 329: this get_unaligned() here doesn't make sense anymore after you copied the struct.

lucasdemarchi (changes requested):
LGTM, but I think we can drop the get_unaligned() too.

libkmod/libkmod-signature.c
lucasdemarchi: on a second thought this looks more a false positive. The alignment it's complaining is for the pointer... but it doesn't make much sense. The only field where the alignment matters is already taken care by using get_unaligned().

stoeckmann: It's a funny false positive.

I've debugged modinfo on my x86_64 system and in fact, with default compiler options and at least gcc 14.2.0 the unaligned access code is optimized away. That's a proper optimization, since x86_64 has no problem with unaligned access.

But ASAN checks the optimized code and warns that unaligned access occurs. Which is objectively correct. It's just not an issue on my system.

So yes, I drop this PR because everything works as intended. I wondered anyway why this was never spotted by anyone over around 10 years. These rather uncommon architectures cannot be THAT uncommon.

lucasdemarchi: I'm not really against applying it if it allows us to enable the sanitizer in future. With the right commit message, we can apply it.

stoeckmann:
I have force-pushed to the branch before re-opening the PR. And now it's not possible for me to re-open it. Follow up is: https://github.com/kmod-project/kmod/pull/147

Sorry for the inconvenience.

thesamesam:
> Run test-modinfo testsuite (probably requires ASAN_OPTIONS)

UBSAN is what's complaining here, not ASAN (hence UBSAN_OPTIONS). Anyway, see e.g. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114217.

PR #141, opened 2024-09-18
testsuite: Fix init_module

stoeckmann:
Apply same logic as in delete_module.c, i.e. pass a reference to a pointer instead of the pointer value, otherwise we cannot update the linked list and added entries are lost.

Spotted with ASAN while running testsuite.


lucasdemarchi:
Applied, thanks.

PR #142, opened 2024-09-19
libkmod: use bufferless zstd decompression

evelikov:
Unlike the other two decompressions, zstd supports streaming bufferless mode. Meaning we don't need to read and realloc in a loop.

Some strace numbers:

        $ strace -e read ./before/depmod -o /tmp/throaway | wc -l
        35265
        $ strace -e fstat ./before/depmod -o /tmp/throaway | wc -l
        1110

        $ strace -e read ./after/depmod -o /tmp/throaway | wc -l
        5677
        $ strace -e fstat ./after/depmod -o /tmp/throaway | wc -l
        6642

.. and valgrind total heap usage statistics:

        before:
        1,039,426 allocs, 1,039,426 frees, 3,679,232,922 bytes allocated

        after:
        1,020,643 allocs, 1,020,643 frees, 1,157,922,357 bytes allocated

The actual runtime is within the error margin.

---

Got inspired by the modinfo (builtin) parser rework :sweat_smile:

### DIFF
--- a/libkmod/libkmod-file-zstd.c
+++ b/libkmod/libkmod-file-zstd.c
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/mman.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
I did some experiments and on my local machine I didn't see much change in runtime. AFAIK the different numbers for memory allocation is just due to the different patterns for doing it in libkmod or relying on the internal buffer from libzma. Anyway, this impl is shorter and we can replace the old one. Applied, thanks

lucasdemarchi:
fwiw by "runtime" I meant the function runtime, using perf to calculate it... Something similar to https://bristot.me/using-perf-probe-to-measure-execution-time-of-user-space-code-on-linux/

PR #143, opened 2024-09-19
testsuite/list: resolve memory leaks

evelikov:
Another one spotted by ASAN.

I'm already looking at the final one - DONE, see https://github.com/kmod-project/kmod/pull/144

/cc @stoeckmann


lucasdemarchi:
Applied, thanks

PR #144, opened 2024-09-19
testsuite: sort modnames only, if available

evelikov:
In some tests we expect zero modules to be loaded. In those cases, skip the sorting - qsort is annotated as non-null(1,2) so we shouldn't provide null as expected/loaded modules.


lucasdemarchi:
Applied, thanks

PR #147, opened 2024-09-23
libkmod: Avoid misaligned memory accesses

stoeckmann:
ASAN incorrectly reports this as unaligned access if compiler optimizes code. Since ASAN support is valued higher than this kind of optimization, just copy these few bytes into a stack variable and avoid false positives in tests.


lucasdemarchi:
This is a replacement for #140

lucasdemarchi:
Applied, thanks.

PR #148, opened 2024-09-23
libkmod: Fix ELFDBG usage

stoeckmann:
If debug is enabled through macro `ENABLE_ELFDBG` in libkmod-elf.c, compilation fails because two callers use wrong arguments.

While at it, fixed a typo.


stoeckmann:
I think that it would make sense to integrate the ELFDBG usage into DBG, so it also takes the log function as specified in context. Not sure why they should be handled so differently compared to other debug messages.

lucasdemarchi:
Applied, thanks.

PR #149, opened 2024-09-23
libkmod: Avoid OOB with huge ELF files

stoeckmann:
On 32 bit systems it is possible to trigger an out of boundary write with excessively huge ELF files.

With modinfo, the calculation of required memory for char pointer vector and strings might overflow, leading to an allocation which is too small. Subsequent memcpy leads to an out of boundary write.

Proof of Concept:
1. Create a module which overflows calculation (715.827.882 "A" strings for modinfo)
```
cat > poc.ko.b64 << EOF
f0VMRgICAAAAAAAAAAAAABERIiIzMzMzRERERERERERVVVVVVVVVVQAAAAAAAABAd3d3d4iImZmq
qgBAAAMAARERERERERERERERERERERERERERERERERERERERERERERERERERERERERERERERERER
EREREREREREREREREREAAAAIERERESIiIiIiIiIiMzMzMzMzMzMAAAAAAAABAAAAAAAAAAALZmZm
Znd3d3eIiIiIiIiIiJmZmZmZmZmZAAAAABEREREiIiIiIiIiIjMzMzMzMzMzAAAAAAAAAQwAAAAA
VVVVVGZmZmZ3d3d3iIiIiIiIiIiZmZmZmZmZmS5tb2RpbmZvAAAAAA==
EOF
base64 -d poc.ko.b64 > poc.ko

cat > poc.c << EOF
#include <stdio.h>

int
main(void)
{
	FILE *fp = fopen("poc.ko", "a");
	for (int i = 0; i < 715827882; i++)
		fwrite("\x41\x00", 2, 1, fp);
	fclose(fp);
	return 0;
}
EOF
cc -o poc poc.c
./poc
```
2. Run modinfo on a 32 bit system
```
modinfo poc.ko
```

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -6,6 +6,7 @@
 #include <assert.h>
 #include <elf.h>
 #include <errno.h>
+#include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -428,6 +429,7 @@ int kmod_elf_get_section(const struct kmod_elf *elf, const char *section,
 int kmod_elf_get_strings(const struct kmod_elf *elf, const char *section, char ***array)
 {
 	size_t i, j, count;
+	size_t vecsz;
 	uint64_t size;
 	const void *buf;
 	const char *strings;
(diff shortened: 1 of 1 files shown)


stoeckmann:
> maybe we can make this a `check_overflow()` inline helper in `shared/` to keep it more legible?

There's alread an `addu64_overflow` inline function in shared/util.h, so that's a start. And it's also used in libkmod-elf.c. We could add more for multiplications etc.

But keep in mind that "newer" gcc/clang versions have things such `__builtin_mul_overflow`. I'm not aware how backwards compatible kmod wants to be, so MAYBE this could be something we can switch to. As a side-note, the shadow project uses them a lot.

lucasdemarchi:
Applied, thanks

PR #150, opened 2024-09-23
ci: do monthly dependabot runs

evelikov:
The CodeQL action is getting 1-2 releases every week with only a handful of changes in each.

This is getting a bit too much, so let's change to monthly checks.

@lucasdemarchi we can try this as an alternative to disabling/removing codeql

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,7 +12,7 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Ok... that would be a little bit saner. Yet, is there a way to tell it to stop rebasing the PR whenever something merges in master branch?

lucasdemarchi:
Great, thanks. Applied.

PR #151, opened 2024-09-23
ci: add more SPDX copyright/license identifiers

evelikov:
This concludes the files, I'm planning to reuse in usbutils so I'll stop before it gets too annoying :-D

---

Ftr I did consider going project wide, although the results are quite drastic (as per `reuse lint`). If you think we could/(should?) continue let me know. Otherwise this will be the final PR on the topic.

```
* Files with copyright information: 90 / 435
* Files with license information: 80 / 435
```

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -1,2 +1,6 @@
+# SPDX-FileCopyrightText: 2024 Emil Velikov <emil.l.velikov@gmail.com>
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
 caf
 parm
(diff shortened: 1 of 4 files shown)


lucasdemarchi:
Humn... do we need these in the simple files?  Why? For the missing ones that make sense we can simply include the license project-wide. Looking at `git ls-files`, there are a lot of files that should **not** have a license. 

With this simple filter:

```
$ git ls-files | grep -v -e '.*\.bin$' | wc -l 
397
```

... we have less files than reported above, so I don't know what `reuse lint` is using.

lucasdemarchi:
Even better approximation of reality:

```
$ git ls-files | grep -v -e '^testsuite/.*[^ch]$' -e '.*\.bin$'  | wc -l 
161
```

lucasdemarchi:
For things that are real source files, edited by humans, I prefer the inline identifiers. For everything else (e.g. all the files in testsuite/rootfs-pristine), we should use another approach

lucasdemarchi:
Applied, thanks

PR #152, opened 2024-09-24
libkmod: Improve strbuf handling in libkmod-index.c

stoeckmann:
- Turn `unsigned int` lengths into `size_t` to stay in sync with latest strbuf changes
- Check strbuf function return values, because they might fail. If they fail, push and pop operations might run out of sync


stoeckmann:
... and fixed a typo while at it.

lucasdemarchi:
Applied, thanks.

PR #153, opened 2024-09-24
libkmod: Plug memory leak on error path

stoeckmann:
Fix memory leak on error path introduced with read_u32 error checks.

Leak introduced with: https://github.com/kmod-project/kmod/commit/9d965d3386b609e81155c55ade59fa35ae991edf

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -338,8 +338,10 @@ struct index_file *index_file_open(const char *filename)
 		goto err;
 
 	new->file = file;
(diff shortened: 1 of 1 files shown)


stoeckmann:
> With the sanitizer issues fixed, I think we can override the ASAN variable (to silence the link order messages) and enable it. Do you want to do it, or shall I?

You can do it, although I would like to know why the ordering is off. Haven't been into this issue yet.

lucasdemarchi:
Applied, thanks

PR #154, opened 2024-09-24
modinfo: Avoid unneeded system call

stoeckmann:
If a file name does not end with ".ko", there is no need to call fstat. Since common use cases are like "modinfo modname", the string check should come first to save a tiny amount of system time.

In fact, this would have saved me some time with strace analysis why a local file wasn't actually read. :)

### DIFF
--- a/tools/modinfo.c
+++ b/tools/modinfo.c
@@ -350,8 +350,8 @@ static bool is_module_filename(const char *name)
 {
 	struct stat st;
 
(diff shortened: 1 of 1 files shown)


stoeckmann:
I think the git commit message is a good location, because if this is ever changed again, there must have been a compelling reason to touch this specific line of code. If not, git blame (hopefully run to figure out why people decided to design the code the way it is) would explain it.

lucasdemarchi:
Applied, thanks

PR #157, opened 2024-09-27
testsuite: Fix test-weakdep with default autoconf settings

stoeckmann:
This breaks test-weakdep:

```
./configure
make
make check
```

The default should work, but it sets SYSCONFDIR to /usr/etc, which probably nobody really uses. Anyway, the test-suite should support configurations kmod allows.


lucasdemarchi:
Applied, thanks.

PR #159, opened 2024-09-27
shared: Fix EOF handling in freadline_wrapped

stoeckmann:
If a file ends with a backslash without a newline, freadline_wrapped adds EOF to the actual string.


lucasdemarchi:
Applied, thanks

PR #160, opened 2024-09-28
ci: enable sanitizers in the meson build

evelikov:
... to prevent a range of issues creeping in.

Currently the sanitizers flag ordering issue(s), that we need to explicitly disable. Thus the sanitizers are only enabled in CI and not for all developer builds (aka build-dev.ini).


lucasdemarchi:
Applied, thanks.

PR #166, opened 2024-09-30
depmod: Actually use scratch buffer memory

stoeckmann:
Accessing the backing "alias" char buffer directly works only as long as all symbols are smaller than 1024, because otherwise the scratch buffer allocates memory and the addresses do not match anymore.

Proof of Concept:
1. Create a temporary modules directory containing a module with two symbols, `aa` and `1025*A` (i.e. longer than 1024 characters)
```
MYDIR=$(mktemp -d)
KDIR=$MYDIR/lib/modules/$(uname -r)
mkdir -p $KDIR/kernel/sub
cat > $MYDIR/mod.ko.xz.b64 << EOF
/Td6WFoAAATm1rRGBMBvmQohARwAAAAAAAAAAMNcVP3gBRgAZ10AP5FFhGhEVEmcAlZtZy+BVw2o
524UVW19oueqRkgFvigdxpHbOYyTU7Kx/hlefDfD+r/tFCQlBcHuPRCbnxuRTPRldwa2DYFC2g0c
DdY1yxHvF6tv9kbWqc8xpwpQ8YWKVB8MXvUAAAAAPR9nVoEFabMAAYsBmQoAAI4ynUexxGf7AgAA
AAAEWVo=
EOF
base64 -d $MYDIR/mod.ko.xz.b64 | xz -cd > $KDIR/kernel/sub/mod.ko
```
2. Run depmod with warnings enabled
```
depmod -b $MYDIR -o $MYDIR/out -w
```

You can see, among other warnings, this one:
```
depmod: WARNING: duplicate module syms:
symbol:aa mod
```

Since the `1025*A` symbol name triggered an allocation within the scratch buffer, the first symbol name `aa` has been added twice. The `1025*A` symbol name can be seen in written `modules.symbols` file, but not in `modules.symbols.bin` (because the scratch buffer issue is triggered in second output).

### DIFF
--- a/tools/depmod.c
+++ b/tools/depmod.c
@@ -2333,6 +2333,7 @@ static int output_symbols_bin(struct depmod *depmod, FILE *out)
 		int duplicate;
 		const struct symbol *sym = v;
 		size_t len;
+		char *s;
 
 		if (sym->owner == NULL)
 			continue;
@@ -2343,12 +2344,12 @@ static int output_symbols_bin(struct depmod *depmod, FILE *out)
 			ret = -ENOMEM;
 			goto err_scratchbuf;
 		}
-		memcpy(scratchbuf_str(&salias) + baselen, sym->name, len + 1);
-		duplicate =
-			index_insert(idx, alias, sym->owner->modname, sym->owner->idx);
+		s = scratchbuf_str(&salias);
+		memcpy(s + baselen, sym->name, len + 1);
+		duplicate = index_insert(idx, s, sym->owner->modname, sym->owner->idx);
 
 		if (duplicate && depmod->cfg->warn_dups)
-			WRN("duplicate module syms:\n%s %s\n", alias, sym->owner->modname);
+			WRN("duplicate module syms:\n%s %s\n", s, sym->owner->modname);
 	}
 
 	index_write(idx, out);
--- a/tools/depmod.c
+++ b/tools/depmod.c
@@ -2310,10 +2310,10 @@ static int output_symbols(struct depmod *depmod, FILE *out)
 static int output_symbols_bin(struct depmod *depmod, FILE *out)
 {
 	struct index_node *idx;
-	char alias[1024];
+	char aliasbuf[1024] = "symbol:";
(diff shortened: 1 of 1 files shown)


tools/depmod.c
```diff
@@ -2343,12 +2344,12 @@ static int output_symbols_bin(struct depmod *depmod, FILE *out)
 			ret = -ENOMEM;
 			goto err_scratchbuf;
 		}
-		memcpy(scratchbuf_str(&salias) + baselen, sym->name, len + 1);
-		duplicate =
-			index_insert(idx, alias, sym->owner->modname, sym->owner->idx);
+		s = scratchbuf_str(&salias);
```
lucasdemarchi, line 2347: ugh... good find. As a follow up commit, can you rename alias to something else like `aliasbuf`  so it doesn't get misused?

lucasdemarchi (commented):
LGTM.

On a related note since you are reviewing these things. I was thinking of blending strbuf and scratchbuf, always allowing to start with a stack buffer and if NULL is passed, then the current behavior for strbuf is retained. Any thoughts?

stoeckmann:
> On a related note since you are reviewing these things. I was thinking of blending strbuf and scratchbuf, always allowing to start with a stack buffer and if NULL is passed, then the current behavior for strbuf is retained. Any thoughts?

I've had an idea like that as well, also possibly backing array creations with scratchbuf. The concept of scratchbuf is awesome, since a lot of allocations could be skipped when it comes to heuristics and common module setups in distributions.

But I didn't make any progress into that direction, so: Yes I would like to see that, but have no code to share. :D

tools/depmod.c
stoeckmann: Adjusted. While at it, inititalized the buffer directly without the need to use memcpy.

lucasdemarchi:
> But I didn't make any progress into that direction, so: Yes I would like to see that, but have no code to share. :D

I started doing that and left saved on a branch... I will try to finish it and submit as a PR

lucasdemarchi:
Applied, thanks.

PR #168, opened 2024-09-30
libkmod: Avoid redundant kmod_pool_get_module call

stoeckmann:
Before kmod_module_new_from_path creates a module with kmod_module_new, it checks with kmod_pool_get_module if a module with given name already exists.

This check can be removed, because kmod_module_new does the same. If the module already existed, the same checks are performed as in current code for the "if (m != NULL)" case. If it did not exist yet, m->path is NULL and the first if-block is entered.

And since kmod_module_new takes care of incrementing the reference counter, the explicit call of kmod_module_ref can be removed from kmod_module_new_from_path as well.


lucasdemarchi:
Applied, thanks.

Issue #162, opened 2024-09-29
Add and use move overflow helpers


evelikov:
@stoeckmann has been diligently going through the code base adding overflow checks. Overall it would be cleaner if we add new variants akin to `addu64_overflow` instead of having them open-coded.

Ideally we'll check and reuse compiler built-ins, where possible. They can be made mandatory with the impact clearly documented in the commit messages - which gcc/clang version introduced them, which distros use those as base.

See https://github.com/kmod-project/kmod/pull/149#issuecomment-2380395360 for some more context and examples.

stoeckmann:
I think this issue is resolved, or do you expect to add more overflow helpers?

PR #169, opened 2024-09-30
Add size_t overflow helpers and use them

evelikov:
This builds on the existing add_u64 helper, adding add_u32 and mul_{u32,u64} variants. Where those are used for the actual size_t ones.

Names are likely suboptimal, so any suggestions are highly appreciated. 

Note: the autotools build explicitly does not check for the extra built-ins since a) it's going away soon (tm) and b) serves as nice exercise that the fallback is correct.

Closes: https://github.com/kmod-project/kmod/issues/162

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -243,7 +243,7 @@ static inline int elf_get_section_info(const struct kmod_elf *elf, uint16_t idx,
 	}
 #undef READV
 
-	if (addu64_overflow(*offset, *size, &min_size) || min_size > elf->size) {
+	if (uadd64_overflow(*offset, *size, &min_size) || min_size > elf->size) {
 		ELFDBG(elf, "out-of-bounds: %" PRIu64 " >= %" PRIu64 " (ELF size)\n",
 		       min_size, elf->size);
 		return -EINVAL;
@@ -330,7 +330,7 @@ struct kmod_elf *kmod_elf_new(const void *memory, off_t size)
 		goto invalid;
 	}
 	shdrs_size = shdr_size * elf->header.section.count;
-	if (addu64_overflow(shdrs_size, elf->header.section.offset, &min_size) ||
+	if (uadd64_overflow(shdrs_size, elf->header.section.offset, &min_size) ||
 	    min_size > elf->size) {
 		ELFDBG(elf, "file is too short to hold sections\n");
 		goto invalid;
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -429,7 +429,7 @@ int kmod_elf_get_section(const struct kmod_elf *elf, const char *section,
 int kmod_elf_get_strings(const struct kmod_elf *elf, const char *section, char ***array)
 {
 	size_t i, j, count;
-	size_t vecsz;
+	size_t tmp_size, vec_size, total_size;
 	uint64_t size;
 	const void *buf;
 	const char *strings;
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -49,8 +49,12 @@ endforeach
 _builtins = [
   ['__builtin_clz', '0', true],
   ['__builtin_types_compatible_p', 'int, int', true],
+  ['__builtin_uadd_overflow', '0U, 0U, (void*)0', false],
   ['__builtin_uaddl_overflow', '0UL, 0UL, (void*)0', false],
   ['__builtin_uaddll_overflow', '0ULL, 0ULL, (void*)0', false],
+  ['__builtin_umul_overflow', '0U, 0U, (void*)0', false],
+  ['__builtin_umull_overflow', '0UL, 0UL, (void*)0', false],
+  ['__builtin_umulll_overflow', '0ULL, 0ULL, (void*)0', false],
```
stoeckmann, line 57: Adding an inline comment as a suggestion for names: We could just stick with this and skip __builtin, i.e.
- `umul64_overflow` for unsigned multiplication of 64 bit values
- `umulsz_overflow` for size_t
And same goes for the additions
- `uadd64_overflow`
- `uaddsz_overflow`

shared/util.h
```diff
@@ -92,7 +92,21 @@ static inline void freep(void *p)
...
 
-static inline bool addu64_overflow(uint64_t a, uint64_t b, uint64_t *res)
+static inline bool add_u32_overflow(uint32_t a, uint32_t b, uint32_t *res)
+{
+#if HAVE___BUILTIN_UADD_OVERFLOW
+#if __SIZEOF_INT__ == 4
+	return __builtin_uadd_overflow(a, b, res);
+#else
+#error "sizeof(int) != 4"
+#endif
```
stoeckmann, line 102: You could just remove the else-error-block here. It would fall through to the current code.

shared/util.h
```diff
@@ -102,7 +116,60 @@ static inline bool addu64_overflow(uint64_t a, uint64_t b, uint64_t *res)
...
+}
+
+static inline bool mul_u32_overflow(uint32_t a, uint32_t b, uint32_t *res)
+{
+#if HAVE___BUILTIN_UMUL_OVERFLOW
+#if __SIZEOF_INT__ == 4
+	return __builtin_umul_overflow(a, b, res);
+#else
+#error "sizeof(int) != 4"
+#endif
```
stoeckmann, line 143: You can drop this check here as well.

shared/util.h
```diff
@@ -102,7 +116,60 @@ static inline bool addu64_overflow(uint64_t a, uint64_t b, uint64_t *res)
...
+static inline bool mul_u64_overflow(uint64_t a, uint64_t b, uint64_t *res)
+{
+#if (HAVE___BUILTIN_UMULL_OVERFLOW && HAVE___BUILTIN_UMULLL_OVERFLOW)
+#if __SIZEOF_LONG__ == 8
+	return __builtin_umull_overflow(a, b, res);
+#elif __SIZEOF_LONG_LONG__ == 8
+	return __builtin_umulll_overflow(a, b, res);
+#else
+#error "sizeof(long long) != 8"
+#endif
```
stoeckmann, line 159: And this check can be removed as well.

libkmod/libkmod-elf.c
```diff
@@ -470,13 +470,16 @@ int kmod_elf_get_strings(const struct kmod_elf *elf, const char *section, char *
 	if (strings[i - 1] != '\0')
 		count++;
 
+	/* (string vector + NULL) * sizeof(char *) + size + extra NULL */
 	/* make sure that vector and strings fit into memory constraints */
```
stoeckmann, line 474: The second comment black can be removed as well, just as in libkmod-builtin.c

libkmod/libkmod-elf.c
```diff
@@ -470,13 +470,16 @@ int kmod_elf_get_strings(const struct kmod_elf *elf, const char *section, char *
 	if (strings[i - 1] != '\0')
 		count++;
 
+	/* (string vector + NULL) * sizeof(char *) + size + extra NULL */
```
stoeckmann, line 473: the last "extra NULL" should be NUL to highlight that we are talking about '\0' not (void *)0.

shared/util.h
```diff
@@ -92,7 +92,21 @@ static inline void freep(void *p)
...
+#if HAVE___BUILTIN_UADD_OVERFLOW
+#if __SIZEOF_INT__ == 4
+	return __builtin_uadd_overflow(a, b, res);
+#else
+#error "sizeof(int) != 4"
+#endif
+#else
+	*res = a + b;
+	return UINT32_MAX - a < b;
+#endif
```
stoeckmann, line 106: Edit: You can skip the error block if you &&-connect the SIZEOF_INT check with HAVE__BUILTIN_UADD_OVERFLOW. Same goes for the other checks where manual implementations are possible

stoeckmann: Didn't even think about the elif, but yes, that would add a lot of portability.

libkmod/libkmod-elf.c
stoeckmann: NUL is a single byte. NULL is a pointer, i.e. 4 or 8 bytes.

stoeckmann (approved):

lucasdemarchi:
Small fixup in one commit message. Applied, thanks.

PR #171, opened 2024-10-03
libkmod: unref in  kmod_module_new_from_path() error path

evelikov:
Preserve the original behaviour, where the module refcount stays the same, if the function fails.

Fixes: e19338da ("libkmod: Avoid redundant kmod_pool_get_module call")

---

Follow-up to https://github.com/kmod-project/kmod/pull/168 ... not sure how I missed it :sweat_smile:


lucasdemarchi:
Applied, thanks

PR #172, opened 2024-10-03
Clang inspired fixes, add clang CI permutation

evelikov:
This series fixes all the issues spotted by clang (+ be that clang itself, scan-build or cland-tidy) and adds it to CI.

At a glance - from the 5 distros we have 4 major clang versions and as indicated in the patches, there's some value in having them.

Understandably, the CI matrix is becoming a bit of a zoo... Although half of it will be gone with the drop of autotools, soon (tm) :sweat_smile:

Plus since they're all parallel jobs, I think we can bare with it.


tools/insmod.c
```diff
@@ -161,7 +161,7 @@ static int do_insmod(int argc, char *argv[])
 	free(opts);
 
 	log_close();
-	return err >= 0 ? EXIT_SUCCESS : EXIT_FAILURE;
+	return r == 0 ? EXIT_SUCCESS : EXIT_FAILURE;
```
lucasdemarchi, line 164: why do we have both err and r as control here? Just get rid of `err` and use `r` for the returns above?

libkmod/libkmod-index.c
```diff
@@ -682,7 +682,7 @@ static struct index_mm_node *index_mm_read_node(struct index_mm *idx, uint32_t o
 	const char *prefix;
 	int i, child_count, value_count, children_padding;
 	uint32_t children[INDEX_CHILDMAX];
-	char first, last;
+	uint8_t first, last;
```
lucasdemarchi, line 685: needs a rebase

libkmod/libkmod.c
```diff
@@ -802,8 +802,15 @@ KMOD_EXPORT int kmod_dump_index(struct kmod_ctx *ctx, enum kmod_index type, int
 	if (ctx == NULL)
 		return -ENOSYS;
 
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wtautological-unsigned-enum-zero-compare"
+#endif
```
lucasdemarchi, line 808: wondering if we should use a _Pragma so we can hide where it's available in a define somewhere out of sight.

lucasdemarchi: maybe when we have one more case...

lucasdemarchi:
Applied most of the patches, leaving some comments in others. Thanks

libkmod/libkmod-index.c
```diff
@@ -699,7 +699,7 @@ static struct index_mm_node *index_mm_read_node(struct index_mm *idx, uint32_t o
 		first = read_char_mm(&p);
 		last = read_char_mm(&p);
 
-		if (first > last || first < 0 || last < 0)
+		if (first > last || first > SCHAR_MAX || last > SCHAR_MAX)
```
stoeckmann, line 702: While technically correct, changing it to `>= INDEX_CHILDMAX` avoids this very technical description. After all, we have to check if the index into child array is valid. If you adjust it, I just mention here again that it turns the `>`-check into `>=`-check. I've failed spotting this while I later adjusted code due to such requests more than once ... :)

lucasdemarchi:
Is this still a draft?

scripts/sanitizer-env.sh
```diff
@@ -2,7 +2,16 @@
 
 # set -euo pipefail # don't set these, since this script is sourced
 
-OUR_PRELOAD=$(gcc -print-file-name=libasan.so)
+if [[ ${CC-} == *gcc* ]]; then
+    OUR_PRELOAD=$(gcc -print-file-name=libasan.so)
```
lucasdemarchi, line 6: should use "$CC"  like below since you are globbing the gcc match... ?

.github/actions/setup-alpine/action.yml
```diff
@@ -17,6 +17,7 @@ runs:
           automake \
           bash \
           build-base \
+          clang \
```
lucasdemarchi, line 20: commit message:
"Note: Fedora 40 doesn't ship shared the shared sanitizer library, while"

one shared too many

testsuite/testsuite.c
```diff
@@ -258,7 +258,7 @@ static inline int test_run_child(const struct test *t, int fdout[2], int fderr[2
 		}
 
 		if (stat_mstamp(&rootfsst) > stat_mstamp(&stampst)) {
-			ERR("rootfs %s is dirty, please run 'make rootfs' before running this test\n",
```
lucasdemarchi, line 261: I don't think these are equivalent. If I run `meson test`, it will actually run the test, which is not what I want. I think we need to expose a target like `ninja -C build create-rootfs`, so the rootfs is prepared for the test to run.

lucasdemarchi: And the right command would be....

```
meson compile testsuite/create-rootfs
```

lucasdemarchi:
I did a small fixup in testsuite/README and applied. Thanks!

Issue #163, opened 2024-09-29
ENABLE_ELFDBG is fragile


evelikov:
As highlighted in https://github.com/kmod-project/kmod/pull/148 - it broke at some point. We can enable it as part of `--enable-debug`, via some other means or remove it.

If part of `--enable-debug` we should double-check for perf impact and/or syslog flood, especially since some distros use `--enable-debug`. We could ask distros to drop the switch?

This is a bit of open-ended question...

PR #173, opened 2024-10-03
Fix and rework logging

evelikov:
While looking at the ENABLE_ELFDBG I noticed that the meson build is a little broken wrt logging/debug. The first commit fixes that, while the others rework things to be a bit more sane IMHO.

Since the changes are somewhat opinionated, I left them separate. If we decide to have them, we should all together drop the first commit, since it's effectively undone with latter ones.

Still not 100% on board with the last one `ENABLE_ELFDBG`, but I'm adding it as a food for thought ... It cannot land as-is, hence the XXX.
 
---

Closes: https://github.com/kmod-project/kmod/issues/163 ... maybe

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -203,6 +203,8 @@ AC_ARG_ENABLE([logging],
 	[], enable_logging=yes)
 AS_IF([test "x$enable_logging" = "xyes"], [
 	AC_DEFINE(ENABLE_LOGGING, [1], [System logging.])
+], [
+	AC_DEFINE(ENABLE_LOGGING, [0], [System logging.])
 ])
 
 AC_ARG_ENABLE([debug],
--- a/configure.ac
+++ b/configure.ac
@@ -215,6 +215,7 @@ AS_IF([test "x$enable_debug" = "xyes"], [
 ], [
 	AC_DEFINE(ENABLE_DEBUG, [0], [Debug messages.])
 ])
+AC_DEFINE(ENABLE_ELFDBG, [0], [Debug elf parsing messages.])
 
 AC_ARG_ENABLE([coverage],
 	AS_HELP_STRING([--enable-coverage], [enable test coverage @<:@default=disabled@:>@]),
--- a/configure.ac
+++ b/configure.ac
@@ -212,6 +212,8 @@ AC_ARG_ENABLE([debug],
 	[], [enable_debug=no])
 AS_IF([test "x$enable_debug" = "xyes"], [
 	AC_DEFINE(ENABLE_DEBUG, [1], [Debug messages.])
+], [
+	AC_DEFINE(ENABLE_DEBUG, [0], [Debug messages.])
 ])
 
 AC_ARG_ENABLE([coverage],
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -56,7 +56,7 @@ struct kmod_elf {
 
 //#define ENABLE_ELFDBG 1
 
-#if defined(ENABLE_LOGGING) && defined(ENABLE_ELFDBG)
+#if (ENABLE_LOGGING == 1) && defined(ENABLE_ELFDBG)
 #define ELFDBG(elf, ...) _elf_dbg(elf, __FILE__, __LINE__, __func__, __VA_ARGS__);
 
 static inline void _elf_dbg(const struct kmod_elf *elf, const char *fname, unsigned line,
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -54,10 +54,14 @@ struct kmod_elf {
 	} header;
 };
 
+//#undef ENABLE_ELFDBG
 //#define ENABLE_ELFDBG 1
 
-#if (ENABLE_LOGGING == 1) && defined(ENABLE_ELFDBG)
-#define ELFDBG(elf, ...) _elf_dbg(elf, __FILE__, __LINE__, __func__, __VA_ARGS__);
+#define ELFDBG(elf, ...)                                                          \
+	do {                                                                      \
+		if (ENABLE_LOGGING == 1 && ENABLE_ELFDBG == 1)                    \
+			_elf_dbg(elf, __FILE__, __LINE__, __func__, __VA_ARGS__); \
+	} while (0);
 
 static inline void _elf_dbg(const struct kmod_elf *elf, const char *fname, unsigned line,
 			    const char *func, const char *fmt, ...)
@@ -70,9 +74,6 @@ static inline void _elf_dbg(const struct kmod_elf *elf, const char *fname, unsig
 	vfprintf(stderr, fmt, args);
 	va_end(args);
 }
-#else
-#define ELFDBG(elf, ...)
-#endif
 
 static int elf_identify(const void *memory, uint64_t size)
 {
--- a/libkmod/libkmod-internal.h
+++ b/libkmod/libkmod-internal.h
@@ -10,22 +10,15 @@
 
 #include "libkmod.h"
 
-static _always_inline_ _printf_format_(2, 3) void kmod_log_null(const struct kmod_ctx *ctx,
-								const char *format, ...)
-{
-}
-
-#define kmod_log_cond(ctx, prio, arg...)                                          \
-	do {                                                                      \
-		if (ENABLE_LOGGING == 1 && kmod_get_log_priority(ctx) >= prio)    \
-			kmod_log(ctx, prio, __FILE__, __LINE__, __func__, ##arg); \
+#define kmod_log_cond(ctx, prio, arg...)                                           \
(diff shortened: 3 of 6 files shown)


tools/log.c
```diff
@@ -62,19 +62,10 @@ _printf_format_(6, 0) static void log_kmod(void *data, int priority, const char
 		return;
 
 	if (log_use_syslog) {
-#ifdef ENABLE_DEBUG
 		syslog(priority, "%s: %s:%d %s() %s", prioname, file, line, fn, str);
-#else
-		syslog(priority, "%s: %s", prioname, str);
```
lucasdemarchi, line 68: wat? this doesn't match the commit message. Now everybody is logging the file, line, function name.

lucasdemarchi (commented):
Not sure I agree with where this PR is heading. I will review again another day when I'm not sleepy.

libkmod/libkmod-elf.c
```diff
@@ -54,9 +54,7 @@ struct kmod_elf {
 	} header;
 };
 
-//#define ENABLE_ELFDBG 1
-
-#if defined(ENABLE_LOGGING) && defined(ENABLE_ELFDBG)
+#if ENABLE_DEBUG
```
stoeckmann, line 57: Although I've first mentioned that I would prefer to have it enabled by default, I have understood now that it's a massive performance bottle neck. Having it in elf_get_uint breaks some of the optimizations which compilers could do, turning this into billions of calls on a default Arch Linux installation.

So, ENABLE_ELFDBG should stay special ...

stoeckmann: That still makes it special from my point of view. As long as it isn't activated for any of these distribution builds, I'm fine with it.

lucasdemarchi: > So, ENABLE_ELFDBG should stay special ...

correct - and the last time I enabled it for my own tests was maybe 10 years ago. We don't want all those strings to pollute the stderr. Should not not even the binary.... contrary to the debug messages, that is nice to enable from time to time when debugging e.g. libkmod with another tool.

tools/log.c
lucasdemarchi: Humn... tools that want a more sophisticate log are supposed to override the ctx->log_fn via kmod_set_log_fn().

But yes. If this is only about the priority name, I'm fine with it.

configure.ac
```diff
@@ -198,13 +198,6 @@ AS_IF([test "x$enable_manpages" = "xyes"], [
 	])])
 AM_CONDITIONAL([BUILD_MANPAGES], [test "x$enable_manpages" = "xyes"])
 
-AC_ARG_ENABLE([logging],
```
lucasdemarchi, line 201:
```
If one is to switch it off, then the libkmod log API becomes
non-functional. In addition, the tools' syslog facility also does
nothing...
```

which is fine for embedded use cases relying only on return codes.

libkmod/libkmod-internal.h
```diff
@@ -10,22 +10,14 @@
...
-
 #define kmod_log_cond(ctx, prio, arg...)                                          \
 	do {                                                                      \
-		if (kmod_get_log_priority(ctx) >= prio)                           \
+		if ((ENABLE_DEBUG || (!ENABLE_DEBUG && prio != LOG_DEBUG)) &&     \
+		    kmod_get_log_priority(ctx) >= prio)                           \
 			kmod_log(ctx, prio, __FILE__, __LINE__, __func__, ##arg); \
 	} while (0)
 
-#ifdef ENABLE_DEBUG
```
lucasdemarchi, line 25: only because you mandate logging... which means all the (sometimes very verbose strings) will be present in the binary

lucasdemarchi:
Applied the first patch here.

libkmod/libkmod-elf.c
```diff
@@ -54,10 +54,14 @@ struct kmod_elf {
 	} header;
 };
 
+//#undef ENABLE_ELFDBG
 //#define ENABLE_ELFDBG 1
```
stoeckmann, line 58: One of both is enough as a hint ;)

stoeckmann: Not sure if my edit went through before your answer. Why do we need these lines at all? We have configure options now.

configure.ac
lucasdemarchi: I actually don't think there's any good reason to use the busybox variant. And I'm aware of my very biased view :)

If you look at e.g. yocto and buildroot, they both allow you to select kmod.  Android uses kmod (it used not to in ~2014, when I a talk in an Android conf about that)

If your init of choice is systemd, then it's a requirement. A decade ago I was using systemd on embedded systems... Just found a video from one of my previous bosses.... https://www.youtube.com/watch?v=ERS1OSOTGpg&ab_channel=TheLinuxFoundation - you can find a bunch (more recent ones) if you google "embedded" and "systemd".

stoeckmann: To their defense, kmod 30 is listed as latest release on GitHub. All others afterwards are just tags. ;)

Of course this can't be the reason, but is there anything kmod repository can do about?

lucasdemarchi:
Cool, and we actually get a few kB back.

```
$ $ cat a.build 
#!/bin/bash

n=$(git rev-list --count origin/master..HEAD)

declare -rA variants=(
	["baseline"]=" -Ddebug-messages=true -Dlogging=true "
	["nodebug"]=" -Ddebug-messages=false -Dlogging=true "
	["nodebug-nologgging"]=" -Ddebug-messages=false -Dlogging=false "
)

for k in "${!variants[@]}"; do
	v="${variants[$k]}"

	meson setup $v build-$n-$k
	ninja -C build-$n-$k
done

$ git rebase -i -x ./a.build origin/master
... # also add ./a.build as the first line to have a comparison
$ size build-*-*/libkmod.so.2.5.0
   text	   data	    bss	    dec	    hex	filename
 111614	   1880	      8	 113502	  1bb5e	build-0-baseline/libkmod.so.2.5.0
 107700	   1880	      8	 109588	  1ac14	build-0-nodebug/libkmod.so.2.5.0
 102286	   1872	      8	 104166	  196e6	build-0-nodebug-nologgging/libkmod.so.2.5.0
 111614	   1880	      8	 113502	  1bb5e	build-1-baseline/libkmod.so.2.5.0
 108034	   1880	      8	 109922	  1ad62	build-1-nodebug/libkmod.so.2.5.0
 102620	   1872	      8	 104500	  19834	build-1-nodebug-nologgging/libkmod.so.2.5.0
 111614	   1880	      8	 113502	  1bb5e	build-2-baseline/libkmod.so.2.5.0
 108034	   1880	      8	 109922	  1ad62	build-2-nodebug/libkmod.so.2.5.0
  97992	   1872	      8	  99872	  18620	build-2-nodebug-nologgging/libkmod.so.2.5.0
 111614	   1880	      8	 113502	  1bb5e	build-3-baseline/libkmod.so.2.5.0
 105459	   1880	      8	 107347	  1a353	build-3-nodebug/libkmod.so.2.5.0
  95315	   1864	      8	  97187	  17ba3	build-3-nodebug-nologgging/libkmod.so.2.5.0
 111614	   1880	      8	 113502	  1bb5e	build-4-baseline/libkmod.so.2.5.0
 105459	   1880	      8	 107347	  1a353	build-4-nodebug/libkmod.so.2.5.0
  95315	   1864	      8	  97187	  17ba3	build-4-nodebug-nologgging/libkmod.so.2.5.0


$ $ size build-*-*/kmod
   text	   data	    bss	    dec	    hex	filename
 192398	   5416	    168	 197982	  3055e	build-0-baseline/kmod
 188435	   5416	    168	 194019	  2f5e3	build-0-nodebug/kmod
 183350	   5416	    168	 188934	  2e206	build-0-nodebug-nologgging/kmod
 192398	   5416	    168	 197982	  3055e	build-1-baseline/kmod
 188769	   5416	    168	 194353	  2f731	build-1-nodebug/kmod
 183684	   5416	    168	 189268	  2e354	build-1-nodebug-nologgging/kmod
 192398	   5416	    168	 197982	  3055e	build-2-baseline/kmod
 188769	   5416	    168	 194353	  2f731	build-2-nodebug/kmod
 179225	   5416	    168	 184809	  2d1e9	build-2-nodebug-nologgging/kmod
 192398	   5416	    168	 197982	  3055e	build-3-baseline/kmod
 186226	   5416	    168	 191810	  2ed42	build-3-nodebug/kmod
 176591	   5416	    168	 182175	  2c79f	build-3-nodebug-nologgging/kmod
 192398	   5416	    168	 197982	  3055e	build-4-baseline/kmod
 186226	   5416	    168	 191810	  2ed42	build-4-nodebug/kmod
 176591	   5416	    168	 182175	  2c79f	build-4-nodebug-nologgging/kmod
```

with most of the reduction coming from the first patch.

lucasdemarchi:
Applied, thanks.

PR #174, opened 2024-10-03
strbuf improvements

evelikov:
As mentioned a while back, converting the strbuf to disown instead of re-allocating kind of makes sense to me. So here is a quick and untested PR for that.

While in there, I also noticed that we could make `strbuf_str` never return NULL so I've fixed that up as well... Haven't added the `returns_nonnull` decoration though.

If people like the idea(s), I will do some testing and check the code for any more if quirks that we can drop.

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -231,12 +231,15 @@ static struct index_node_f *index_read(FILE *in, uint32_t offset)
 
 	if (offset & INDEX_NODE_PREFIX) {
 		struct strbuf buf;
+		ssize_t len;
+
 		strbuf_init(&buf);
-		if (buf_freadchars(&buf, in) < 0) {
+		len = buf_freadchars(&buf, in);
+		if (len < 0) {
 			strbuf_release(&buf);
 			return NULL;
 		}
-		prefix = strbuf_steal(&buf);
+		prefix = len ? strbuf_steal(&buf) : strdup("");
 	} else
 		prefix = strdup("");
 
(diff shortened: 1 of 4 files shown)


shared/strbuf.h
```diff
@@ -15,8 +15,8 @@ void strbuf_init(struct strbuf *buf);
 void strbuf_release(struct strbuf *buf);
 void strbuf_clear(struct strbuf *buf);
 
-/* Destroy buffer and return a copy as a C string */
-char *strbuf_steal(struct strbuf *buf);
+/* Invalidate the buffer and return its C string */
```
lucasdemarchi, line 18: Can we keep the name though?

shared/strbuf.c
```diff
@@ -62,8 +66,6 @@ char *strbuf_steal(struct strbuf *buf)
 
 const char *strbuf_str(struct strbuf *buf)
 {
-	if (!buf_grow(buf, buf->used + 1))
-		return NULL;
```
qhill, line 66: That makes something like:
```
struct strbuf buf;
strbuf_init(&buf);
if (condition)
    strbuf_pushchars(&buf, "test");
const char *s = strbuf_str(&buf);
```
problematic rather than strbuf_str just returning NULL when condition is false.

libkmod/libkmod-index.c
```diff
@@ -231,12 +231,15 @@ static struct index_node_f *index_read(FILE *in, uint32_t offset)
 
 	if (offset & INDEX_NODE_PREFIX) {
 		struct strbuf buf;
+		ssize_t len;
```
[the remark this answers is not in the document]
lucasdemarchi: I'm renaming the other one to len to make it consistent.

lucasdemarchi:
I'm leaving the last 2 commits out. We can come back to it later - it has a potential for introducing bugs and I'm not really happy with the corner cases like below:

```
strbuf_init(buf);
....
if (foo)
        strbuf_pushchar(buf, 'f');

s = strbuf_steal(buf);
```

So now we don't return NULL anymore, but we have to check in the caller, somewhere else, that we actually pushed something into the buffer. If we fail to do so it will hit a runtime assert, which may be hard to trigger if a call to strbuf_steal()/strbuf_str() in this situation is part of error handling.

PR #175, opened 2024-10-11
libkmod: Remove elf->changed

stoeckmann:
Keep changed ELF data only around as long as necessary. Otherwise it can happen that subsequent module operations lead to unintuitive results.

I've prepared a program which does the following:
1. Forcefully load a given module file
2. Try to load it without force if forcefully loading failed
3. Print version information

```
#include <err.h>
#include <libkmod.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

int main(int argc, char *argv[])
{
        struct kmod_ctx *ctx;
        struct kmod_module *mod;
        struct kmod_list *l, *list;
        const char *null_config;
        int flags, res;

        if (argc != 2)
                errx(1, "usage: poc modfile");

        ctx = kmod_new(NULL, &null_config);
        if (ctx == NULL)
                err(1, "kmod_new");

        res = kmod_module_new_from_path(ctx, argv[1], &mod);
        if (res < 0)
                err(1, "kmod_module_new_from_path");

        puts("forcefully load");
        flags = 0;
        flags |= KMOD_PROBE_FORCE_MODVERSION;
        flags |= KMOD_PROBE_FORCE_VERMAGIC;
        res = kmod_module_insert_module(mod, flags, NULL);
        if (res < 0) {
                puts("try without force");
                flags = 0;
                res = kmod_module_insert_module(mod, flags, NULL);
                if (res < 0)
                        fputs("loading module failed\n", stderr);
        }

        res = kmod_module_get_info(mod, &list);
        if (err < 0)
                err(1, "kmod_module_get_info");

        kmod_list_foreach(l, list) {
                const char *key = kmod_module_info_get_key(l);
                const char *value = kmod_module_info_get_value(l);

                if (strcmp(key, "vermagic") == 0)
                        printf("%s=%s\n", key, value);
        }

        kmod_module_unref(mod);
        kmod_unref(ctx);
        return 0;
}
```

I have compiled a kernel with module versioning support, i.e. `CONFIG_MODVERSIONS=y` which can be found in main menu config under `Enable loadable module support` during kernel configuration. My kernel has no module compression, but libkmod is compiled with xz support. Also, forced module loading is disabled, i.e. `CONFIG_MODULE_FORCE_LOAD=n`.

After all that, I have copied the module ntfs3.ko (anyone works, as long as they have version information) to a temporary directory, and created a compressed version of it as well which my kernel does not understand.
```
MYDIR=$(mktemp -d)
cd $MYDIR
cp /lib/modules/$(uname -r)/kernel/fs/ntfs3/ntfs3.ko .
xz --keep ntfs3.ko
```

As an unprivileged user, I have run the following commands:
```
$ ./poc ntfs3.ko
forcefully load
try without force
loading module failed
vermagic=6.11.3 SMP preempt mod_unload modversions
$ ./poc ntfs3.ko.xz
forcefully load
try without force
loading module failed
```

As you can see, it makes a difference how this module file is processed. If it is native or compressed in a form which the kernel understands, the ELF memory content is never modified. If it is modified, it is never turned back to normal, i.e. the version information is lost no matter what's done afterwards.

You can see this impact when the program is run as root user:
```
# ./poc ntfs3.ko.xz
forcefully load
try without force
# rmmod ntfs3
# ./poc ntfs3.ko
forcefully load
try without force
vermagic=6.11.3 SMP preempt mod_unload modversions
```

In fact, both versions lead to a loaded module with version information, but only the latter keeps the information in its module. The former one overwrites the ELF content, even though ELF data with version information is loaded into kernel. How? Without force, it calls `kmod_file_get_contents` which is the unchanged ELF data, but later modinfo calls access `kmod_file_get_elf` which is modified.

TLDR: It's very unintuitive what happens in these scenarios, so properly drop modified memory after requiring it.

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -35,7 +35,6 @@ struct kmod_modversion64 {
 
 struct kmod_elf {
 	const uint8_t *memory;
-	uint8_t *changed;
 	uint64_t size;
 	enum kmod_elf_class class;
 	struct kmod_elf_header {
@@ -143,15 +142,15 @@ static inline uint64_t elf_get_uint(const struct kmod_elf *elf, uint64_t offset,
 	return ret;
 }
 
-static inline int elf_set_uint(struct kmod_elf *elf, uint64_t offset, uint64_t size,
-			       uint64_t value)
+static inline int elf_set_uint(const struct kmod_elf *elf, uint64_t offset, uint64_t size,
+			       uint64_t value, uint8_t *changed)
 {
 	uint8_t *p;
 	size_t i;
 
 	ELFDBG(elf,
 	       "size=%" PRIu16 " offset=%" PRIu64 " value=%" PRIu64 " write memory=%p\n",
-	       size, offset, value, elf->changed);
+	       size, offset, value, changed);
 
 	assert(size <= sizeof(uint64_t));
 	assert(offset + size <= elf->size);
@@ -163,16 +162,7 @@ static inline int elf_set_uint(struct kmod_elf *elf, uint64_t offset, uint64_t s
 		return -1;
 	}
 
-	if (elf->changed == NULL) {
-		elf->changed = malloc(elf->size);
-		if (elf->changed == NULL)
-			return -errno;
-		memcpy(elf->changed, elf->memory, elf->size);
-		elf->memory = elf->changed;
-		ELFDBG(elf, "copied memory to allow writing.\n");
-	}
-
-	p = elf->changed + offset;
+	p = changed + offset;
 	if (elf->class & KMOD_ELF_MSB) {
 		for (i = 1; i <= size; i++) {
 			p[size - i] = value & 0xff;
@@ -292,7 +282,6 @@ struct kmod_elf *kmod_elf_new(const void *memory, off_t size)
 	}
 
 	elf->memory = memory;
-	elf->changed = NULL;
 	elf->size = size;
 	elf->class = class;
 
@@ -360,7 +349,6 @@ invalid:
 
 void kmod_elf_unref(struct kmod_elf *elf)
 {
-	free(elf->changed);
 	free(elf);
 }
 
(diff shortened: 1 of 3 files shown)


libkmod/libkmod-module.c
```diff
@@ -670,21 +671,13 @@ static int do_init_module(struct kmod_module *mod, unsigned int flags, const cha
...
-				     strerror(-err));
-		}
-
-		mem = kmod_elf_get_memory(elf);
+		stripped = kmod_elf_strip(elf, flags);
+		if (stripped == NULL) {
+			INFO(mod->ctx, "Failed to strip version information: %s\n",
+			    strerror(errno));
+			mem = kmod_elf_get_memory(elf);
+		} else
```
lucasdemarchi, line 679: Please balance the { }

libkmod/libkmod-module.c
```diff
@@ -670,21 +671,13 @@ static int do_init_module(struct kmod_module *mod, unsigned int flags, const cha
...
-				INFO(mod->ctx, "Failed to strip vermagic: %s\n",
-				     strerror(-err));
-		}
-
-		mem = kmod_elf_get_memory(elf);
+		stripped = kmod_elf_strip(elf, flags);
+		if (stripped == NULL) {
+			INFO(mod->ctx, "Failed to strip version information: %s\n",
+			     strerror(errno));
+			mem = kmod_elf_get_memory(elf);
```
lucasdemarchi, line 678: separate fix, but we are missing an error return here:  turn that INFO into error and return instead of continuing.

libkmod/libkmod-module.c
stoeckmann: Adjusted.

libkmod/libkmod-module.c
stoeckmann: And removed. Some tests fail and I can't reproduce it on my system.

stoeckmann: See https://github.com/kmod-project/kmod/pull/176 for details

libkmod/libkmod-elf.c
```diff
@@ -642,25 +631,50 @@ int kmod_elf_strip_vermagic(struct kmod_elf *elf)
...
 }
 
+const void *kmod_elf_strip(const struct kmod_elf *elf, unsigned int flags)
+{
+	uint8_t *changed;
+	int err = 0;
+
+	if ((flags & (KMOD_INSERT_FORCE_MODVERSION | KMOD_INSERT_FORCE_VERMAGIC)) == 0) {
+		errno = EINVAL;
+		return NULL;
```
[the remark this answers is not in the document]
stoeckmann: Adjusted

lucasdemarchi:
Applied, thanks.

PR #177, opened 2024-10-11
Trivial (random) man changes

evelikov:
- Document the rmmod behaviour, when a list of modules is provided
 - One silly regression when I've added BUGS section :sweat_smile: 
 - Couple of more fixes from the man page folks, that I've missed


lucasdemarchi:
Adding a fixup to the first commit. @evelikov does that look good?

lucasdemarchi:
Applied, thanks.

PR #179, opened 2024-10-15
Fix GCC sanitizer load ordering

evelikov:
Split from https://github.com/kmod-project/kmod/pull/172

Adjust the testsuite to properly preserve LD_PRELOAD and swap the ASAN_OPTIONS workaround, with LD_PRELOAD based wrapper.

Thus devs can use the sanitizers locally with zero setup. This exact same approach will be reused for clang based sanitizers.

### DIFF
--- a/testsuite/meson.build
+++ b/testsuite/meson.build
@@ -97,7 +97,8 @@ _testsuite = [
 foreach input : _testsuite
   test(
     input,
-    executable(
+    test_wrapper,
+    args : executable(
       input,
       files('@0@.c'.format(input)),
       include_directories : top_include,
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,7 +91,7 @@ jobs:
 
       - name: test (meson)
         if: ${{ matrix.build == 'meson' }}
-        run: cd build && meson test
+        run: cd build && meson test || meson test --verbose
 
       - name: test (autotools)
         if: ${{ matrix.build == 'autotools' }}
@@ -123,4 +123,4 @@ jobs:
 
       - name: test (32bit) (meson)
         if: ${{ matrix.build == 'meson' && matrix.container.multilib == 'true' }}
-        run: cd build32 && meson test
+        run: cd build32 && meson test || meson test --verbose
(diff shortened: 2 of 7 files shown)


testsuite/meson.build
```diff
@@ -97,7 +97,8 @@ _testsuite = [
 foreach input : _testsuite
   test(
     input,
-    executable(
+    test_wrapper,
+    args : executable(
```
lucasdemarchi, line 101: couldn't *_OPTIONS checks be done in meson itself and then just set the env accordingly, possibly with `source`?
This way when calling the binaries outside of meson, the only thing we'd need is:

```
# once..
. scripts/sanitizer-env.sh

call tests how many times we want
```

rather than having to wrap the test call with the test-wrapper.sh for each invocation

lucasdemarchi: Ok, let's go with this version that works rather than trying to polish it too much. Thanks

lucasdemarchi: we may need to update the README in the testsuite. It used to be the case that we could just `make rootfs && <path/to/test>`, which is not the case anymore. Example:

```
=================================== 16/18 ====================================
test:         test-depmod
start time:   14:11:04
duration:     0.32s
result:       exit status 0
command:      MESON_TEST_ITERATION=1 MALLOC_PERTURB_=117 MSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 LD_LIBRARY_PATH=/home/ldmartin/p/kmod/build/testsuite ASAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1 UBSAN_OPTIONS=halt_on_error=1:abort_on_error=1:print_summary=1:print_stacktrace=1 /home/ldmartin/p/kmod/scripts/test-wrapper.sh testsuite/test-depmod
```

lucasdemarchi: So... I think this should be handled by the testsuite itself, possibly adding a --disable-sanitizer for all the tests for the case in which "test has been built with sanitizer and we want to run it without". Anyway, something we can think about doing on top.

lucasdemarchi:
Applied, thanks.

PR #180, opened 2024-10-15
Clang all the warnings

evelikov:
Split from https://github.com/kmod-project/kmod/pull/172

Bunch of fixes reported by clang
 - some corner-case arguments handling (insmod, rmmod)
 - implementation specific signed/unsigned promotion
 - undefined behaviour - incompatible function type, adding zero to a NULL

### DIFF
--- a/shared/array.c
+++ b/shared/array.c
@@ -69,10 +69,14 @@ int array_append(struct array *array, const void *element)
 int array_append_unique(struct array *array, const void *element)
 {
 	void **itr = array->array;
-	void **itr_end = itr + array->count;
-	for (; itr < itr_end; itr++)
-		if (*itr == element)
-			return -EEXIST;
+
+	if (array->count != 0) {
+		void **itr_end = itr + array->count;
+
+		for (; itr < itr_end; itr++)
+			if (*itr == element)
+				return -EEXIST;
+	}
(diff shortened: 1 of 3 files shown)


shared/array.c
```diff
@@ -66,10 +66,14 @@ int array_append(struct array *array, const void *element)
 int array_append_unique(struct array *array, const void *element)
 {
 	void **itr = array->array;
-	void **itr_end = itr + array->count;
-	for (; itr < itr_end; itr++)
-		if (*itr == element)
-			return -EEXIST;
+
+	if (itr) {
```
lucasdemarchi, line 70: that's a weird condition for what this is fixing. We should do nothing if array->count is 0, rather than do nothing if itr is NULL, which it never is.

shared/hash.c
```diff
@@ -61,9 +61,11 @@ void hash_free(struct hash *hash)
 		if (hash->free_value) {
 			struct hash_entry *entry, *entry_end;
 			entry = bucket->entries;
-			entry_end = entry + bucket->used;
-			for (; entry < entry_end; entry++)
-				hash->free_value((void *)entry->value);
+			if (entry) {
```
lucasdemarchi, line 64: ditto

PR #181, opened 2024-10-15
libkmod: Add another section size check

stoeckmann:
Make sure that section is actually large enough to hold a 32 bit value.


lucasdemarchi:
Applied, thanks.

PR #182, opened 2024-10-16
man/insmod: Fix style

stoeckmann:
Add missing newline for better formatting and to stay in sync with other options.


stoeckmann:
I've restarted reading the tools after all these changes of the last weeks, and insmod was one of the smaller ones. Spotted the error code regression and wanted to check the manual page if something's stated about these error cases like "I won't insert modules from stdin."

And while at it, spotted the missing line.

lucasdemarchi:
Applied, thanks.

PR #183, opened 2024-10-16
build: add builtin overflow checks to configure.ac

stoeckmann:
Without these, the build becomes quite noisy and doesn't add builtin functions even though they exist.


stoeckmann:
Relates to https://github.com/kmod-project/kmod/issues/162

Even though autotools are pretty much obsolete now, we can see all these warnings during build in GitHub CI. Let's fix it for now before removing autotools completely. Is there any idea when this will happen?

stoeckmann:
So let's not hurt the late adopters during v34 then. After all it would be a shame to see distributors running without our optimized size checks. :)

lucasdemarchi:
Applied, thanks.

PR #184, opened 2024-10-16
modinfo: Fix OOB access and signed integer overflow

stoeckmann:
It is possible to triger undefined behavior with modinfo and kernel modules with strings longer than INT_MAX.

Proof of Concept:

1. Create a kernel module with two identical parameter strings longer than INT_MAX
```
echo > poc.ko.zstd.b64 << EOF
KLUv/WQMACUDABJEDxiwqTEsiruIPi2K8lIhjb2jCJoqNBGye6f/TB1mYU1BDE5VVRidKvgRDbx+
z7nWdnfXipGZqRTiKSHJ8QQRIFDDLN0DQBC7N9CK2wH5OMI/9ESNlRPG764EZThEvvkU3Qg7Px9Z
EOF
base64 -d poc.ko.zstd.b64 | zstd -d > poc.ko
(echo -n "parm="; dd if=/dev/zero bs=4096 count=524288 | tr '\0' A; echo -en ":B\x00") >> poc.ko
(echo -n "parm="; dd if=/dev/zero bs=4096 count=524288 | tr '\0' A; echo -en ":B\x00") >> poc.ko
```
2. Run modinfo
```
modinfo -p poc.ko
```

Crashes on my system due to OOB access. Alternatively, compile with address sanitizer.

The second commit fixes signed integer overflow. You can trigger it by modifying the previous command and replace `parm=` with `parmx` or anything else which will turn the parameter key into an arbitrary key. Run modinfo without `-p` then. I recommend to compile with `-fsanitize=undefined` to see the issue.

### DIFF
--- a/tools/modinfo.c
+++ b/tools/modinfo.c
@@ -34,14 +34,19 @@ struct param {
 	int typelen;
 };
 
-static struct param *add_param(const char *name, int namelen, const char *param,
-			       int paramlen, const char *type, int typelen,
+static struct param *add_param(const char *name, size_t namelen, const char *param,
+			       size_t paramlen, const char *type, size_t typelen,
 			       struct param **list)
 {
 	struct param *it;
 
+	if (namelen > INT_MAX || paramlen > INT_MAX || typelen > INT_MAX) {
+		errno = EINVAL;
+		return NULL;
+	}
+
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Applied, thanks

PR #189, opened 2024-10-17
libkmod: Use pread where appropriate

stoeckmann:
Since we do not want to modify the current position in file, use pread instead of read + lseek. Removes one lseek call per module, which for depmod on Arch Linux means 6143 calls.

### DIFF
--- a/libkmod/libkmod-file.c
+++ b/libkmod/libkmod-file.c
@@ -95,8 +95,7 @@ struct kmod_file *kmod_file_open(const struct kmod_ctx *ctx, const char *filenam
 		return NULL;
 	}
 
-	sz = read_str_safe(file->fd, buf, sizeof(buf));
-	lseek(file->fd, 0, SEEK_SET);
+	sz = pread_str_safe(file->fd, buf, sizeof(buf), 0);
 	if (sz != (sizeof(buf) - 1)) {
 		if (sz < 0)
 			errno = -sz;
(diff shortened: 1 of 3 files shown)


libkmod/libkmod-file.c
```diff
@@ -75,15 +75,41 @@ struct kmod_elf *kmod_file_get_elf(struct kmod_file *file)
 	return file->elf;
 }
 
+static inline ssize_t pread_safe(int fd, char *buf, size_t buflen, off_t offset)
```
lucasdemarchi, line 78: it's almost a 1:1 copy of read_str_safe(). Better to add this as pread_str_safe() in util.c

stoeckmann: It's copied from there. ;) Adjusted

lucasdemarchi:
> Fwiw we could inline `read_str_ulong()` into the only place that uses it and convert all `read_str_safe()` instances to 

then when we have a second user we move it back? No, I think we are fine with that in the util.c

>`pread_str_safe()`... Not sure if we want to go there, just thought it was worth mentioning.

only the ones that do a lseek() after the call.

lucasdemarchi:
I think these other changes can be done on top. Applied, thanks.

evelikov:
Could easily port/adapt the `read_str_safe` tests to `pread_str_safe`

PR #191, opened 2024-10-17
build: use cp instead of rsync

stoeckmann:
Having rsync as a build dependency is quite heavy.

@evelikov is there a strong reason to use rsync over cp here? I don't have it installed so running tests always involved modifying the script for me.

### DIFF
--- a/scripts/setup-modules.sh
+++ b/scripts/setup-modules.sh
@@ -8,7 +8,7 @@ MODULE_PLAYGROUND=$3
 
 # TODO: meson allows only out of tree builds
 if test "$SRCDIR" != "$BUILDDIR"; then
-    rsync --recursive --times "$SRCDIR/$MODULE_PLAYGROUND/" "$MODULE_PLAYGROUND/"
+    cp -a "$SRCDIR/$MODULE_PLAYGROUND/" "$MODULE_PLAYGROUND/"
 fi
 
 export MAKEFLAGS=${MAKEFLAGS-"-j$(nproc)"}


stoeckmann:
So, right now I have a hard time reproducing the issue in https://github.com/kmod-project/kmod/pull/121. About my system:
- coreutils 9.5
- meson 1.5.2
- ninja 1.12.1

I perform these steps with `cp -r` in place, as it was for the mentioned issue:
```
mkdir build
cd build
meson setup --native-file ../build-dev.ini -Ddocs=false . ..
meson compile
meson test
meson compile
```

The modules are compiled only during first `meson compile`. I have to manually run the following command to see them being compiled again:
```
for i in $(find ../testsuite/module-playground/); do touch testsuite/module-playground/$(basename $i); done
```

It seems that my `cp -r` call preserves the timestamps, while the one from @lucasdemarchi does not. And that's why I suggested `-a`. It's supposed to preserve timestamps (granted, it's a GNU extension). If GNU extensions are not wanted, we can even run `touch -r` which is part of POSIX, although we already require `/bin/bash` to be in place.

@lucasdemarchi, can you test this one on your system?

lucasdemarchi:
Doing just a cp without the rm we had before has the limitation that stale modules aren't cleaned up.

lucasdemarchi:
> It seems that my `cp -r` call preserves the timestamps, while the one from @lucasdemarchi does not. And that's why I suggested `-a`. It's supposed to preserve timestamps (granted, it's a GNU extension). If GNU extensions are not wanted, we can even run `touch -r` which is part of POSIX, although we already require `/bin/bash` to be in place.

we use GNU extensions

stoeckmann:
> Doing just a cp without the rm we had before has the limitation that stale modules aren't cleaned up.

Just tested this on my Arch system with rsync installed:

```
$ rm ../testsuite/module-playground/mod-simple.c
$ meson compile
$ ls testsuite/module-playground/mod-simple*
testsuite/module-playground/mod-simple.c    testsuite/module-playground/mod-simple.mod.c
testsuite/module-playground/mod-simple.ko   testsuite/module-playground/mod-simple.mod.o
testsuite/module-playground/mod-simple.mod  testsuite/module-playground/mod-simple.o
```

Would have surprised me if rsync would keep some kind of state to create a full diff and `--delete` would delete already compiled modules.

I have updated the PR to contain `cp -a` again, since that's what an equivalent of rsync should be here, using GNU extensions.

stoeckmann:
Thanks for adding all this. Closing this one (which should have been marked as a draft because it was more like an issue + code example of what I try to explain). :)

PR #192, opened 2024-10-17
Swap rsync for cp --archive for module sources copying

evelikov:
As Tobias reported, rsync is a bit of heavyweight dependency. We introduced it, as a replacement for the rm/cp -r previously used.

The rsync was inspired since, unlike make, meson will build all the test binaries/artefacts even without calling "meson test".

We can go back to cp with --archive (--preserve=timestamps at least), which will ensure we don't get stale files. To ensure the second run doesn't copy the source folder as _subfolder_ of the dest we need to wildcard the copy... Plus we need a proper destination folder in the first place.

With this, we get a no-op second+ builds - be that with meson or make. Since the explicit always-dirty state is by design, drop the meson TODO and document the output variable.

Confirmed by comparing both the `make --debug` output and the execution times.

Reported-by: Tobias Stoeckmann <tobias@stoeckmann.org>

### DIFF
--- a/.github/actions/setup-alpine/action.yml
+++ b/.github/actions/setup-alpine/action.yml
@@ -23,7 +23,6 @@ runs:
           linux-edge-dev \
           meson \
           openssl-dev \
-          rsync \
           scdoc \
           tar \
           xz-dev \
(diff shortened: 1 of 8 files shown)


stoeckmann (approved):
Thanks for spotting the cp bug, i.e. not copying with wildcards!

PR #195, opened 2024-10-19
tools: getopt-log polish

evelikov:
Adjust the loop to follow the in-documentation example.

Namely, we don't need the unused idx and removing the seemingly
infinite loop helps static analysers better reason about the code flow.

---

Only insmod, lsmod and rmmod for now. The others will come at later date as I do through and document/deprecation/update the existing option handling.

### DIFF
--- a/tools/insmod.c
+++ b/tools/insmod.c
@@ -8,6 +8,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 
 #include <shared/util.h>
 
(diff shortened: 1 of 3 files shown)


lucasdemarchi (approved):
humn... I'm pulling this because of the unused idx removal and just let the other change ride along.

changing a loop from:

```
for (;;) }
    if (foo)
        break;
}
```

to 

```
while (!foo) {

}
```

isn't changing any possible infinite loop and the 1st version is perfectly fine for avoiding the assign-and-check statement. 

If a static analyzer thinks these are different, maybe we shouldn't use that static analyzer.

lucasdemarchi:
Applied, thanks.

PR #197, opened 2024-10-20
tools/depmod: use separate arrays for alias,xxxdep values

evelikov:
Currently, we walk the info list multiples times each time filtering all but one key. Just create a few arrays to avoid that, saving 2-3% cycles at the cost of extra ~500bytes per module.

---

NOTE: with this PR I have started using loop counter declaration in the init-statement. I would love if we can have these, since they help for variable scoping and tidier code.

### DIFF
--- a/tools/depmod.c
+++ b/tools/depmod.c
@@ -897,6 +897,9 @@ struct mod {
 	char *uncrelpath; /* same as relpath but ending in .ko */
 	struct kmod_list *info_list;
 	struct kmod_list *dep_sym_list;
+	struct array alias_values;
+	struct array softdep_values;
+	struct array weakdep_values;
 	struct array deps; /* struct symbol */
 	size_t baselen; /* points to start of basename/filename */
 	size_t modnamesz;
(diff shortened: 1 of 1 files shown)


tools/depmod.c
```diff
@@ -1572,6 +1582,25 @@ static int depmod_load_modules(struct depmod *depmod)
 
 load_info:
 		kmod_module_get_info(mod->kmod, &mod->info_list);
+		kmod_list_foreach(l, mod->info_list) {
+			const char *key = kmod_module_info_get_key(l);
+
+			if (streq(key, "alias")) {
+				const char *value = kmod_module_info_get_value(l);
+				array_append(&mod->alias_values, value);
```
stoeckmann, line 1590: These can return error values in OOM conditions and should be handled accordingly.

stoeckmann (changes requested):

stoeckmann:
> saving 2-3% cycles at the cost of extra ~500bytes per module.

That's around 4 MB on Arch Linux. Sounds like a worthy trade-off for a one-shot tool like depmod. :)

> NOTE: with this PR I have started using loop counter declaration in the init-statement. I would love if we can have these, since they help for variable scoping and tidier code.

+1 for tighter scope

tools/depmod.c
```diff
@@ -1572,6 +1582,25 @@ static int depmod_load_modules(struct depmod *depmod)
 
 load_info:
 		kmod_module_get_info(mod->kmod, &mod->info_list);
```
stoeckmann, line 1584: We can even save some memory, because the requirement of `mod->info_list` is gone. All its use cases can be covered with the newly introduced arrays now.

stoeckmann: Oh, I see... Too bad then :/

lucasdemarchi:
Applied, thanks

PR #203, opened 2024-10-22
libkmod: Check node offset in index_mm_read_node

stoeckmann:
Add a cheap but important check to make sure that offsets do not point outside of memory-mapped area.


evelikov (approved):
Nicely spotted.

Can we also get a equivalent for the FILE side APIs? This PR, separate one - whichever works for you.

stoeckmann:
We don't carry the file size with us around, so we would have to add it. It's not impossible, but as soon as anything would be read, we run into EOF and return NULL anyway.

lucasdemarchi:
Applied, thanks

PR #205, opened 2024-10-22
scripts/kmod-symlink.sh: ignore prefix when provided absolute dir

evelikov:
The end-user can provide either relative (to prefix) or an absolute directory for bindir. In case of the latter, we need to manually construct the path it seems.

---

Noticed this while trying to update the Gentoo ebuild. While they have migrated to merged /usr (+ bin/sbin) although split-usr is still an option. As the equivalent of `--prefix /usr --bindir /bin` is passed to meson.

In theory we could try and polish all environment variables so they don't add the prefix unless needed, although I'm inclined to deffer on as-needed basis. This includes the bindir listing on `meson setup`.

### DIFF
--- a/scripts/kmod-symlink.sh
+++ b/scripts/kmod-symlink.sh
@@ -7,4 +7,4 @@ set -euo pipefail
 #
 # For context read through https://github.com/mesonbuild/meson/issues/9
 
-ln -sf kmod "$MESON_INSTALL_DESTDIR_PREFIX/$1"
+ln -sf kmod "$DESTDIR/$1"
--- a/meson.build
+++ b/meson.build
@@ -186,6 +186,7 @@ features = []
 sysconfdir = get_option('sysconfdir')
 cdata.set_quoted('SYSCONFDIR', sysconfdir)
 
+bindir = join_paths(get_option('prefix'), get_option('bindir'))
 libdir = join_paths(get_option('prefix'), get_option('libdir'))
 
 distconfdir = get_option('distconfdir')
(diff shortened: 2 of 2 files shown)


scripts/kmod-symlink.sh
```diff
@@ -7,4 +7,9 @@ set -euo pipefail
 #
 # For context read through https://github.com/mesonbuild/meson/issues/9
 
-ln -sf kmod "$MESON_INSTALL_DESTDIR_PREFIX/$1"
+if [[ $1 == /* ]]; then
+    echo "Absolute destination folder provided, ignoring prefix"
+    ln -sf kmod "$DESTDIR/$1"
```
eli-schwartz, line 12: https://github.com/kmod-project/kmod/blob/33e3d24cdb270cdfab71e9bd1c856e6430b3c11e/meson.build#L445-L446

join_paths is already abspath aware -- just move path building for PREFIX into meson.build

Note recent versions of meson can natively install symlinks.

eli-schwartz (changes requested):

scripts/kmod-symlink.sh
evelikov: > join_paths is already abspath aware -- just move path building for PREFIX into meson.build

Update to always use prepend prefix. Thanks o/

> Note recent versions of meson can natively install symlinks.

That would be awesome, do you have any references?

eli-schwartz: https://mesonbuild.com/Reference-manual_functions.html#install_symlink

You would need to bump the minimum required version from 0.60 to 0.61 but I'm not aware of any situation where that is a hardship, e.g. updating that version still fits within the last version to support python 3.6 (which was 0.61) and there are no distros that ship with exactly meson 0.60 other than Fedora 36 which feels a bit overly specific.

PR #206, opened 2024-10-23
meson: Fix installing without DESTDIR

lucasdemarchi:
Script fails because of `set -u`. Make sure DESTDIR is set to something

Reported-by: Sedat Dilek <sedat.dilek@gmail.com>


dileks:
Feel free to add:
```
Tested-by: Sedat Dilek <sedat.dilek@gmail.com> # meson v1.6.0
```

lucasdemarchi:
Applied, thanks.

PR #209, opened 2024-10-24
Meson bits

evelikov:
A bit of a mix bag there... As I started testing the meson 0.61 install_link() I was we had some other loose ends in the handling of distconfig/moduledir/*shcompletiondir. So here's a small series to address them.


meson.build
```diff
@@ -206,6 +206,14 @@ foreach tuple : _customdirs
   customdir = get_option(dir_option)
   if customdir == ''
     customdir = def_path
+  else
+    if customdir.endswith('/')
+      customdir = customdir.substring(0, -1)
+    endif
```
lucasdemarchi, line 212: someone will always want to cause trouble by doing ` meson setup -D moduledir=/got/you///`

instead of the `if`, maybe just do: `customdir = customdir.strip(strip_chars="'')`

meson.build
```diff
@@ -466,11 +466,9 @@ _tools = [
   'rmmod',
 ]
 
-kmod_symlink = find_program('scripts/kmod-symlink.sh')
 foreach tool : _tools
   if get_option('tools')
-    symlink = join_paths(bindir, tool)
-    meson.add_install_script(kmod_symlink, symlink)
+    install_symlink(tool, pointing_to: 'kmod', install_dir: bindir)
```
lucasdemarchi, line 471: do we have the exact same tree output?

meson.build
evelikov: Initially opted against `.strip()` since it removes both leading and trailing symbols.

Updated to use them and add the leading one... As a bonus, it will convert `///bar/baz////` to `/bar/baz`.

meson.build
evelikov: Just had another look and they seems identical.

evelikov:
Added an extra meson commit to reduce the duplication in the shell-completions handling, not that strongly attach to it though.

meson.build
lucasdemarchi: humn.... and we don't have rstrip(). :-/

meson.build
```diff
@@ -183,12 +183,14 @@ features = []
...
+
 sysconfdir = get_option('sysconfdir')
 cdata.set_quoted('SYSCONFDIR', sysconfdir)
 
-bindir = join_paths(get_option('prefix'), get_option('bindir'))
-includedir = join_paths(get_option('prefix'), get_option('includedir'))
-libdir = join_paths(get_option('prefix'), get_option('libdir'))
+bindir = join_paths(prefixdir, get_option('bindir'))
+includedir = join_paths(prefixdir, get_option('includedir'))
+libdir = join_paths(prefixdir, get_option('libdir'))
```
lucasdemarchi, line 193: I'm thinking that maybe we should move all the xxxxdir = ... get_option() here and into a single block. Something like below:

```
prefixdir = get_option('prefix')
sysconfdir = get_option('sysconfdir')
bindir = join_paths(prefixdir, get_option('bindir'))
includedir = join_paths(prefixdir, get_option('includedir'))
libdir = join_paths(prefixdir, get_option('libdir'))
datadir =  .... ?
bashcompletiondir =  ...
fishcompletiondir =  ...
zshcompletiondir = ...

```

Because it seems we forgot to also add prefixdir to datadir.  Then you use the variables above in the arrays to denote the default value from user, if set.

meson.build
```diff
@@ -219,13 +219,13 @@ foreach tuple : _customdirs
 endforeach
 
 _completiondirs = [
-  ['bashcompletiondir', 'bash-completion', 'bash-completion/completions', 'shell-completion/bash/@0@'],
-  ['fishcompletiondir', 'fish',            'fish/vendor_functions.d',     'shell-completion/fish/@0@.fish'],
-  ['zshcompletiondir',  '',                'zsh/site-functions',          'shell-completion/zsh/_@0@'],
+  ['bash', 'bash-completion', 'bash-completion/completions', '@0@'],
+  ['fish', 'fish',            'fish/vendor_functions.d',     '@0@.fish'],
+  ['zsh',  '',                'zsh/site-functions',          '_@0@'],
```
lucasdemarchi, line 224: see other comment.... I think we should split the part that calls get_option() to be all in a single place.

meson.build
```diff
@@ -257,7 +257,7 @@ foreach tuple : _completiondirs
   if completiondir != 'no'
     foreach comp : _completions
       install_data(
-        files(ins_path.format(comp)),
+        files('shell-completion/@0@/'.format(tuple[0]) + ins_path.format(comp)),
```
lucasdemarchi, line 260: we should prefer join_paths() for path building. Or event better... use / operator that is available since meson 0.49

meson.build
evelikov: Not going to lie, the way meson deals with paths is quite annoying IMHO. In particular - if the user has provided (implicitly or explicitly) relative one, the prefix will be added when working with meson constructs but otherwise you have to add it.

</rant>

Sure can do. I will also take the opportunity to sort the summary, based on your example.

meson.build
evelikov: Ack, switched to `/` with prep commit and updated the patches to use `/`.

meson.build
```diff
@@ -213,6 +213,13 @@ foreach tuple : _customdirs
   customdir = get_variable(dir_option)
   if customdir == ''
     customdir = def_path
+  else
+    if not customdir.startswith('/')
+      error('User provided @0@, \'@1@\' is not an absolute path.'
+          .format(dir_option, customdir))
+    endif
+    # Strip all leading/trailing and re-add only the leading one.
+    customdir = / customdir.strip('/')
```
lucasdemarchi, line 221: humn... I don't think this works as there are no left side operand.... you'd need to do '' / customdir.strip('/')

meson.build
```diff
@@ -183,21 +183,28 @@ features = []
...
 sysconfdir = get_option('sysconfdir')
-cdata.set_quoted('SYSCONFDIR', sysconfdir)
-
-bindir = get_option('prefix') / get_option('bindir')
-includedir = get_option('prefix') / get_option('includedir')
-libdir = get_option('prefix') / get_option('libdir')
+bindir = prefixdir / get_option('bindir')
+includedir = prefixdir / get_option('includedir')
+libdir = prefixdir / get_option('libdir')
+datadir = prefixdir / get_option('datadir')
```
lucasdemarchi, line 190: nice... this makes it more consistent and we don't print in the summmary:

```
$ meson setup -Ddatadir=share2 build5
  User defined options
    datadir       : share2
```
which is confusing, although it does the right thing and on installation creates /usr/share2.

meson.build
lucasdemarchi: indeed:

```
kmod $ meson setup -Dmoduledir=/test/foo build6
...
meson.build:222:16: ERROR: Unknown statement.
```

lucasdemarchi (approved):

meson.build
lucasdemarchi: Fixed this one

evelikov: Thanks :heart: 

Would it make sense to add a few `meson setup` permutations to our CI (without building)?

lucasdemarchi: yeah, not a bad idea

evelikov: Opened https://github.com/kmod-project/kmod/pull/245

PR #210, opened 2024-10-24
Avoid copying elf symbol names, cache elf_get_section_info results

evelikov:
Couple of tiny tweaks, which shave a ~4MB of memory and ~5% CPU cycles off a depmod run, although the actual perf difference is within the noise margin.

Grab whichever commit make sense.

OOC: any ideas when all `.strtab` started getting only `__crc_` symbols? Going back to ~2011 and it seems to have always been the case.

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -644,10 +644,9 @@ static int kmod_elf_get_symbols_symtab(const struct kmod_elf *elf,
 {
 	uint64_t i, last, off, size;
 	const char *strings;
-	char *itr;
 	struct kmod_modversion *a;
 	int count, err;
-	size_t vec_size, total_size;
+	size_t total_size;
 
 	*array = NULL;
 
@@ -681,9 +680,8 @@ static int kmod_elf_get_symbols_symtab(const struct kmod_elf *elf,
 		}
 	}
 
-	/* sizeof(struct kmod_modversion) * count + size */
-	if (umulsz_overflow(sizeof(struct kmod_modversion), count, &vec_size) ||
-	    uaddsz_overflow(size, vec_size, &total_size)) {
+	/* sizeof(struct kmod_modversion) * count */
+	if (umulsz_overflow(sizeof(struct kmod_modversion), count, &total_size)) {
 		return -ENOMEM;
 	}
 
@@ -691,21 +689,16 @@ static int kmod_elf_get_symbols_symtab(const struct kmod_elf *elf,
 	if (*array == NULL)
 		return -errno;
 
-	itr = (char *)(a + count);
 	last = 0;
 	for (i = 0, count = 0; i < size; i++) {
 		if (strings[i] == '\0') {
-			size_t slen = i - last;
 			if (last == i) {
 				last = i + 1;
 				continue;
 			}
 			a[count].crc = 0;
 			a[count].bind = KMOD_SYMBOL_GLOBAL;
-			a[count].symbol = itr;
-			memcpy(itr, strings + last, slen);
-			itr[slen] = '\0';
-			itr += slen + 1;
+			a[count].symbol = strings + last;
 			count++;
 			last = i + 1;
 		}
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -483,10 +483,9 @@ static inline void elf_get_modversion_lengths(const struct kmod_elf *elf, size_t
 /* array will be allocated with strings in a single malloc, just free *array */
 int kmod_elf_get_modversions(const struct kmod_elf *elf, struct kmod_modversion **array)
 {
-	size_t off, crclen, namlen, slen, verlen;
+	size_t off, crclen, namlen, verlen;
 	uint64_t sec_off, size;
 	struct kmod_modversion *a;
-	char *itr;
 	int i, count, err;
 
 	elf_get_modversion_lengths(elf, &verlen, &crclen, &namlen);
@@ -505,10 +504,12 @@ int kmod_elf_get_modversions(const struct kmod_elf *elf, struct kmod_modversion 
 
 	count = size / verlen;
 
-	off = sec_off;
-	slen = 0;
+	*array = a = malloc(sizeof(struct kmod_modversion) * count);
+	if (*array == NULL)
+		return -errno;
 
-	for (i = 0; i < count; i++, off += verlen) {
+	for (i = 0, off = sec_off; i < count; i++, off += verlen) {
+		uint64_t crc = elf_get_uint(elf, off, crclen);
 		const char *symbol = elf_get_mem(elf, off + crclen);
 		size_t nlen = strnlen(symbol, namlen);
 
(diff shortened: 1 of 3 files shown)


libkmod/libkmod-elf.c
```diff
@@ -682,43 +712,34 @@ static int kmod_elf_get_symbols_symtab(const struct kmod_elf *elf,
...
 		}
 	}
 	if (strings[i - 1] != '\0') {
-		size_t slen = i - last;
 		a[count].crc = 0;
 		a[count].bind = KMOD_SYMBOL_GLOBAL;
-		a[count].symbol = itr;
-		memcpy(itr, strings + last, slen);
-		itr[slen] = '\0';
+		a[count].symbol = strings + last;
```
stoeckmann, line 742: This could lead to an unterminated string, after all `strings[i - 1] != '\0'`. According to latest changes to kernel and according to spec, such a section is invalid. We should check if the section ends with '\0' and if it does not, consider the section invalid. We have such a check in `kmod_elf_new` already, anyway

libkmod/libkmod-elf.c
```diff
@@ -682,43 +712,34 @@ static int kmod_elf_get_symbols_symtab(const struct kmod_elf *elf,
 	if (strings[i - 1] != '\0')
 		count++;
 
-	/* sizeof(struct kmod_modversion) * count + size + 1 */
+	/* sizeof(struct kmod_modversion) * count + 1 */
```
stoeckmann, line 715: And with previously mentioned change, this additional 1 can be skipped.

stoeckmann (changes requested):

libkmod/libkmod-elf.c
evelikov: Nicely spotted. Do you have any reference to these latest changes and the spec?

stoeckmann: - https://git.kernel.org/pub/scm/linux/kernel/git/modules/linux.git/commit/?h=modules-next&id=837031e052af32c747906238fb1feb87778e4fe0
- man 5 elf
- https://lkml.org/lkml/2024/10/19/334
- https://lkml.org/lkml/2024/10/19/336

libkmod/libkmod-elf.c
evelikov: Indeed, as does the `count++` couple of lines above. Thanks o/

libkmod/libkmod-elf.c
evelikov: Awesome, thank you.

libkmod/libkmod-elf.c
```diff
@@ -679,48 +709,30 @@ static int kmod_elf_get_symbols_symtab(const struct kmod_elf *elf,
 			last = i + 1;
 		}
 	}
-	if (strings[i - 1] != '\0')
-		count++;
```
stoeckmann, line 683: You must keep this check in place, but instead of trying to handle this case, just free memory and return -EINVAL or something like that to indicate that this section is not valid.

stoeckmann (changes requested):

libkmod/libkmod-elf.c
evelikov: Indeed - misread the `\0` validation we have in `kmod_elf_new()`.

Maybe it would make more sense to have the validation in `kmod_elf_get_section()`,instead of duplicating it in each caller?

evelikov:
> OOC: any ideas when all .strtab started getting only __crc_ symbols? Going back to ~2011 and it seems to have always been the case.

Grabbed a random 3.10 kernel and it rarely hits the fallback path, in contrary to my 5.15 one... Should probably check if we can swap the order, although my gut feeling is that it'll cause breakage.

The generated index files do vary across the two kernels. At a glance an extra `crc16` module is inserted in a few more places and/or it's order is different. Such old kernels lack `modules.order` so I suspect this is intentional. Will add to the TODO of things to check properly.

libkmod/libkmod-elf.c
stoeckmann: I'd actually keep it there. The idea is to support as many (weird) modules as possible and if we don't even want to access symbols, we should not fail, since we do these checks to prevent bugs within kmod, not within the kernel. After all it's the kernel's job to make sure that the modules are valid after all.

stoeckmann: https://github.com/kmod-project/kmod/pull/222 is another example that these sections need different treatments. Going through __versions for such a verification in a general "let's find offset and size" function seems off.

evelikov: Since it will conflict with your PR and need some fixing - as per reviews - I have switched this to a draft.

stoeckmann: It would be nice to get this in, because I've just noticed that the current code could lead to overflows on 32 bit systems. If the string section contains a sufficiently long string which a lot of symbols point to (i.e. name offset is identical), the `slen` variable can be incremented past its `SIZE_MAX` limit or at least close enough that the size calculation afterwards overflows. This in turn would mean that our memory allocation is too small, subsequently leading to out ouf boundary writes.

evelikov: I think the updated PR should handle your initial comment. Please have another look - thanks in advance.

evelikov:
v2: quit looping when found, ELFDBG print missing sections

Seemingly only ~20% of modules have `__ksymtab_strings` and most modules have dozen+ sections. So the extra check helps a tiny bit there. As a bonus the malloc -> calloc change is no longer needed.

libkmod/libkmod-elf.c
```diff
@@ -248,6 +269,56 @@ static inline int elf_get_section_info(const struct kmod_elf *elf, uint16_t idx,
...
+		const char *n;
+		int err = elf_get_section_info(elf, i, &off, &size, &n);
+		if (err < 0)
+			continue;
+		if (streq("__ksymtab_strings", n)) {
+			elf->sections[KMOD_ELF_SECTION_KSYMTAB].offset = off;
+			elf->sections[KMOD_ELF_SECTION_KSYMTAB].size = size;
+			found_sec |= 1 << KMOD_ELF_SECTION_KSYMTAB;
+			continue;
+		}
```
stoeckmann, line 288: These if-blocks can be combined into one, put into another for-loop. The name itself can be retrieved through loop-variable from `section_name_map`. Having just one if-block makes it easier to audit the code.

stoeckmann: This code behaves differently than current implementation and kernel code. Right now, kmod and kernel take the first section information if multiple entries with the same name exist, but this code would take the last. To avoid any form of confusion between kmod and kernel, let's use only the first entry and add a warning through ELF_DBG (if you want to).

stoeckmann (changes requested):

libkmod/libkmod-elf.c
stoeckmann: Yes, this works fine. Maybe there are other occurrences in the code base, but nothing that belongs into this PR.

libkmod/libkmod-elf.c
stoeckmann: Manual creation of module files. You could set the name offset of an already existing section to the same offset of another one. Or just add a completely new section etc.

Definitely no valid files, but let's make sure that it's not possible to let kmod read other sections than the kernel would.

stoeckmann (approved):

lucasdemarchi:
Applied, thanks

PR #211, opened 2024-10-24
libkmod: Some cleanups

stoeckmann:
Collection of small cleanups gathered during the last weeks which did not properly fit into other PRs.

Mostly about code deduplication, removal of unused return values, and better error handling.

### DIFF
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -291,7 +291,6 @@ KMOD_EXPORT int kmod_module_new_from_name(struct kmod_ctx *ctx, const char *name
 int kmod_module_new_from_alias(struct kmod_ctx *ctx, const char *alias, const char *name,
 			       struct kmod_module **mod)
 {
-	int err;
 	char key[PATH_MAX];
 	size_t namelen = strlen(name);
 	size_t aliaslen = strlen(alias);
@@ -303,11 +302,7 @@ int kmod_module_new_from_alias(struct kmod_ctx *ctx, const char *alias, const ch
 	memcpy(key + namelen + 1, alias, aliaslen + 1);
 	key[namelen] = '\\';
 
-	err = kmod_module_new(ctx, key, name, namelen, alias, aliaslen, mod);
-	if (err < 0)
-		return err;
-
-	return 0;
+	return kmod_module_new(ctx, key, name, namelen, alias, aliaslen, mod);
 }
 
 KMOD_EXPORT int kmod_module_new_from_path(struct kmod_ctx *ctx, const char *path,
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -146,6 +146,7 @@ void kmod_module_parse_depline(struct kmod_module *mod, char *line)
 	p++;
 	for (p = strtok_r(p, " \t", &saveptr); p != NULL;
 	     p = strtok_r(NULL, " \t", &saveptr)) {
+		struct kmod_list *l_new;
 		struct kmod_module *depmod = NULL;
 		const char *path;
 		int err;
@@ -164,7 +165,12 @@ void kmod_module_parse_depline(struct kmod_module *mod, char *line)
 
 		DBG(ctx, "add dep: %s\n", path);
 
-		list = kmod_list_prepend(list, depmod);
+		l_new = kmod_list_prepend(list, depmod);
+		if (l_new == NULL) {
+			ERR(ctx, "could not add dependency for %s\n", mod->name);
+			goto fail;
+		}
+		list = l_new;
 	}
 
 	DBG(ctx, "%zu dependencies for %s\n", n, mod->name);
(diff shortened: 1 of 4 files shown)


libkmod/libkmod-module.c
```diff
@@ -586,12 +584,7 @@ KMOD_EXPORT const char *kmod_module_get_path(const struct kmod_module *mod)
 		return NULL;
 
 	/* lazy init */
-	line = kmod_search_moddep(mod->ctx, mod->name);
-	if (line == NULL)
-		return NULL;
-
-	kmod_module_parse_depline((struct kmod_module *)mod, line);
-	free(line);
+	module_get_dependencies_noref((struct kmod_module *)mod);
```
evelikov, line 579: No need to cast away the const here.

I see you're changing the signature with a latter commit. Let's introduce the cast with that commit.

libkmod/libkmod-module.c
```diff
@@ -105,30 +104,30 @@ static inline bool module_is_inkernel(struct kmod_module *mod)
...
 	if (p == NULL)
-		return 0;
+		return;
 
 	*p = '\0';
 	dirname = kmod_get_dirname(mod->ctx);
 	dirnamelen = strlen(dirname);
 	if (dirnamelen + 2 >= PATH_MAX)
-		return 0;
+		return;
```
evelikov, line 130: Should this be an error aka `mod->init.dep = false;`?

libkmod/libkmod-module.c
```diff
@@ -138,17 +137,18 @@ int kmod_module_parse_depline(struct kmod_module *mod, char *line)
 	if (mod->path == NULL) {
 		const char *str = path_join(line, dirnamelen, buf);
 		if (str == NULL)
-			return 0;
+			return;
```
evelikov, line 140: Ditto for these + the one just below.

libkmod/libkmod-module.c
```diff
@@ -138,17 +137,18 @@ int kmod_module_parse_depline(struct kmod_module *mod, char *line)
...
 		mod->path = strdup(str);
 		if (mod->path == NULL)
-			return 0;
+			return;
```
evelikov, line 143: We should probably `free(mod->path); mod->path = NULL;` in the error path.

Just like the others it's pre-existing to feel free to keep it separate patch/PR.

libkmod/libkmod-module.c
```diff
@@ -509,23 +506,17 @@ KMOD_EXPORT int kmod_module_get_filtered_blacklist(const struct kmod_ctx *ctx,
...
 
-		if (line == NULL)
-			return NULL;
-
-		kmod_module_parse_depline((struct kmod_module *)mod, line);
-		free(line);
-
-		if (!mod->init.dep)
-			return NULL;
+		if (line != NULL) {
```
evelikov, line 516: Don't think we need the NULL check here. Both `strchr(NULL, ':');` and `free(NULL)` are safe.

libkmod/libkmod-module.c
```diff
@@ -506,14 +506,14 @@ KMOD_EXPORT int kmod_module_get_filtered_blacklist(const struct kmod_ctx *ctx,
 	return kmod_module_apply_filter(ctx, KMOD_FILTER_BLACKLIST, input, output);
 }
 
-static void module_get_dependencies_noref(const struct kmod_module *mod)
+static void module_get_dependencies_noref(struct kmod_module *mod)
```
evelikov, line 510: AFAICT the premise is that since `kmod_module` and friends are opaque from the end user POV, modifying internal state should be safe (kind of).

That said, internal API which knows the distinction shouldn't be (ab)using that IMHO.

evelikov (changes requested):
Looking good. Left a few suggestions, with the pre-existing issues in `kmod_module_parse_depline` are optional.

libkmod/libkmod-module.c
stoeckmann: What makes you believe that `strchr(NULL, ':')` is safe? Definitely not the case for glibc and not part of the standard. Just make sure that the compiler does not outsmart you:

```
cat > search.c << EOF
#include <string.h>

char *search(char *p)
{
        return strchr(p, ':');
}
EOF
cat > poc.c << EOF
#include <stddef.h>

extern char *search(char *);

int
main(int argc, char **argv)
{
        search(NULL);
        return 0;
}
EOF
cc -c test.c poc.c
cc -o poc test.o poc.o
./poc
```

leads to

```
Segmentation fault (core dumped)
```

libkmod/libkmod-module.c
stoeckmann: Adjusted

libkmod/libkmod-module.c
evelikov: I failed at reading `man 3 strchr` it seems. Thanks for the correction :bow:

evelikov (approved):

lucasdemarchi:
Applied, thanks

Issue #214, opened 2024-10-29
Alias and dependency regressions since "libkmod: Use ELF offsets more often"


evelikov:
While rummaging for https://github.com/kmod-project/kmod/issues/213, I noticed that commit 25ab561bd59338992a2ab7288abbfecb1e1d018f ("libkmod: Use ELF offsets more often") has causes a regressions with older kernels. Namely [this one](https://archive.org/download/archlinux_pkg_linux-lts/linux-lts-3.10.12-1-x86_64.pkg.tar.xz) from the Arch/archive.org historical archive.

See the individual sections for the respective changes.

@stoeckmann can you have a look? Thanks in advance o/

<details>

<summary>Alias changes</summary>

```
--- /tmp/before/lib/modules/3.10.12-1-lts/modules.symbols	2024-10-29 14:00:36.206784440 +0000
+++ /tmp/after/lib/modules/3.10.12-1-lts/modules.symbols	2024-10-29 14:21:26.386554435 +0000
@@ -25,7 +25,6 @@
 alias symbol:v4l2_m2m_buf_remove v4l2_mem2mem
 alias symbol:vsock_addr_bound vsock
 alias symbol:fat_build_inode fat
-alias symbol:ks0108_isinited ks0108
 alias symbol:mmc_read_bkops_status mmc_core
 alias symbol:nand_lock nand
 alias symbol:scsi_add_device scsi_mod
@@ -46,7 +45,6 @@
 alias symbol:svc_proc_unregister sunrpc
 alias symbol:xdr_partial_copy_from_skb sunrpc
 alias symbol:crc7 crc7
-alias symbol:cxgb4_l2t_release cxgb4
 alias symbol:kvm_before_handle_nmi kvm
 alias symbol:nfs_instantiate nfs
 alias symbol:usb_remove_function libcomposite
@@ -88,7 +86,6 @@
 alias symbol:libipw_set_geo libipw
 alias symbol:snd_hda_apply_pincfgs snd_hda_codec
 alias symbol:stop_airo_card airo
-alias symbol:atm_dev_deregister atm
 alias symbol:cfg80211_wext_siwrts cfg80211
 alias symbol:copy_from_user_toio snd
 alias symbol:cx88_risc_databuffer cx88xx
@@ -125,9 +122,7 @@
 alias symbol:md_wakeup_thread md_mod
 alias symbol:rpc_mkpipe_data sunrpc
 alias symbol:si476x_core_cmd_am_acf_status si476x_core
[...]

Issue #213, opened 2024-10-25
Add CI step to check indexed before/after commits


evelikov:
Our test coverage is by no means extensive. Once quick and easy way to catch issues like https://github.com/kmod-project/kmod/issues/207 is to extend the CI by:
 - install kernel matching the existing kernel headers
 - build the indexes with and without the PR changes
 - compare the two and error on variance

This will not catch all potential problems, but it will greatly reduce the blast radius.

It will flag intentional changes like https://github.com/kmod-project/kmod/pull/188. Since those are bound to be very rare, we can ignore the CI. We can reconsider if they become common though.

lucasdemarchi:
Instead of building kmod without the PR applied, shouldn't we just use kmod from the distro and expect to be similar?

lucasdemarchi:
although then it would be hard to differentiate failures from "acceptable changes/improvements". While we don't have such a test, I think we should go back and make sure we didn't break anything after v33.

evelikov:
> Instead of building kmod without the PR applied, shouldn't we just use kmod from the distro and expect to be similar?

Assuming it's not too noisy, we can have both.

> although then it would be hard to differentiate failures from "acceptable changes/improvements".

My gut feeling is that noise should be once a year/couple of years If it proves more common, then we can look for solutions.

Generally I also considered the junit.xml that meson produces, nicely rendered in Github. Although since there is no have native support, we require third-party actions which ask for quite a lot of permissions... One day, but not now.

>  While we don't have such a test, I think we should go back and make sure we didn't break anything after v33.

Definitely on my todo list. Thanks o/

lucasdemarchi:
> Generally I also considered the junit.xml that meson produces, nicely rendered in Github. Although since there is no have native support, we require third-party actions which ask for quite a lot of permissions... One day, but not now.
> 
testlog.json would be more palatable I think, with quick reporting via a loop + jq or python script.

evelikov:
> I think we should go back and make sure we didn't break anything after v33.

Comparing v33 and 4891b4b095aee69d6afeff9eb2f6bb67b936e0f6, across different kernels:
 - 6.11 - identical
 - 5.15 - identical
 - 3.10 - missing some alias symbols and crc-foo dependencies et al

Kicked of a bisection to track things down.

evelikov:
While the regression is fixed, the testing plan (outlined here) is not enough to catch similar problems, in the future. 

[...]

PR #215, opened 2024-10-29
libkmod: Properly skip first symbol entry

stoeckmann:
This fixes a regression introduced while converting pointer to offset arithmetics.

The for-loop itself starts at 1 already, so reflect this with the manually performed offset + length calculation right at the start.

Fixes https://github.com/kmod-project/kmod/issues/214

At hindsight, this is fortunately rather obvious. Much better than a subtle error due to incorrect calculations.

This can only be seen in rather rare cases because the last symbol must have a special meaning. Fortunately @evelikov spotted it. Thanks!


lucasdemarchi:
Nice!! we will definitely need #213 to continue improving the index.

evelikov (approved):
Nit: please convert the "Fixes github issue" to "Closes: full-url" as documented in the contributions guide.

lucasdemarchi:
Applied, thanks

evelikov:
As a test coverage for this, we could use a 3.10 kernel as mentioned in the linked issue.

PR #217, opened 2024-10-31
meson: don't escape module_foo in kmod.pc

evelikov:
The kmod.pc file, contains module_signatures and module_compressions space-separated lists, indicating how kmod is built. For the meson generated file, any spaces are escaped where the autotools one does not do that.

Update the meson build to be consistent with the original autotools one and omit the escaping that we don't want.

Aside: seems like autotools does not escape the directory variables either, so if the path(s) have any spaces in them pkg-config (pkgconf at least) produces rubbish, quite rightfully IMHO.

Eg.
  $ cat .../autotools-file.pc
  prefix=/usr
  includedir=${prefix}/include bar
  Cflags: -I${includedir}
  ...
  $ pkg-config --cflags autotools-file.pc
  bar

  $ cat .../meson-file.pc
  prefix=/usr
  includedir=${prefix}/include\ bar
  Cflags: -I${includedir}
  ...
  $ pkg-config --cflags meson-file.pc
  -I/usr/include\ bar

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -446,19 +446,25 @@ _kmod_variables = [
   'sysconfdir=' + sysconfdir,
   'distconfdir=' + distconfdir,
   'module_directory=' + moduledir,
+]
+
+# Don't (space) escape variables with space-separated lists, for consistency
+# with the autotools build.
+_kmod_unescaped_variables = [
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
Applied, thanks

PR #219, opened 2024-11-01
Re-enable sanitizer on Fedora

evelikov:
Fedora 41 is out :partying_face: 

Being the first distro to ship clang 19, this time with actual sanitizer DSO for us to use :tada:. The filename has been changed and seemingly that went undocumented :facepalm:

With a tweak to the name + fallback for everyone else - things just work.

In before anyone asks: Why do we bother with checking for gcc vs clang, instead of just doing the `test -f` trick?

The `gcc` DSO usually lives within `$libdir`, thus is found, but not used, by `clang`. That may change in the future, plus there is no guarantee that the `clang` one won't be moved to `$libdir` at some point.

### DIFF
--- a/scripts/sanitizer-env.sh
+++ b/scripts/sanitizer-env.sh
@@ -7,9 +7,12 @@ if [[ ${CC-} == *gcc* ]]; then
 elif [[ ${CC-} == *clang* ]]; then
     OUR_PRELOAD=$("$CC" -print-file-name=libclang_rt.asan-x86_64.so)
 else
-    echo "Unknown compiler CC=\"${CC-}\" - gcc and clang are supported."
-    echo "Assuming \"gcc\", manually set the variable and retry if needed."
-    echo
+    cat <<- EOF
+
+    Unknown compiler CC="${CC-}" - gcc and clang are supported.
+    Assuming "gcc", manually set the variable and retry if needed.
+
+EOF
     OUR_PRELOAD=$(gcc -print-file-name=libasan.so)
 fi
 
@@ -23,7 +26,11 @@ if test -n "$OUR_PRELOAD"; then
 
     LD_PRELOAD=${LD_PRELOAD+${LD_PRELOAD}:}$OUR_PRELOAD
     export LD_PRELOAD
-    echo "LD_PRELOAD has been set to \"$LD_PRELOAD\"."
-    echo "The sanitizer might report issues with ANY process you execute."
+    cat <<- EOF
+
+    LD_PRELOAD has been set to "$LD_PRELOAD".
+    The sanitizer might report issues with ANY process you execute.
+
+EOF
 fi
 unset OUR_PRELOAD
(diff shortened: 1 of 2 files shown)


scripts/sanitizer-env.sh
```diff
@@ -25,5 +25,14 @@ if test -n "$OUR_PRELOAD"; then
     export LD_PRELOAD
     echo "LD_PRELOAD has been set to \"$LD_PRELOAD\"."
     echo "The sanitizer might report issues with ANY process you execute."
+else
+    echo
+    echo "WARNING: compiler returned non-existing library name $OUR_PRELOAD."
```
lucasdemarchi, line 34: it looks like this belongs to stderr?

if we use heredoc, we can simply redirect a single command to >&2

scripts/sanitizer-env.sh
```diff
@@ -5,7 +5,11 @@
 if [[ ${CC-} == *gcc* ]]; then
     OUR_PRELOAD=$("$CC" -print-file-name=libasan.so)
 elif [[ ${CC-} == *clang* ]]; then
-    OUR_PRELOAD=$("$CC" -print-file-name=libclang_rt.asan-x86_64.so)
+    # With v19, the library lacks the CPU arch in its name
+    OUR_PRELOAD=$("$CC" -print-file-name=libclang_rt.asan.so)
+    if ! test -f "$OUR_PRELOAD"; then
+       OUR_PRELOAD=$("$CC" -print-file-name=libclang_rt.asan-x86_64.so)
```
lucasdemarchi, line 11: humn... pre-existent issue:  I wonder if we should check arch instead of hardcoding x86_64

lucasdemarchi (changes requested):

scripts/sanitizer-env.sh
evelikov: Added extra commits to update the extra script and convert the Unknown compiler case to warn/stderr.

scripts/sanitizer-env.sh
evelikov: Considering that clang is moving way from it, I'd rather avoid adding more complexity unless we get bug reports.

scripts/sanitizer-env.sh
```diff
@@ -23,7 +26,11 @@ if test -n "$OUR_PRELOAD"; then
 
     LD_PRELOAD=${LD_PRELOAD+${LD_PRELOAD}:}$OUR_PRELOAD
     export LD_PRELOAD
-    echo "LD_PRELOAD has been set to \"$LD_PRELOAD\"."
-    echo "The sanitizer might report issues with ANY process you execute."
+    cat <<- EOF
+
+    LD_PRELOAD has been set to \"$LD_PRELOAD\"."
```
lucasdemarchi, line 35: One leftover `"`.... But you also don't need to escape them, so just remove the escapes and the extra one

lucasdemarchi (approved):
Other than the comment, LGTM. Thanks.

scripts/sanitizer-env.sh
evelikov: Nicely spotted. Updated + added some quoted the `OUR_PRELOAD` message when it's not a file.

lucasdemarchi:
Applied, thanks

PR #222, opened 2024-11-01
libkmod: Verify that symbol names in ELF files are NUL-terminated

stoeckmann:
We have to verify that symbol names are NUL-terminated, i.e. use strnlen instead of strlen, especially before calling streq.

Proof of Concept (ideally, compile with ASAN):

1. Create a module with an unterminated symbol name
```
cat > poc.ko.zst.b64 << EOF
KLUv/WQAD3UDABQEf0VMRgICABERIiIzMzMzRFUAQHd3d3eIiJmZqqoAQAADAAERAAAACCIzAAEA
C2ZmZmaZDEBfX3ZlcnNpb25zQUIUIFADLXEDsIYvhM89CCM4cnnoZbcR83mE/9ATG3MnjN9ZCcpg
iHzzKaoRVssS5Q==
EOF
base64 -d poc.ko.zst.b64 > poc.ko.zst
```

2. Try to show modversion information
```
modprobe --show-modversions poc.ko.zst
```

If you do not compile with ASAN, uncompress `poc.ko.zst` so `mmap` will be called. Most likely you see invalid data or encounter a crash.

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -461,7 +461,7 @@ int kmod_elf_get_strings(const struct kmod_elf *elf, const char *section, char *
 }
 
 static inline void elf_get_modversion_lengths(const struct kmod_elf *elf, size_t *verlen,
-					      size_t *crclen)
+					      size_t *crclen, size_t *namlen)
 {
 	assert_cc(sizeof(struct kmod_modversion64) == sizeof(struct kmod_modversion32));
 
@@ -470,24 +470,26 @@ static inline void elf_get_modversion_lengths(const struct kmod_elf *elf, size_t
 
 		*verlen = sizeof(*mv);
 		*crclen = sizeof(mv->crc);
+		*namlen = sizeof(mv->name);
 	} else {
 		struct kmod_modversion64 *mv;
 
 		*verlen = sizeof(*mv);
 		*crclen = sizeof(mv->crc);
+		*namlen = sizeof(mv->name);
 	}
 }
 
 /* array will be allocated with strings in a single malloc, just free *array */
 int kmod_elf_get_modversions(const struct kmod_elf *elf, struct kmod_modversion **array)
 {
-	size_t off, crclen, slen, verlen;
+	size_t off, crclen, namlen, slen, verlen;
 	uint64_t sec_off, size;
 	struct kmod_modversion *a;
 	char *itr;
 	int i, count, err;
 
-	elf_get_modversion_lengths(elf, &verlen, &crclen);
+	elf_get_modversion_lengths(elf, &verlen, &crclen, &namlen);
 
 	*array = NULL;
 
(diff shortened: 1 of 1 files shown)


libkmod/libkmod-elf.c
```diff
@@ -476,10 +476,12 @@ int kmod_elf_get_modversions(const struct kmod_elf *elf, struct kmod_modversion
 		struct kmod_modversion32 *mv;
 
 		crclen = sizeof(mv->crc);
+		namlen = sizeof(mv->name);
 	} else {
 		struct kmod_modversion64 *mv;
 
 		crclen = sizeof(mv->crc);
+		namlen = sizeof(mv->name);
```
evelikov, line 484: Instead of duplicating this even more, let's make it a helper function.

The `verlen` piece (seen below) can become a static assert (`assert_cc` in kmod), since 32/64bit versions of the struct are intentionally kept identical.

libkmod/libkmod-elf.c
```diff
@@ -501,11 +503,15 @@ int kmod_elf_get_modversions(const struct kmod_elf *elf, struct kmod_modversion
 
 	for (i = 0; i < count; i++, off += MODVERSION_SEC_SIZE) {
 		const char *symbol = elf_get_mem(elf, off + crclen);
+		size_t nlen = strnlen(symbol, namlen);
+
+		if (nlen == namlen)
```
evelikov, line 508: Nit: let's add a `ELFDBG` for all instances, not just the `get_dependency_symbols` one.

libkmod/libkmod-elf.c
```diff
@@ -1101,8 +1116,20 @@ int kmod_elf_get_dependency_symbols(const struct kmod_elf *elf,
...
 				name = elf_get_mem(elf, ver_off + i * verlen + crclen);
-				slen += strlen(name) + 1;
+				nlen = strnlen(name, namlen);
+
+				if (nlen == namlen) {
+					ELFDBG(elf, "symbol name at index %d too long\n",
+					       i);
+					free(visited_versions);
+					free(symcrcs);
+					*array = NULL;
```
evelikov, line 1129: Nit: no need to clear output variable on error.

evelikov (changes requested):
Not sure if coffee has kicked in yet, but we seem to be missing step 2 in the reproducer.

libkmod/libkmod-elf.c
stoeckmann: I've moved it into its own inline function, keeping `assert_cc` and `verlen`. Since the function is inlined and compilers are smart enough, the verlen is a constant anyway, so the preprocessor definition can be removed instead.

libkmod/libkmod-elf.c
stoeckmann: Added

libkmod/libkmod-elf.c
stoeckmann: Removed

libkmod/libkmod-elf.c
evelikov: Speaking on inline: We should probably nuke the "Wno-inline" flag ... on a rainy day.

libkmod/libkmod-elf.c
evelikov: Can we get one for the `kmod_elf_crc_find()` case as well?

evelikov (approved):
I'd probably call the new helper something like `elf_get_modversion_lengths` or `elf_get_modversion_member_lengths`. But in either case, the PR look good.

libkmod/libkmod-elf.c
stoeckmann: That's what I get when I just look for `nlen == namlen`, hah. Adjusted

stoeckmann:
> `elf_get_modversion_lengths`

Chose this one, adjusted.

lucasdemarchi (approved):

lucasdemarchi:
Applied, thanks

PR #224, opened 2024-11-04
Index bugfix and papercuts

evelikov:
While looking at how to make the index API properly handle error conditions, I noticed that we completely broke the API.

The first commit fixes that + reintroduces earlier proposal of mine. The initial counter argument had a good point but was a bit absolute (aka thinking of a extreme case, that isn't applicable to us).

The rest are bunch of paper cuts, that I've accumulated as part of the initial goal.

/cc @stoeckmann can you check the first commit? can I offer some :cookie: for adding an in-tree test?

Random notes:
 - there is only one external user (AFAICT) of the regressed API :sweat_smile: 
 - a "test-needed" label to tag PRs (merged or not) would be awesome
   - cannot create labels or I would have done it already :cry:  
 - `index_value::priority` is ignored when dumping... unlike the `depmod` equivalent
 - de-duplicating and dog-fooding the dumping API across libkmod and depmod
   - introduce new one and deprecate the old?

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -613,7 +613,7 @@ struct index_mm_value {
 };
 
 struct index_mm_node {
-	struct index_mm *idx;
+	const struct index_mm *idx;
 	const char *prefix; /* mmap'ed value */
 	unsigned char first;
 	unsigned char last;
@@ -657,7 +657,8 @@ static inline void read_value_mm(const void **p, struct index_mm_value *v)
 }
 
 /* reads node into given node struct and returns its address on success or NULL on error. */
-static struct index_mm_node *index_mm_read_node(struct index_mm *idx, uint32_t offset,
+static struct index_mm_node *index_mm_read_node(const struct index_mm *idx,
+						uint32_t offset,
 						struct index_mm_node *node)
 {
 	const void *p;
@@ -797,7 +798,7 @@ void index_mm_close(struct index_mm *idx)
 	free(idx);
 }
 
-static struct index_mm_node *index_mm_readroot(struct index_mm *idx,
+static struct index_mm_node *index_mm_readroot(const struct index_mm *idx,
 					       struct index_mm_node *root)
 {
 	return index_mm_read_node(idx, idx->root_offset, root);
@@ -854,7 +855,7 @@ static void index_mm_dump_node(struct index_mm_node *node, struct strbuf *buf, i
 	strbuf_popchars(buf, pushed);
 }
 
-void index_mm_dump(struct index_mm *idx, int fd, bool alias_prefix)
+void index_mm_dump(const struct index_mm *idx, int fd, bool alias_prefix)
 {
 	struct index_mm_node nbuf, *root;
 	struct strbuf buf;
@@ -911,7 +912,7 @@ static char *index_mm_search_node(struct index_mm_node *node, const char *key)
  *
  * Returns the value of the first match
  */
-char *index_mm_search(struct index_mm *idx, const char *key)
+char *index_mm_search(const struct index_mm *idx, const char *key)
 {
 	// FIXME: return value by reference instead of strdup
 	struct index_mm_node nbuf, *root;
@@ -1035,7 +1036,7 @@ static void index_mm_searchwild_node(struct index_mm_node *node, struct strbuf *
  *
  * Returns a list of all the values of matching keys.
  */
-struct index_value *index_mm_searchwild(struct index_mm *idx, const char *key)
+struct index_value *index_mm_searchwild(const struct index_mm *idx, const char *key)
 {
 	struct index_mm_node nbuf, *root;
 	struct strbuf buf;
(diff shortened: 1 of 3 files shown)


libkmod/libkmod-index.c
```diff
@@ -401,7 +401,7 @@ void index_dump(struct index_file *in, int fd, const char *prefix)
 		return;
 
 	strbuf_init(&buf);
-	if (strbuf_pushchars(&buf, prefix))
+	if (prefix[0] = '\0' || strbuf_pushchars(&buf, prefix))
```
stoeckmann: If the API was "completely" broken, this breaks the other half. ;)

This effectively sets the prefix to '\0' instead of checking for it, i.e. `=` vs `==`.

Great work spotting this regression though!

libkmod/libkmod-index.c
```diff
@@ -866,7 +866,7 @@ void index_mm_dump(struct index_mm *idx, int fd, const char *prefix)
...
```
stoeckmann: And in the mm version as well.

libkmod/libkmod-index.c
```diff
@@ -451,7 +451,6 @@ static char *index_search__node(struct index_node_f *node, const char *key, int
  */
 char *index_search(struct index_file *in, const char *key)
 {
-	// FIXME: return value by reference instead of strdup
```
stoeckmann, line 454: I'd keep the comment. If @lucasdemarchi adds a strbuf backed by scratchbuf, then it might be possible to skip the strdup. After all the index_search function could trigger an allocation, while index_mm_search would initialize the scratch buffer with a pointer and proper size, so no free would be needed.

Haven't made any further progress in this, because it's one strdup per module insertion, too much exclusive scratchbuf-code to add for this and would duplicate future work.

libkmod/libkmod-index.c
```diff
@@ -913,9 +911,8 @@ static char *index_mm_search_node(struct index_mm_node *node, const char *key)
  *
  * Returns the value of the first match
  */
-char *index_mm_search(struct index_mm *idx, const char *key)
+char *index_mm_search(const struct index_mm *idx, const char *key)
 {
-	// FIXME: return value by reference instead of strdup
```
stoeckmann, line 918: Same as above. I would skip the commit.

libkmod/libkmod.c
```diff
@@ -444,6 +445,15 @@ static char *lookup_builtin_file(struct kmod_ctx *ctx, const char *name)
 	return line;
 }
 
+static bool lookup_builtin_file(struct kmod_ctx *ctx, const char *name)
+{
+	char *line = lookup_file(ctx, KMOD_INDEX_MODULES_BUILTIN, name);
+	bool found = line != NULL;
+
+	free(line);
+	return found;
```
stoeckmann, line 454: You can simplify this function by using `_cleanup_free_ char *line` and just `return line != NULL`.

libkmod/libkmod-index.c
evelikov: Yeah, seems like our dump coverage is worse than I thought.

libkmod/libkmod-index.c
evelikov: Sure, dropped.

libkmod/libkmod.c
evelikov: That was my initial plan although scan-build (aka clang analyse) does not properly handle `__attribute__((cleanup))` and flags it as memory leak.

I am aiming to get ^^ as part of the CI and this gets us a step closer.

stoeckmann: We have this construct already in code base, see: https://github.com/kmod-project/kmod/blob/master/libkmod/libkmod.c#L498

I don't like the idea to keep existing features in a more expressive way just to make an analyzer happy, since it must complain about the existing one already, too.

stoeckmann (changes requested):

libkmod/libkmod.c
evelikov: Yes, I'm aware that we have ~dozen instances in-tree. Pretty much all of those get flagged by clang-analyse.

Similar to the analysers and generally having wider compiler coverage - there is more value in biting the bullet "fixing" a few instances and keeping the tools happy IMHO.

Ideally clang will get fixed - if you want to give it a go, that would be deeply appreciated.

stoeckmann: So do you plan to remove the existing feature? Right now, it's a mix. Doesn't seem worthwhile for anyone.

lucasdemarchi: nack on removal

lucasdemarchi:
@stoeckmann  are you ok with this version?

stoeckmann:
> @stoeckmann are you ok with this version?

Won't stop merge process. I'd prefer `_cleanup_free_` but it's not like this is the only function which does it manually.

lucasdemarchi:
@stoeckmann agreed. I created #233  so we don't forget. Let's see what is the clang-analyzer situation with that. Applied, thanks.

evelikov:
We need tests for `kmod_dump_index` with all the different indexen.

PR #226, opened 2024-11-04
Sort the compiler flags, remove most of -Wno

evelikov:
Nothing too spectacular here ... Although the sorting does break the Pink Floyd meme ;-)

Seems like the Android folks have copied it literally - too subtle?


lucasdemarchi:
> Nothing too spectacular here ... Although the sorting does break the Pink Floyd meme ;-)

ok... I added a commit on top. It was hard to let it go: it was one of the first songs I learned to play on guitar.
But it's the right thing to do.

evelikov:
Ftr I don't mind keeping it and/or moving things a bit so it makes sense.

I even tried keeping it in meson, although meson complains loudly (with no way to silence) when we manually try to use `-Wall` :-(

lucasdemarchi:
ae54b2b5-df41-41ce-8d91-c7c58de1c8e5

PR #229, opened 2024-11-06
Fixup code coverage, add to CI

evelikov:
This PR starts off with a trivial annoyance fix the last day or so - removing the tests, use Ubuntu 24 for CodeQL.

Then an earlier commit of mine is reverted which broke the code coverage and finally a CI action is added. For the last piece a CodeCov account is needed + trivial setup for the `CODECOV_TOKEN`.

Edit: here is an illustrative link to the CodeCov web UI https://app.codecov.io/github/evelikov/kmod

### DIFF
--- a/.github/actions/setup-debian/action.yml
+++ b/.github/actions/setup-debian/action.yml
@@ -11,6 +11,8 @@ runs:
     - name: Install dependencies
       shell: bash
       run: |
+        export DEBIAN_FRONTEND=noninteractive
+        export TZ=Etc/UTC
         apt-get update
         apt-get install --yes \
           autoconf \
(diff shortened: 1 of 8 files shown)


evelikov:
If we find the PR notifications to be too much (say thee threshold is too low so new PRs are repeatedly flagged as low coverage), we can tweak them via the [codecov.yml file](https://docs.codecov.com/docs/codecovyml-reference).

evelikov:
As indicated the Ubuntu 22.04 is still semi-borked and I'm somewhat tempted to remove it. All other distros have one version/variant and the newer Ubuntu 24.04 seems stable - famous last words...

lucasdemarchi:
Added the same commit as in #229 here. Let's see if it passes now.

evelikov:
I think you meant #235 and yes it does. Thanks a bunch

lucasdemarchi:
Applied, thanks

PR #233, opened 2024-11-07
libkmod: Simplify lookup_builtin_file()

lucasdemarchi:
Use cleanup attribute.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -279,6 +279,7 @@ CC_CHECK_FLAGS_APPEND(with_cflags, [CFLAGS], [\
 		       -Wmissing-noreturn \
 		       -Wmissing-prototypes \
 		       -Wnested-externs \
+		       -Wno-attributes=clang::suppress \
 		       -Wno-unused-parameter \
 		       -Wold-style-definition \
 		       -Wpointer-arith \
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -241,7 +241,7 @@ static int kmod_config_add_blacklist(struct kmod_config *config, const char *mod
 
 	DBG(config->ctx, "modname=%s\n", modname);
 
-	p = strdup(modname);
+	_clang_suppress_alloc_ p = strdup(modname);
 	if (!p)
 		return -ENOMEM;
 
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -828,7 +828,7 @@ static int module_do_install_commands(struct kmod_module *mod, const char *optio
 		size_t suffixlen = cmdlen - prefixlen - varlen;
 		size_t slen = cmdlen - varlen + options_len;
 		char *suffix = p + varlen;
-		char *s = malloc(slen + 1);
+		_clang_suppress_alloc_ char *s = malloc(slen + 1);
 		if (!s)
 			return -ENOMEM;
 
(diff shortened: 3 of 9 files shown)


evelikov:
Tl:Dr: `_cleanup_` coverage is at ~1%, with 1 instance added in ~450 patches since 2024/01/01 yet it contributes to 60% of false reports.

Longer version:
I guess we can step back and look at the bigger picture: how much do we use `_cleanup_` and does it outweigh the benefits of adding analyser like `scan-build` (aka clang analyze)?

 - looking around - we had 2 new instances over the last 1.5 years (both of those my me :stuck_out_tongue:) and third one back in 2018. In terms of count there are only 19 instances in the whole codebase, while they could easily be 10-20x as much.

- glance at recent~ish `scan-build` report shows 38 issues, 23 of which from our use of `_cleanup_` - the rest, at least in part, are genuine - writing to a `malloc(0)` block, passing `NULL to `_nonnull_` annotated API, etc

Objectively looking our `_cleanup_` use is at 1%, while it contributes ~60% of noise to a `scan-build` run. This doesn't seems worth it.

On the subjective side - it's sad and annoying to remove a _really_ cool feature. Personally, if `scan-build` properly supported the attribute, I would have used a lot more than in 2 out of 200+ patches. Alas that's not the case ...

Plus there is nothing blocking us from re-introducing it later on when scan-build gets wiser.

lucasdemarchi:
I'm still of the opinion of scan-build needs to get wiser or we need to convince it of a proper way to understand it. Because it seems they are already even checking for improper use of the cleanup attribute. How the heck they don't understand what the attribute does?

You are creating a chicken and egg situation: it doesn't get used more because of X and X doesn't improve because it's not used enough. When something is that dumb, I prefer to simply disable it and use something smarter to cover the same thing. We can simply disable the unix.Malloc check and rely on valgrind for that.

Another alternative: try to convince the analyzer to stop bothering us with those false positives

evelikov:
Fully agree that clang-analyze should get wiser - yet again I'm just the messenger here. The chicken and egg seems valid if clang devs see kmod (plus other users like systemd) as moderate-high interest/priority project. Hope I don't sound like a cold-hearted a** here - trying to be realistic.

Thanks for the `::suppress` trick - didn't knew it existed :bow: 

At a glance it seems that the allocation needs to be annotated, not the variable declaration. While moving the assignment/allocation on declaration works, we cannot reliably use it where we don't know the size early "enough" since we explicitly ban declaration after statements.

If we remove `-Wdeclaration-after-statement` (I'm in favour btw) this would help us move things as needed.

lucasdemarchi:
See the last commit with the fixup where I return it to where it was. The problem is, like I said, even with those it seems like the leak check in clang-analyzer is not smart enough and has other problems. To the point I think we'd be better off turning off that specific check.

evelikov:
Apologies - had the reply written on the day before and didn't refresh/check before posting :bow: 

Looking at the 23 mem leak reports we have:
 - attribute cleanup related 10
 - kmod_list ownership transfer 7
 - legitimate 6 - some of which in the tests

Suspect we could "fix" the kmod_list ones although it'll be alot of churn. Plus there is a 10+ year [old issue](https://github.com/llvm/llvm-project/issues/4260) about the cleanup attribute with limited activity.

So despite the legitimate catches, the noise is too much to have `unix.Malloc` enabled by default :sob:

evelikov (approved):
Thumbs up for reintroducing the `_cleanup_free_` instance. The clang/suppress patch can be deferred for a day when the false-positives are fewer.

lucasdemarchi:
Yeah, I think I will just check for running with clang_analyzer and remove the attribute check.

lucasdemarchi:
Applied, thanks

PR #234, opened 2024-11-07
libkmod: Avoid overflows while parsing files

stoeckmann:
Specially crafted files could overflow internal counters, allowing out of boundary writes. Make sure that neither counters nor resulting calculations overflow.

Another round of "is it even possible?"

Proof of Concept:

1. Create a file which contains more than 2^32 pre dependencies
```
(echo -n "softdep poc pre: "; yes a | tr '\n' ' ' | dd bs=4096 count=2097153; echo) > poc.conf
```

2. Try to insert poc module
```
modprobe -C poc.conf poc
```

This should lead to a segmentation fault or
```
SUMMARY: AddressSanitizer: heap-buffer-overflow ../libkmod/libkmod-config.c:371 in kmod_config_add_softdep
```

The file will be around 8 GB in size, memory peaks around that as well (didn't measure exactly, better have 16 GB or more). So, yes. It can happen. It's highly unlikely but we have the overflow checks already available, so let's just check instead of adjusting various data types and return values from `unsigned int` to `size_t`. Not worth it.

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -269,7 +269,7 @@ static int kmod_config_add_softdep(struct kmod_config *config, const char *modna
 	char *itr;
 	unsigned int n_pre = 0, n_post = 0;
 	size_t modnamelen = strlen(modname) + 1;
-	size_t buflen = 0;
+	size_t buflen = 0, size, size_pre, size_post;
 	bool was_space = false;
 	enum { S_NONE, S_PRE, S_POST } mode = S_NONE;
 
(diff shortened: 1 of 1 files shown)


libkmod/libkmod-config.c
```diff
@@ -318,9 +328,21 @@ static int kmod_config_add_softdep(struct kmod_config *config, const char *modna
...
 	DBG(config->ctx, "%u pre, %u post\n", n_pre, n_post);
 
-	dep = malloc(sizeof(struct kmod_softdep) + modnamelen +
-		     n_pre * sizeof(const char *) + n_post * sizeof(const char *) +
-		     buflen);
+	/*
+	 * sizeof(struct kmod_softdep) + modnamelen + buflen +
+	 * n_pre * sizeof(const char *) + n_post * sizeof(const char *)
+	 */
+	if (umulsz_overflow(n_pre, sizeof(const char *), &size_pre) ||
```
evelikov, line 335: Here and below - let's keep the ordering identical across the comment and code.

evelikov (approved):
Thank you

lucasdemarchi:
Applied, thanks

PR #235, opened 2024-11-07
ci: Fix debian/ubuntu installation

lucasdemarchi:
Make sure we are running in non-interactive mode and that we don't get blocked on tzdata update.


evelikov (approved):
I had a local patch for `DEBIAN_FRONTEND` that I was going to send out, but didn't knew about TZ.

Yes, please

lucasdemarchi:
Applied, thanks

PR #239, opened 2024-11-12
Drop scratchbuf

lucasdemarchi:
Slightly tested - probably need more CI and eyeballs.


lucasdemarchi:
@stoeckmann this is what I mentioned some time ago about doing with the scratchbuf impl.

shared/strbuf.c
```diff
@@ -55,12 +55,11 @@ char *strbuf_steal(struct strbuf *buf)
...
 	char *bytes;
 
-	if (!buf_realloc(buf, buf->used + 1)) {
-		free(buf->bytes);
+	if (!buf_realloc(buf, buf->used + 1))
 		return NULL;
-	}
 
 	bytes = buf->bytes;
+	buf->bytes = NULL;
```
evelikov, line 77: The previous behaviour of "we failed to realloc, let's nuke everything" meant we had an UAF waiting to happen. I'm glad it's gone plus the extra NULL helps truly invalidate the strbuf.

Hell yeah \o/

shared/strbuf.c
```diff
@@ -8,6 +8,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/param.h>
```
evelikov, line 11: OOC: why do we need the extra header?

shared/strbuf.c
```diff
@@ -18,10 +19,20 @@ static bool buf_realloc(struct strbuf *buf, size_t sz)
 {
 	void *tmp;
 
-	tmp = realloc(buf->bytes, sz);
-	if (sz > 0 && tmp == NULL)
-		return false;
+	if (!buf->heap)
+		tmp = malloc(sz);
+	else
+		tmp = realloc(buf->bytes, sz);
```
evelikov, line 25: For any fans of the ternary operator, this could be written as `tmp = realloc(!buf->heap ? NULL : buf->bytes, sz);`.

shared/strbuf.h
```diff
@@ -12,10 +13,24 @@ struct strbuf {
...
 void strbuf_init(struct strbuf *buf);
 
+/*
+ * Initialize strbuf with an initial buffer on stack. The @sz must be a build-time
+ * constant, as if the buffer had been declared on stack.
+ */
+#define strbuf_init_with_stack(buf__, sz__)            \
+	({                                             \
+		assert_cc(__builtin_constant_p(sz__)); \
+		(buf__)->bytes = alloca(sz__);         \
```
evelikov, line 28: Was thinking about adding `-Walloca` because of the obvious fallouts with it. The `__builtin_constant_p` does help with those, but I wonder if platforms like Android/macOS or other C runtimes don't impose restrictions on its use.

As a potential replacement we could use a `static char foo[sz__];`, which will invalidate the documented caveat - aka "It may also not survive a return from current function if it was initialized with stack space" can go.

This will grow the BSS (non-issue IMHO) although I wonder if it'll have other side-effects.

Mostly thinking out loud, than an objection.

shared/strbuf.c
```diff
@@ -39,15 +39,17 @@ static bool buf_realloc(struct strbuf *buf, size_t sz)
 	return true;
 }
 
-static bool buf_grow(struct strbuf *buf, size_t newsize)
+bool strbuf_reserve(struct strbuf *buf, size_t n)
```
evelikov, line 42: Would be great if we can have a way to instrument the compiler/analyzers that follow-up `strbuf_str()` and `strbuf_push*()` calls cannot fail, when reservation large enough. As humans we'll likely get it wrong at some point.

Any clever trick that comes to mind?

shared/strbuf.c
```diff
@@ -82,9 +82,11 @@ char *strbuf_steal(struct strbuf *buf)
 
 const char *strbuf_str(struct strbuf *buf)
 {
-	if (!strbuf_reserve(buf, 1))
```
evelikov: `s/re-allocated/shrunk/` in the commit message, to make it more obvious?

tools/depmod.c
```diff
@@ -2306,32 +2304,37 @@ static int output_symbols_bin(struct depmod *depmod, FILE *out)
...
+		size_t sym_sz;
 
 		if (sym->owner == NULL)
 			continue;
 
-		len = strlen(sym->name);
+		sym_sz = strlen(sym->name) + 1;
 
-		if (scratchbuf_alloc(&salias, baselen + len + 1) < 0) {
+		if (!strbuf_reserve(&salias, sym_sz)) {
```
evelikov, line 2320: Why did we drop `baselen` from the total/reservation size?

tools/depmod.c
```diff
@@ -2289,10 +2288,9 @@ static int output_symbols(struct depmod *depmod, FILE *out)
 static int output_symbols_bin(struct depmod *depmod, FILE *out)
 {
 	struct index_node *idx;
-	char aliasbuf[1024] = "symbol:";
-	_cleanup_(scratchbuf_release) struct scratchbuf salias =
-		SCRATCHBUF_INITIALIZER(aliasbuf);
-	const size_t baselen = sizeof("symbol:") - 1;
+	_cleanup_strbuf_ struct strbuf salias;
+	const char *base = "symbol:";
+	const size_t baselen = strlen(base);
```
evelikov, line 2293: The project commonly uses `sizeof("string-literal") - 1` instead of `strlen("string-literal")`. I am guessing that's because older compilers could not deduce/optimise the latter to a compile-time constant.

Is that right, or am I missing something obvious?

evelikov (commented):
Have a slight inclination towards not using `alloca` plus the `strbuf_reserve()` instance added at the end seems off.

As a whole, I like the idea - esp the robustness fixes for `strbuf_steal()`

testsuite/test-strbuf.c
```diff
@@ -62,6 +62,10 @@ static int test_strbuf_pushchars(const struct test *t)
 		lastwordlen = strlen(c);
 	}
 
+	/*
+	 * Replace the last space char, which also guarantees there's at least 1 char
+	 * available for the '\0'
```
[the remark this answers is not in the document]
lucasdemarchi: Not sure why we'd move to shared/. The test does a popchar to remove the space and at the same time guarantees there's space for the NUL byte. Do you mean adding documentation to the popchar function?

shared/strbuf.h
lucasdemarchi: The static variable is a no go. We don't ever want a a "stack variable" to implicitly survive a return from the function.

Alternative IMO is what scratchbuf does with the additional variable. We could define a DECLARE_STRBUF_WITH_STACK that does that and hide the variable under a unique name.

shared/strbuf.c
lucasdemarchi: MIN()

shared/strbuf.c
lucasdemarchi: I tolerate it, I'm not a fan :).... here I'm ok if we flip them:  `tmp = realloc(buf->heap ? buf->bytes : NULL, sz);`

shared/strbuf.h
lucasdemarchi: > Was thinking about adding `-Walloca` because of the obvious fallouts with it.

in this case we'd just add some _Pragma() to push/pop -Walloca. There are uses for it, this being one of them.  What I like about not using alloca() here is actually that the buffer would be defined together with the declaration rather than together with strbuf_init_with_stack().

tools/depmod.c
lucasdemarchi: scratchbuf_alloc() passes the total space. strbuf_reserve() reserves **adiditional** space... here we don't need to track the total space since we add popchars/popchar where needed. On a hindsight just _reserve() may not be clear enough about what's the behavior. std::vector::reserve() is actually about the total sapce. Linux kernel's xarray is about 1 position.

Maybe add 2 functions?
strbuf_reserve() / strbuf_reserve_total()
strbuf_alloc() / strbuf_alloc_total()
strbuf_ensure_space() / strbuf_ensure_total_space()
strbuf_add() /

tools/depmod.c
lucasdemarchi: that's right

tools/depmod.c
evelikov: Indeed C++ came to mind when I read `strbuf_reserve()` and didn't think too much about it :facepalm: The xarray one never crossed my mind, since it takes an instance/index and not a size as an argument.

Having two functions makes sense. Alternatively we can use something like `strbuf_reserve_extra()` to indicate the "in-addition to" aspect. Don't have any strong preference fwiw.

shared/strbuf.h
evelikov: Same - scope and lifetime of using alloca() is a bit tricky. The DECLARE_STRBUF_WITH_STACK approach seems reasonable IMHO.

testsuite/test-strbuf.c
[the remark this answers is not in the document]
evelikov: > Do you mean adding documentation to the popchar function?

Precisely - the newly added comment looks like a very good candidate.

shared/strbuf.h
stoeckmann: > This will grow the BSS (non-issue IMHO) although I wonder if it'll have other side-effects.

A side-effect occurs in multi-threaded environments, because then multiple threads share the same static variable. Definitely undefined behavior, so that wouldn't be an option for library code.

stoeckmann: The good thing about `alloca` hidden in a macro here is that we have an "anonymous" stack memory which we can never accidentally access again like it happened before in depmod. I think it's the first reasonable example of `alloca` usage I've seen so far and with these checks in place, it cannot eventually overflow.

I'm more concerned about the lifetime here. And it's no difference if it's a named and easily accessible array on the stack or something hidden behind a macro.

tools/depmod.c
```diff
@@ -1308,14 +1306,13 @@ static int depmod_modules_search_dir(struct depmod *depmod, DIR *d, size_t basel
...
-		if (scratchbuf_alloc(s_path, baselen + namelen + 2) < 0) {
+		if (!strbuf_reserve(path, namelen + 2)) {
 			err = -ENOMEM;
 			ERR("No memory\n");
 			continue;
 		}
 
-		path = scratchbuf_str(s_path);
-		memcpy(path + baselen, name, namelen + 1);
+		strbuf_pushmem(path, name, namelen);
```
stoeckmann, line 1315: Since `namelen = strlen(name)`, this can become `strbuf_pushchars`.

stoeckmann: Adding this note as a second comment to the same line: The API of strbuf as it was to be used before, would expect a check to see if adding something to strbuf was successful or not. Even if we could make a static analyzer smart enough to see that it fits ... Is it our job as a strbuf user to care about its internal status? A simple approach would be to wrap all strbuf functions with a check and we are done. Everything else expects us to be smart ... And at worst smarter than we really are. :)

tools/depmod.c
```diff
@@ -1351,53 +1348,51 @@ static int depmod_modules_search_dir(struct depmod *depmod, DIR *d, size_t basel
...
 		err = -ENOMEM;
 		goto out;
 	}
-	path_buf = scratchbuf_str(&s_path_buf);
 
-	memcpy(path_buf, path, baselen);
-	path_buf[baselen] = '/';
-	baselen++;
-	path_buf[baselen] = '\0';
+	strbuf_pushmem(&s_path_buf, path, baselen - 1);
```
stoeckmann, line 1394: `baselen - 1` is `strlen(path) + 1 - 1` is `strlen(path)` so we can use `strbuf_pushchars` here as well.

tools/depmod.c
```diff
@@ -2310,32 +2304,37 @@ static int output_symbols_bin(struct depmod *depmod, FILE *out)
 
 	hash_iter_init(depmod->symbols, &iter);
 
+	strbuf_init_with_stack(&salias, 1024);
+	strbuf_pushmem(&salias, base, baselen);
```
stoeckmann, line 2308: Here we can use `strbuf_pushchars` as well because `baselen = strlen(base)`

tools/depmod.c
```diff
@@ -2310,32 +2304,37 @@ static int output_symbols_bin(struct depmod *depmod, FILE *out)
...
+		if (!strbuf_reserve(&salias, sym_sz)) {
 			ret = -ENOMEM;
-			goto err_scratchbuf;
+			goto err_alloc;
 		}
-		s = scratchbuf_str(&salias);
-		memcpy(s + baselen, sym->name, len + 1);
-		duplicate = index_insert(idx, s, sym->owner->modname, sym->owner->idx);
+
+		strbuf_pushmem(&salias, sym->name, sym_sz);
```
stoeckmann, line 2325: `sym_sz` is `strlen(sym->name) + 1` so here we already add the ending NUL byte. If we skip it, even this one could become `strbuf_pushchars`.

shared/strbuf.c
```diff
@@ -8,75 +8,111 @@
...
+	buf->heap = true;
 	buf->bytes = tmp;
 	buf->size = sz;
+
 	return true;
 }
 
+bool strbuf_reserve(struct strbuf *buf, size_t n)
+{
+	n += buf->used;
```
stoeckmann, line 44: We could use `uaddsz_overflow` here to make sure to never overflow, no matter how huge function argument `n`.

stoeckmann (changes requested):
I like the approach. Ideally we can keep the API in a way that all return values have to be checked for error to never accidentally fool ourselves through reserving "enough" memory and eventually figuring out that our implicit code contains an error.

The cleanup attribute is awesome and I never expected to see a good `alloca` usage. If we want to keep `alloca` though, I'm rather indifferent.

No good idea so far if we could somehow make the API more robust against accidentally calling `strbuf_str` on a strbuf with stack memory and returning it from a function or storing it somewhere in a heap struct. But perhaps we have to live with it and actually have to think what we are doing here. :)

testsuite/test-strbuf.c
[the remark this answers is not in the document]
lucasdemarchi: Humn... I can add the documentation, but it seems different from this comment which is aiming to clarify that the call to strbuf_popchar() that needs to be done (to remove the last ' ') has the side effect of also guaranteeing the buffer has at least one char available, so the check `assert_return(result1 == buf.bytes, ...` can be done even if we called strbuf_str(), that could have changed it.

shared/strbuf.h
lucasdemarchi: I will add a commit on top with one of the alternatives I have in mind, then we decide and I can squash or leave it as is.

lucasdemarchi: The buffer variable will still be hidden from normal use.

shared/strbuf.c
lucasdemarchi: I think it's worthy. In several places we have the strategy "1) calculate the size; 2) allocate; 3) use it". Allowing to reserve space is IMO a good thing to do

testsuite/test-strbuf.c
```diff
@@ -115,7 +115,7 @@ static int test_strbuf_with_stack(const struct test *t)
 	strbuf_init_with_stack(&buf, 256);
 	strbuf_pushchars(&buf, test);
 	assert_return(streq(test, strbuf_str(&buf)), EXIT_FAILURE);
-	p =  strbuf_steal(&buf);
+	p = strbuf_steal(&buf);
```
lucasdemarchi, line 118: it looks like I squashed a fix in the wrong commit.

tools/depmod.c
lucasdemarchi: In fact, let me do it in an update so the compiler can see the 2 calls to strlen()

tools/depmod.c
lucasdemarchi: strbuf_reserve_extra() looks like a good name. I will change to have just that one

tools/depmod.c
lucasdemarchi: I was adding the '\0' on purpose to match the previous code, but indeed we can keep that out.

lucasdemarchi:
> No good idea so far if we could somehow make the API more robust against accidentally calling strbuf_str on a strbuf with stack memory and returning it from a function or storing it somewhere in a heap struct. But perhaps we have to live with it and actually have to think what we are doing here. :)

That's the same situation of declaring a buffer on stack and returning a pointer to it. I guess all the sanitizers we have can help us with that. If not, at least valgrind is here to the rescue :)

lucasdemarchi:
Please take a look in the last commit. That would be a replacement for the alloca. If we feel it's a better approach I can polish it more.

shared/strbuf.c
evelikov: ^^ :100:

shared/strbuf.h
```diff
@@ -17,6 +17,19 @@ struct strbuf {
...
 
+#define DECLARE_STRBUF_WITH_STACK(name__, sz__)   \
+	assert_cc(__builtin_constant_p(sz__));    \
+	char name__##_storage__[sz__];            \
+	_cleanup_strbuf_ struct strbuf name__ = { \
+		.bytes = name__##_storage__,      \
+		.size = sz__,                     \
+	}
+
+#define STRBUF_DECLARE(name__, sz__)            \
```
evelikov, line 28: Nits:
 - `s/STRBUF_DECLARE/DECLARE_STRBUF/g` to be consistent with the stack one
 - drop the unused `sz__` to make it obvious it doesn't pre-allocate (stack or heap) anything
 - can use use it across the series?

I greatly prefer these macros over `strbuf_init_with_stack()`.

shared/strbuf.h
stoeckmann: I like the alloca-free approach as well. Actually a bit more because it's without alloca. :)

tools/depmod.c
stoeckmann: Right, if we have needed data stored already, we can just reuse ist.

shared/strbuf.c
lucasdemarchi: done

testsuite/test-strbuf.c
lucasdemarchi: fixed

shared/strbuf.c
```diff
@@ -34,15 +34,17 @@ static bool buf_realloc(struct strbuf *buf, size_t sz)
 	return true;
 }
 
-static bool buf_grow(struct strbuf *buf, size_t newsize)
+bool strbuf_reserve_extra(struct strbuf *buf, size_t n)
```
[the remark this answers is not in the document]
lucasdemarchi: fixed

evelikov (approved):
One thing I forgot to mention, non-blocker:

When doing a release build the `asserts()` are omitted. Since we build `libshared.a` aka `strbuf.o` once, the xfail test likely will not trigger. One option is to move the function as inline to `strbuf.h`, then the explicit `-UNDEBUG` we add to the tests will keep the assert... Although we need to find a better solution for the long run.

Tl;Dr: xfail test will succeed on `meson setup --buildtype release` builds, move `strbuf_shrink_to()` to the header as WA.

Regardless of the above comment, lets land this PR - it's looking solid IMHO.

lucasdemarchi:
and it seems we need to review those asserts in strbuf.... so I'm leaving this as is for now.

lucasdemarchi:
Applied, thanks for the reviews

evelikov:
> we should have an assert_return(), not affected by NDEBUG

Counter intuitive of the name, the macro never `assert()` but does a simple `if/return`. Regardless, opened a task for the general `assert()` yak shaving.

PR #240, opened 2024-11-12
libkmod: Do not set errno in libkmod-index

Maaxxs:
There are two implementations for the index. The mmap index does not use errno at all. Which implementation is used, is based on the usage of kmod_load_resources and whether it succeeds or not. Therefore, callers using libkmod cannot rely on useful information in errno.

Part of the reason why we investigated this is the fact that an invalid magic value read by `index_file_open` leads to a NULL return value with errno set to `0`. This happens because `read_u32s` sets errno to `0` on success, while `index_file_open` initializes errno to `EINVAL` before performing any read operations.

No proof of concept because errno isn't checked 🙃

@stoeckmann

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -167,24 +167,16 @@ static int add_value(struct index_value **values, const char *value, size_t len,
 	return 0;
 }
 
-static int read_char(FILE *in)
+static inline int read_char(FILE *in)
 {
-	int ch;
-
-	errno = 0;
-	ch = getc_unlocked(in);
-	if (ch == EOF)
-		errno = EINVAL;
-	return ch;
+	return getc_unlocked(in);
 }
 
 static bool read_u32s(FILE *in, uint32_t *l, size_t n)
 {
 	size_t i;
 
-	errno = 0;
 	if (fread_unlocked(l, sizeof(uint32_t), n, in) != n) {
-		errno = EINVAL;
 		return false;
 	}
 	for (i = 0; i < n; i++)
@@ -312,7 +304,6 @@ struct index_file *index_file_open(const char *filename)
 	file = fopen(filename, "re");
 	if (!file)
 		return NULL;
-	errno = EINVAL;
 
 	if (!read_u32(file, &magic) || magic != INDEX_MAGIC)
 		goto err;
@@ -332,7 +323,6 @@ struct index_file *index_file_open(const char *filename)
 	new->tmp = NULL;
 	new->tmp_size = 0;
 
-	errno = 0;
 	return new;
 err:
 	fclose(file);


evelikov:
> callers using libkmod cannot rely on useful information in errno.

There are zero mentions of errno in the API documentation, so external users should not relying on it. On the other hand, there are multiple places within libkmod that do depend on it.

Have you checked all direct and indirect users of the changed functions don't check errno?

lucasdemarchi:
> Therefore, callers using libkmod cannot rely on useful information in errno.

errno is an interface with libc. You can't call a library method and then inspect errno - the things the library will call from libc will change according to configuration/improvements/kernel version etc. Our interface for error code is the return value from the functions.

As for the code change here, if it can be accepted, I will leave it to @stoeckmann to take a look.

stoeckmann:
> Have you checked all direct and indirect users of the changed functions don't check errno?

It took us some time, until we noticed a fact we described in the second paragraph. I guess @Maaxxs was right and we should have illustrated it with a code example. So ...

The main entry point for FILE based index processing is `index_file_open`. It is always surrounded by an if-statement like
```
if (ctx->indexes[index_number] != NULL) {
    ... memory-mapped index handling happens here ...
} else {
    ... FILE-based index handling happens here ...
}
```

This means that we can focus solely on the else-block, because outside of the if-statement, the code must handle both cases. And since we know that memory-mapped index handling doesn't set errno at all, everything outside of the if-statement can be ignored for validation.

Within these else-blocks, we haven't found any errno checks. They are all located within `libkmod.c`:
1. index_file_open -> index_searchwild -> index_file_close
2. index_file_open -> index_search -> index_file_close
3. index_file_open -> index_dump -> index_file_close

This implies that these else-blocks only call FILE-based index functions without doing anything fancy by calling other functions.

As far as I can be considered an independent reviewer of a PR which Max and I created together, I would say it's okay. :)
And hopefully, this explanation helps someone else to look at it again. It's fortunately not that much code outside of `libkmod-index.c` although one would expect otherwise at first.

libkmod/libkmod-index.c
```diff
@@ -167,24 +167,16 @@ static int add_value(struct index_value **values, const char *value, size_t len,
 	return 0;
 }
 
-static int read_char(FILE *in)
+static inline int read_char(FILE *in)
```
evelikov, line 170: There is no need for the `inline` here - the compiler will do it as applicable.

evelikov (changes requested):
The file vs mmap difference isn't particularly interesting nor is the users (if any) wrongly relying on errno. Let's change the commit message to something vaguely like:

`Don't set errno in FUNCTIONs, it's an libc interface which we should not be overriding. None of the functions up the call stack (some examples) check on it.`

Aside:
Looking around, we have other instances of similar `errno` misuse - `libkmod-builtin.c`, `libkmod-elf.c` `libkmod-file-zlib.c`, `libkmod-file.c`, `libkmod-module.c` - can we get some patches for those?

libkmod/libkmod-index.c
Maaxxs:
```
$ grep -E '^static .+ read_' libkmod/libkmod-index.c
static inline int read_char(FILE *in)
static bool read_u32s(FILE *in, uint32_t *l, size_t n)
static inline bool read_u32(FILE *in, uint32_t *l)
static inline uint32_t read_u32_mm(const void **p)
static inline uint8_t read_char_mm(const void **p)
static inline void read_value_mm(const void **p, struct index_mm_value *v)
```
adding inline there make all the `read_*` more consistent. Anyway, as you said, the compiler will do what's applicable.

stoeckmann:
> Aside: Looking around, we have other instances of similar `errno` misuse - `libkmod-builtin.c`, `libkmod-elf.c` `libkmod-file-zlib.c`, `libkmod-file.c`, `libkmod-module.c` - can we get some patches for those?

Looking forward to your contributions. At least it sounds like you investigated these files already. I mean, you did more than just `grep -l 'errno = ' libkmod/*`, right? :D

lucasdemarchi:
Not sure what y'all are talking about. There's nothing wrong on using errno inside the library and setting errno may be needed depending on what we are doing.  Any user calling a libkmod function and then checking errno is not properly checking for error - errno is a libc interface and libc is used by both libkmod and the caller.

PR #241, opened 2024-11-13
libkmod: Improve comment in libkmod-index.c

stoeckmann:
The 0xB007FA57 (INDEX_MAGIC_OLD) magic is very old, since it was already old in kmod v1. Let's not talk about "newer binary files" when refering to 0xB007F457 (INDEX_MAGIC) anymore. Restructured the sentence to make it easier to understand after comments were moved from `libkmod-index.h` to `libkmod-index.c` (see [here](https://github.com/kmod-project/kmod/commit/c4cbdf8e175acdd0d907ae9e250fd2fe7acf9a00#diff-0f42cd27fb1b02778916c8172b6bc772d9dc1c5df403b4828d932d149299d85bL28) for an example how the sentence was more fitting back then).

While at it, added another commit which fixes a typo in another comment.

CC: @Maaxxs

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -25,10 +25,9 @@
  * Integers are stored as 32 bit unsigned in "network" order, i.e. MSB first.
  * All files start with a magic number.
  *
- * Magic spells "BOOTFAST". Second one used on newer versioned binary files.
- * #define INDEX_MAGIC_OLD 0xB007FA57
+ * Magic spells "BOOTFAST". Deprecated versions encoded it as 0xB007FA57.
  *
- * We use a version string to keep track of changes to the binary format
+ * We use a version string to keep track of changes to the binary format.
  * This is stored in the form: INDEX_MAJOR (hi) INDEX_MINOR (lo) just in
  * case we ever decide to have minor changes that are not incompatible.
  */
--- a/shared/missing.h
+++ b/shared/missing.h
@@ -6,7 +6,7 @@
 /*
  * Macros pulled from linux/module.h, to avoid pulling the header.
  *
- * In practise very few people have it installed and distros do not install the
+ * In practice, very few people have it installed and distros do not install the
  * relevant package during their build.
  *
  * Values are UAPI, so they cannot change.


libkmod/libkmod-index.c
```diff
@@ -25,10 +25,9 @@
  * Integers are stored as 32 bit unsigned in "network" order, i.e. MSB first.
  * All files start with a magic number.
  *
- * Magic spells "BOOTFAST". Second one used on newer versioned binary files.
- * #define INDEX_MAGIC_OLD 0xB007FA57
+ * Magic spells "BOOTFAST". Deprecated versions encoded it as 0xB007FA57.
```
evelikov, line 28: Which versions are deprecated - where those produced with kmod v1, v2 or even earlier with module-init-tools?

stoeckmann: The referenced file had this old magic number even commented out in version 1. Definitely deprecated. Since I didn't start to investigate module-init-tools to trace its origin, it's save to say that whatever software version used it, is deprecated.

lucasdemarchi: /me reporting for my archaeology class
 
```
commit 44d7ac4f9f61db23a8d9feb74d31a0e97df7c6a5
Author: Jon Masters <jcm@jonmasters.org>
Date:   Fri Jan 23 01:05:48 2009 -0500

    add versioning to the binary module files
    
    We need to have a version attached to the binary trie in case we ever
    need to change the format in some fashion. So this patch does that, and
    also changes the MODULE_MAGIC to 0xB007F457 (change the 'A' to a '4')
    so that we can notice that we have the second version of the files.
    
    Signed-off-by: Jon Masters <jcm@jonmasters.org>
```

a few weeks later, the old version got commented out.

evelikov: Wait a minute ... TIL there is a git repo for module-init-tools :sweat_smile: 

Would be amazing if we something like the ^^ details as an inline comment, the cryptic one currently.

lucasdemarchi: I think we had a repo for modutils too (that got replaced by module-init-tools in the 2.4 -> 2.6 transition, but all I can find now are the tarballs and binary packages: https://www.kernel.org/pub/linux/utils/kernel/modutils/

lucasdemarchi:
Applied, thanks

PR #245, opened 2024-11-15
ci: add meson setup -D {distconf,module}dir=relative/ checks

evelikov:
Based on the autotools build, meson does not support relative directories for distconfig and moduledir. We fixed that recently, but never added a check so we don't regress.


lucasdemarchi:
Applied, thanks

PR #246, opened 2024-11-15
Various test prep work

evelikov:
Nothing too spectacular here, maybe apart from the last commit:

Some dedup (props to Lucas for the idea), use 77 for skipped tests, run _all_ tests, use distinct test names... remove need_spawn = false path

While working on some more tests/cleanups, I may or may-not have accidentally fork bombed my system, props to the two code-paths in our test suite. That combined with the buggy (seemingly since day 1) output let me to nuke one of them.

If anyone is strongly in favour of keeping it, I would request a patch for fix the output in return :-P

### DIFF
--- a/testsuite/README
+++ b/testsuite/README
@@ -39,27 +39,22 @@ pay attention when writing a test:
 
 4 - Fill in the config vector. Setting any of these configuration will make
     testsuite to export LD_PRELOAD with the necessary override libs before
-    executing the test. If you are not exec'ing an external binary, you need to
-    pass "need_spawn = true" below, otherwise it will not work (LD_PRELOAD is
-    only applied when exec'ing a binary). See each config description in
-    testsuite.h
+    executing the test. See each config description in testsuite.h
 
-5 - need_spawn: if testsuite will exec itself before running a test
-
-6 - expected_fail: if that test is expected to fail, i.e. the return code is
+5 - expected_fail: if that test is expected to fail, i.e. the return code is
     expected not to be 0.
 
-7 - The rootfs is populated by copying the entire contents of rootfs-pristine
+6 - The rootfs is populated by copying the entire contents of rootfs-pristine
     through setup-rootfs.sh then running setup-modules.sh to copy generated
     modules from module-playground. Update the latter script to include any
     modules your test needs.
 
-8 - Tests can be run individually, outside of 'meson test'. strace and gdb
+7 - Tests can be run individually, outside of 'meson test'. strace and gdb
     work too, as long as you tell them to operate on child process.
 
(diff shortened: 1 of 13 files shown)


testsuite/test-init.c
```diff
@@ -39,15 +39,15 @@ static noreturn int test_load_resources(const struct test *t)
...
+DEFINE_TEST_WITH_FUNC(test_load_resource1, test_load_resources,
 	    .description = "test if kmod_load_resources works (recent modprobe on kernel without modules.builtin.modinfo)",
 	    .config = {
 		[TC_ROOTFS] = TESTSUITE_ROOTFS "test-init-load-resources/",
 		[TC_UNAME_R] = "5.6.0",
 	    },
 	    .need_spawn = true);
 
-DEFINE_TEST(test_load_resources,
+DEFINE_TEST_WITH_FUNC(test_load_resource2, test_load_resources,
```
lucasdemarchi: I didn't even remember we had this variant of the macro. :-o

testsuite/test-init.c
```diff
@@ -144,13 +146,11 @@ static noreturn int test_remove(const struct test *t)
 
 	exit(EXIT_SUCCESS);
 }
-DEFINE_TEST(test_remove,
-	.description = "test if libkmod's remove_module returns ok",
+DEFINE_TEST(
+	test_remove, .description = "test if libkmod's remove_module returns ok",
```
lucasdemarchi, line 150: wth is clang-format doing here. I wise we had a _Pragma() or attribute we could give to this struct to convince clang-format to stop bothering. I wonder if it works if we add a `// clang-format off` before the define.

testsuite/README
```diff
@@ -39,27 +39,22 @@ pay attention when writing a test:
 
 4 - Fill in the config vector. Setting any of these configuration will make
     testsuite to export LD_PRELOAD with the necessary override libs before
-    executing the test. If you are not exec'ing an external binary, you need to
-    pass "need_spawn = true" below, otherwise it will not work (LD_PRELOAD is
-    only applied when exec'ing a binary). See each config description in
-    testsuite.h
```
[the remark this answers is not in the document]
lucasdemarchi: but indeed, the way it's coded is confusing and I'm ok with dropping it if it works

testsuite/test-init.c
evelikov: Indeed - not ultra happy with the reshuffle. Tinkering about - neither a `// clang-format off` on the definition helps, nor does a trailing comma.

AFAICT we are left with `//  clang-format off` individual instances, files and/or just carrying on. Let me know which you prefer.

lucasdemarchi:
Applied, thanks.

Issue #26, opened 2023-03-16
Segmentation fault when running depmod/kmod


martinclauss:
Hi :)

I first reported the bug here: https://bugs.archlinux.org/task/77868 but since it is an upstream issue, I want to report it here as well.

I built a custom Linux kernel with debugging information. If I'm correct the bug occurs when keeping the debug information in the modules. Excerpt from the PKGBUILD file (Arch Linux):

```
  echo "Installing modules..."
  make INSTALL_MOD_PATH="$pkgdir/usr" INSTALL_MOD_STRIP="--only-keep-debug" \
    DEPMOD=/doesnt/exist modules_install  # Suppress depmod
```

`INSTALL_MOD_STRIP="--only-keep-debug"` was `INSTALL_MOD_STRIP=1` before the modification.

While running `sudo pacman -U linux-dbg-6.2.6.arch1-1-x86_64.pkg.tar.zst ` and with

```
gdb -ex "set disassembly-flavor intel" -ex "run" -ex "bt full" -ex "info registers" -ex 'x/50i $pc' -ex 'x/50gx $sp' -args depmod -b "$BUILDROOT" "$KERNELVERSION"
```
(in `/usr/lib/initcpio/functions` and similar in `/usr/share/libalpm/scripts/depmod`) I get the following crash information:

```
Program received signal SIGSEGV, Segmentation fault.
0x000055555555943e in index_insert (node=0x555555fad950, 
    node@entry=0x555555fe4d60, 
    key=key@entry=0x7fffffffa9e0 "symbol:", <incomplete sequence \375>, 
    value=<optimized out>, priority=<optimized out>) at tools/depmod.c:290
290					prefix[j] = '\0';
#0  0x000055555555943e in index_insert (node=0x555555fad950, 
    node@entry=0x555555fe4d60, 
    key=key@entry=0x7fffffffa9e0 "symbol:", <incomplete sequence \375>, 
    value=<optimized out>, priority=<optimized out>) at tools/depmod.c:290
        prefix = 0x7ffff79f0000 <argp_default_options+96> "m4\233\367\377\177"
        n = <optimized out>
        j = 0
        child = <optimized out>
        i = <optimized out>
        ch = 109
#1  0x000055555555b9a4 in output_symbols_bin (depmod=0x7fffffffaf90, 
    out=0x55555558a4f0) at tools/depmod.c:2352
        duplicate = <optimized out>
        sym = 0x5555556fefc0
        len = 2
[...]

PR #247, opened 2024-11-15
depmod: Properly check index keys

stoeckmann:
Synchronize character checks with libkmod-index.c. 8-bit ASCII chars would turn negative (due to signed char), which index__checkstring does not properly check for.

Fixes segmentation fault observed in https://github.com/kmod-project/kmod/issues/26


evelikov (approved):
I made a patch to mass convert `int ch` to `unsigned char` a while ago, but never really sent it out :facepalm: 

Thanks for looking at the issue, the commit is spot on.

PR #248, opened 2024-11-15
shared: Use unsigned char in hash calculation

stoeckmann:
Spotted these while investigating https://github.com/kmod-project/kmod/issues/26.

- Use hash_superfast as intended (UBSAN gave a warning and yes, this modifies the hash result)
- Improve handling of out of memory conditions
- Add yet another fgets check (copy & paste code, last time I have just seen one of them)
- Use proper format specifiers

Since these don't fix the issue but would basically lead all to a single PR, I've grouped them in a single one instead.

### DIFF
--- a/shared/hash.c
+++ b/shared/hash.c
@@ -94,7 +94,7 @@ static inline unsigned int hash_superfast(const char *key, unsigned int len)
 	case 3:
 		hash += get_unaligned((uint16_t *)key);
 		hash ^= hash << 16;
-		hash ^= key[2] << 18;
+		hash ^= ((uint8_t)key[2]) << 18;
 		hash += hash >> 11;
 		break;
 


evelikov (approved):
Haven't looked at the `hash_superfast` commit and I suspect we _really_ need a test for it to land.

The rest of the patches look great.

lucasdemarchi:
> * Use hash_superfast as intended (UBSAN gave a warning and yes, this modifies the hash result)

not very happy with changing the result... if we are changing I think we need to add some instrumentation and compare the amound of collisions we have with the current hash table size. Also... if we are changing, I'd rather add a pre step in which we (possibly) process the first element to get an aligned buffer on u16, so we don't have to use get_unaligned().

However most of our strings are really small, so I'm not sure it will make a noticeable difference.

lucasdemarchi:
Applied all but the first. Please rebase and let's continue the discussion on how to land this fix.

lucasdemarchi:
Applied, thanks

PR #250, opened 2024-11-16
libkmod: Improve index dump performance

stoeckmann:
Use FILE for output to reduce the amount of system calls. Removes the only use case of write_str_safe, which can be removed as well.

This has another advantage, beside faster execution: Having a FILE in `kmod_dump_index` allows us easier error detection in the future (by using `ferror` here instead of multiplying it into FILE-based and memory-mapped index functions).

Talking about execution times, these are collected from modules of an Arch Linux installation:

### current master

Uses 198497 write calls.
Takes around 224 ms on this system.

```
$ strace -c modprobe -c > output.txt
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 99.98    0.224581           1    198497           write
  0.02    0.000045           7         6           munmap
  0.00    0.000000           0        13           read
  0.00    0.000000           0        17           close
  0.00    0.000000           0        18           fstat
  0.00    0.000000           0        30           mmap
  0.00    0.000000           0         7           mprotect
  0.00    0.000000           0         3           brk
  0.00    0.000000           0         2           pread64
  0.00    0.000000           0         1         1 access
  0.00    0.000000           0         1           execve
  0.00    0.000000           0         3           fcntl
  0.00    0.000000           0         1           arch_prctl
  0.00    0.000000           0         4           getdents64
  0.00    0.000000           0         1           set_tid_address
  0.00    0.000000           0        20         3 openat
  0.00    0.000000           0         9         5 newfstatat
  0.00    0.000000           0         1           set_robust_list
  0.00    0.000000           0         1           prlimit64
  0.00    0.000000           0         1           getrandom
  0.00    0.000000           0         1           rseq
------ ----------- ----------- --------- --------- ----------------
100.00    0.224626           1    198637         9 total
```

```
$ size tools/kmod libkmod/.libs/libkmod.so.2.5.0
 158343    5852     160  164355   28203 tools/kmod
  96173    2248       8   98429   1807d libkmod/.libs/libkmod.so.2.5.0
```

### new

Uses 573 write calls.
Takes around 7 ms.

Binary size shrinks by around 200 bytes, library size increases by around 60 bytes.
Speed increasement based on `strace`: around 30x
Speed increasement based on `time`: 4x to 10x (depending if you redirect to /dev/null or a real file)

```
$ strace -c modprobe -c > output.txt
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 84.74    0.005751          10       573           write
  2.95    0.000200           6        30           mmap
  2.08    0.000141           7        20         3 openat
  1.93    0.000131           6        19           close
  1.66    0.000113           5        20           fstat
  1.31    0.000089          14         6           munmap
  1.21    0.000082           6        13           read
  0.96    0.000065           9         7           mprotect
  0.80    0.000054           6         9         5 newfstatat
  0.52    0.000035           5         6           fcntl
  0.50    0.000034           8         4           getdents64
  0.24    0.000016           5         3           brk
  0.22    0.000015           7         2           pread64
  0.18    0.000012           6         2           dup
  0.13    0.000009           9         1           set_tid_address
  0.12    0.000008           8         1           set_robust_list
  0.10    0.000007           7         1           arch_prctl
  0.10    0.000007           7         1           prlimit64
  0.10    0.000007           7         1           getrandom
  0.09    0.000006           6         1           rseq
  0.07    0.000005           5         1           lseek
  0.00    0.000000           0         1         1 access
  0.00    0.000000           0         1           execve
------ ----------- ----------- --------- --------- ----------------
100.00    0.006787           9       723         9 total
```

```
$ size tools/kmod libkmod/.libs/libkmod.so.2.5.0
 158143    5852     160  164155   2813b tools/kmod
  96208    2272       8   98488   180b8 libkmod/.libs/libkmod.so.2.5.0
```

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -360,7 +360,7 @@ static struct index_node_f *index_readchild(const struct index_node_f *parent, i
 	return NULL;
 }
 
-static void index_dump_node(struct index_node_f *node, struct strbuf *buf, int fd)
+static void index_dump_node(struct index_node_f *node, struct strbuf *buf, FILE *fp)
 {
 	struct index_value *v;
 	size_t pushed;
@@ -369,10 +369,10 @@ static void index_dump_node(struct index_node_f *node, struct strbuf *buf, int f
 	pushed = strbuf_pushchars(buf, node->prefix);
 
 	for (v = node->values; v != NULL; v = v->next) {
-		write_str_safe(fd, buf->bytes, buf->used);
-		write_str_safe(fd, " ", 1);
-		write_str_safe(fd, v->value, strlen(v->value));
-		write_str_safe(fd, "\n", 1);
+		fwrite(buf->bytes, 1, buf->used, fp);
+		fputc(' ', fp);
+		fputs(v->value, fp);
+		fputc('\n', fp);
 	}
 
 	for (ch = node->first; ch <= node->last; ch++) {
@@ -382,7 +382,7 @@ static void index_dump_node(struct index_node_f *node, struct strbuf *buf, int f
 			continue;
 
 		if (strbuf_pushchar(buf, ch)) {
-			index_dump_node(child, buf, fd);
+			index_dump_node(child, buf, fp);
 			strbuf_popchar(buf);
 		}
 	}
@@ -391,7 +391,7 @@ static void index_dump_node(struct index_node_f *node, struct strbuf *buf, int f
 	index_close(node);
 }
 
-void index_dump(struct index_file *in, int fd, bool alias_prefix)
+void index_dump(struct index_file *in, FILE *fp, bool alias_prefix)
 {
 	struct index_node_f *root;
 	struct strbuf buf;
@@ -402,7 +402,7 @@ void index_dump(struct index_file *in, int fd, bool alias_prefix)
 
 	strbuf_init(&buf);
 	if (!alias_prefix || strbuf_pushchars(&buf, "alias "))
-		index_dump_node(root, &buf, fd);
+		index_dump_node(root, &buf, fp);
 	strbuf_release(&buf);
 }
 
@@ -821,7 +821,7 @@ static struct index_mm_node *index_mm_readchild(const struct index_mm_node *pare
 	return NULL;
 }
 
-static void index_mm_dump_node(struct index_mm_node *node, struct strbuf *buf, int fd)
+static void index_mm_dump_node(struct index_mm_node *node, struct strbuf *buf, FILE *fp)
 {
 	const void *p;
 	size_t i, pushed;
@@ -833,10 +833,10 @@ static void index_mm_dump_node(struct index_mm_node *node, struct strbuf *buf, i
 		struct index_mm_value v;
 
 		read_value_mm(&p, &v);
-		write_str_safe(fd, buf->bytes, buf->used);
-		write_str_safe(fd, " ", 1);
-		write_str_safe(fd, v.value, v.len);
-		write_str_safe(fd, "\n", 1);
+		fwrite(buf->bytes, 1, buf->used, fp);
+		fputc(' ', fp);
+		fwrite(v.value, 1, v.len, fp);
+		fputc('\n', fp);
 	}
 
 	for (ch = node->first; ch <= node->last; ch++) {
@@ -847,7 +847,7 @@ static void index_mm_dump_node(struct index_mm_node *node, struct strbuf *buf, i
 			continue;
 
 		if (strbuf_pushchar(buf, ch)) {
-			index_mm_dump_node(child, buf, fd);
+			index_mm_dump_node(child, buf, fp);
 			strbuf_popchar(buf);
 		}
 	}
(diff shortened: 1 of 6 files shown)


shared/util.c
```diff
@@ -237,32 +237,6 @@ ssize_t read_str_safe(int fd, char *buf, size_t buflen)
...
 
-ssize_t write_str_safe(int fd, const char *buf, size_t buflen)
-{
-	size_t todo = buflen;
-	size_t done = 0;
-
-	assert_cc(EAGAIN == EWOULDBLOCK);
-
-	do {
-		ssize_t r = write(fd, buf + done, todo);
```
evelikov, line 248: Why did we remove the helper instead of adapting it to use `fwrite()`? AFAICT all the safety handling it does is still applicable.

libkmod/libkmod.c
```diff
@@ -810,22 +810,37 @@ KMOD_EXPORT void kmod_unload_resources(struct kmod_ctx *ctx)
 
 KMOD_EXPORT int kmod_dump_index(struct kmod_ctx *ctx, enum kmod_index type, int fd)
 {
+	int err = 0, fd2;
+	FILE *fp;
+
 	if (ctx == NULL)
 		return -ENOSYS;
 
+	fd2 = dup(fd);
```
evelikov, line 819: AFAICT libraries should really be using CLOEXEC, aka `fcntl(fd, F_DUPFD_CLOEXEC...)` or alike.

libkmod/libkmod.c
```diff
@@ -810,22 +810,37 @@ KMOD_EXPORT void kmod_unload_resources(struct kmod_ctx *ctx)
...
+	if (fd2 == -1)
+		return -errno;
+	fp = fdopen(fd2, "ae");
+	if (fp == NULL) {
+		err = -errno;
+		close(fd2);
+		return err;
```
evelikov, line 826: Let's move all the whole dup/fdopen hunk after _all_ the input validation - aka after the `type` check below.

libkmod/libkmod-index.c
```diff
@@ -833,10 +833,10 @@ static void index_mm_dump_node(struct index_mm_node *node, struct strbuf *buf, i
...
 
 		read_value_mm(&p, &v);
-		write_str_safe(fd, buf->bytes, buf->used);
-		write_str_safe(fd, " ", 1);
-		write_str_safe(fd, v.value, v.len);
-		write_str_safe(fd, "\n", 1);
+		fwrite(buf->bytes, 1, buf->used, fp);
+		fputc(' ', fp);
+		fwrite(v.value, 1, v.len, fp);
+		fputc('\n', fp);
```
evelikov, line 839: Out of curiosity: have you tried placing the contents into a buffer (on the stack) and doing a single write for given value instead of 4? Does it make a difference?

evelikov (changes requested):
The numbers look amazing - nicely spotted. The benefit might shrink a bit when(if) we reintroduce `fwrite_str_safe()`, but as a whole it should still be a win.

shared/util.c
stoeckmann: The problem with FILE output is that `fclose` itself could still write data. And this is a one-shot approach. We cannot call it multiple times until the remaining data is successfully written.

I mistakenly believed that FILE itself handles short writes, but I was wrong. So if we want to keep interrupt-safe writes, we cannot use FILE here. At best, we would adjust the API, but ... Not in a simple PR. :)

libkmod/libkmod.c
stoeckmann: The `fdopen` would set ´O_CLOEXEC`. Unfortunately I cannot simply use `dup3`, which would do it atomically for us, because then I would have to figure out a free fd first (or open one just to reserve it).

libkmod/libkmod-index.c
stoeckmann: This already helps by almost 50 %, which is nice. At first I tried scratchbuf (and soon strbuf), but we could also keep a buffer around for longer. See https://github.com/kmod-project/kmod/pull/252 as an example.

libkmod/libkmod.c
stoeckmann: Would adjust it if we decide if it's even worth it:

- FILE loses EINTR handling (modprobe -c shouldn't really care, since it first uses FILE and then passes a file descriptor to library, but let's not break API and whatever other users might rely on)
- wbuf speeds up processing, but introduces custom code, not improving the error handling

So ... let's see. I wouldn't mind keeping everything as it is right now and, if we ever adjust the API in a release, reconsider turning it FILE-based.

lucasdemarchi: I second that. What we could have is one additional API that receives a FILE* as argument.... we are not really going to change the API and break users because of this.

However if we add the new API, then we don't drop the old code since converting it to use FILE* would mean to lose EINTR handling. So.... I'm leaning towards the wbuf approach.

stoeckmann:
Will close this one since we prefer https://github.com/kmod-project/kmod/pull/252.

PR #252, opened 2024-11-17
libkmod: Improve index dump performance

stoeckmann:
Buffer data before calling write for output. This reduces the amount of total write system calls and leads to faster execution times.

This is a simple approach to imitate FILE's I/O buffering, which we cannot use because we want to properly handle interrupts.

Alternative solution to https://github.com/kmod-project/kmod/pull/250

### DIFF
--- a/libkmod/libkmod-index.c
+++ b/libkmod/libkmod-index.c
@@ -3,6 +3,8 @@
  * Copyright (C) 2011-2013  ProFUSION embedded systems
  */
 
+#include <sys/param.h>
+
 #include <arpa/inet.h>
 #include <assert.h>
 #include <errno.h>
@@ -133,6 +135,44 @@ enum node_offset {
 
 	INDEX_NODE_MASK = 0x0FFFFFFF, /* Offset value */
 };
+
+struct wrtbuf {
+	char bytes[4096];
+	size_t len;
+	int fd;
+};
+
+static void wrtbuf_init(struct wrtbuf *buf, int fd)
+{
+	buf->len = 0;
+	buf->fd = fd;
+}
+
+static void wrtbuf_flush(struct wrtbuf *buf)
+{
+	if (buf->len > 0) {
+		write_str_safe(buf->fd, buf->bytes, buf->len);
+		buf->len = 0;
+	}
+}
+
+static void wrtbuf_write(struct wrtbuf *buf, const char *p, size_t len)
+{
+	size_t todo = len;
+	size_t done = 0;
+
+	do {
+		size_t n = MIN(sizeof(buf->bytes) - buf->len, todo);
+
+		memcpy(buf->bytes + buf->len, p + done, n);
+		buf->len += n;
+		todo -= n;
+		done += n;
+
+		if (buf->len == sizeof(buf->bytes))
+			wrtbuf_flush(buf);
+	} while (todo > 0);
+}
 
 void index_values_free(struct index_value *values)
 {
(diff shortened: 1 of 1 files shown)


libkmod/libkmod-index.c
```diff
@@ -135,6 +135,40 @@ enum node_offset {
 	INDEX_NODE_MASK = 0x0FFFFFFF, /* Offset value */
 };
 
+/* Write buffer to imitate FILE I/O optimization */
+struct wbuf {
+	char bytes[4096];
```
evelikov, line 140: Nits:
 - let's drop the comment and use more indicative name (say "write_buffer")
 - could probably bump this to 16K or more based on the stats (aka perf and/or how large our buffers are)

libkmod/libkmod-index.c
```diff
@@ -135,6 +135,40 @@ enum node_offset {
...
+	}
+}
+
+static void wbuf_write(struct wbuf *buf, const char *p, size_t len)
+{
+	if (buf->len + len > sizeof(buf->bytes)) {
+		wbuf_flush(buf);
+	}
+	if (len > sizeof(buf->bytes)) {
+		write_str_safe(buf->fd, p, len);
```
evelikov, line 165: Should probably check and adjust the buffer size so this doesn't trigger. Thus fewer code paths, greater coverage, etc.

evelikov (commented):
Added a couple of notes, but ideally we'll hear from Lucas which route he prefers before we go too deep.

lucasdemarchi (commented):
I think I prefer this approach. I actually think that maybe we could add a kmod_dump_index() that receives FILE* as argument. Then the struct wbuf, instead of mimic'ing stdio, simply chooses one or the other.

I doubt there are many users out there of kmod_dump_index()... so if we migrate modprobe to the new API, we likely migrate 100% or 99% of the users.

From a quick look in github, the only other user I found is https://github.com/falconindy/modtree, that doesn't seem maintained anymore (side note that having a modtree output in kmod would be nice)

libkmod/libkmod-index.c
evelikov: Can you elaborate on the stack concern? Should we be using `-Wframe-larger-than=XXX` (as a separate commit/PR)?

stoeckmann: There is no need to exceed 4 KB (a page size). Using a page size is already very good, after all it's what FILE implementations chose to do as well. Copying 2 or more pages won't gain much, especially in our scenario with only a few hundred write calls left.

I would have more of a concern due to musl-based systems which allocate only 128 KB stacks for threads. Since we are in library code here, let's not exceed what we already do by much more due to PATH_MAX (4096) etc.

stoeckmann: > Can you elaborate on the stack concern? Should we be using `-Wframe-larger-than=XXX` (as a separate commit/PR)?

That might make sense imho

libkmod/libkmod-index.c
stoeckmann: I'll adjust the code to always align with 4 KB writes, so the kernel wouldn't see a difference between a FILE implementation and ours. This check is required because index files could contain huge strings. In reality they won't, so this if-block is never entered. But no matter how large we make our buffer, we can always parse a string which is longer than our size on the stack. 

If you want to be funny with index files, nothing stops you from at least UINT_MAX chars in a string. The offsets would limit us, but there is no limit after a node starts ... :)

libkmod/libkmod-index.c
stoeckmann: Since we have `struct strbuf sbuf`, I switched from `wbuf` to `struct wrtbuf wbuf`. Looks nicer in functions where `sbuf` and `wbuf` are used next to each other. ;)

libkmod/libkmod-index.c
evelikov: > This check is required because index files could contain huge strings. In reality they won't, so this if-block is never entered

Sounds like we can move the check in the one(?) caller where it could trigger (for malicious input really). Might even make the function argument `const char buf[static 4096]` or alike, so compiler/static tools don't miss-report "omg this can overflow" scenarios.

lucasdemarchi:
Applied, thanks.

PR #254, opened 2024-11-19
More cmdline test

lucasdemarchi:
A question was raised today and I didn't know the answer from top of my head or looking at the documentation. Make sure this doesn't change, even if unlikely, as users rely on this behavior. While at it, cleanup.


testsuite/test-modprobe.c
```diff
@@ -244,9 +244,7 @@ DEFINE_TEST(modprobe_param_kcmdline_show_deps,
 	},
 	.output = {
 		.out = TESTSUITE_ROOTFS "test-modprobe/module-param-kcmdline/correct.txt",
-	},
-	.modules_loaded = "",
-	);
+	});
```
evelikov, line 247: As-is this test also ensures that we are not unintentionally loading the module, as part of `--show-depends`. I suspect it's a case we want to check for plus equivalent ones for the other `--show*` options.

The core purpose of test is to check for something else, so :+1: from me.

Let me know if you see value in the test cases mentioned above and we'll add them to the todo list.

evelikov (approved):
More tests - yes please.

testsuite/test-modprobe.c
```diff
@@ -47,6 +47,14 @@ DEFINE_TEST(modprobe_show_depends2,
 	.output = {
 		.out = TESTSUITE_ROOTFS "test-modprobe/show-depends/correct-mod-simple.txt",
 	});
+DEFINE_TEST_WITH_FUNC(modprobe_show_depends_no_load, modprobe_show_depends2,
```
evelikov, line 50: I wonder what the extra files were meant for. Glancing around:

It seems like modprobe has been using `dry-run` (aka don't load the module) and `ignore-loaded` (aka don't look at /proc or /sys) since day one... Where the porting commit eeb627004b6ee3610913f98ef143e84eb0546360 added these seemingly unused files.

AFAICT the extra test looks good so :+1: from me.

evelikov (approved):

lucasdemarchi:
Applied, thanks.

PR #256, opened 2024-11-21
libkmod: expand the deprecated/0xB007FA57 comment

evelikov:
Provide a summary how the magic and version fields have evolved. It serves as better reference than the current "deprecated versions" note.


lucasdemarchi:
Applied, thanks

PR #257, opened 2024-11-21
Bug fixes from incremental depmod PR

mwilck:
This PR splits off the bug fixes from #253, as requested by @evelikov.

I found these issues while making the kmod CI tests work with the "incremental depmod" patch set.

### DIFF
--- a/shared/hash.c
+++ b/shared/hash.c
@@ -258,7 +258,7 @@ int hash_del(struct hash *hash, const char *key)
 		hash->free_value((void *)entry->value);
 
 	entry_end = bucket->entries + bucket->used;
-	memmove(entry, entry + 1, (entry_end - entry) * sizeof(struct hash_entry));
+	memmove(entry, entry + 1, (entry_end - entry - 1) * sizeof(struct hash_entry));
 
 	bucket->used--;
 	hash->count--;
--- a/shared/hash.c
+++ b/shared/hash.c
@@ -249,6 +249,9 @@ int hash_del(struct hash *hash, const char *key)
 		.value = NULL,
 	};
 
+	if (bucket->entries == NULL)
+		return -ENOENT;
+
 	entry = bsearch(&se, bucket->entries, bucket->used, sizeof(struct hash_entry),
 			hash_entry_cmp);
 	if (entry == NULL)
--- a/shared/hash.c
+++ b/shared/hash.c
@@ -41,7 +41,7 @@ struct hash *hash_new(unsigned int n_buckets, void (*free_value)(void *value))
 	hash->n_buckets = n_buckets;
 	hash->free_value = free_value;
 	hash->step = n_buckets / 32;
-	if (hash->step == 0)
+	if (hash->step < 4)
 		hash->step = 4;
 	else if (hash->step > 64)
 		hash->step = 64;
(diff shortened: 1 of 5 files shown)


shared/hash.c
```diff
@@ -258,7 +258,7 @@ int hash_del(struct hash *hash, const char *key)
 		hash->free_value((void *)entry->value);
 
 	entry_end = bucket->entries + bucket->used;
-	memmove(entry, entry + 1, (entry_end - entry) * sizeof(struct hash_entry));
+	memmove(entry, entry + 1, (entry_end - entry - 1) * sizeof(struct hash_entry));
```
evelikov, line 263: Can we trigger this with the existing tests, alternatively let's add a test in our testsuite

shared/hash.c
```diff
@@ -249,6 +249,8 @@ int hash_del(struct hash *hash, const char *key)
 		.value = NULL,
 	};
 
+	if (bucket->entries == NULL)
+		return -ENOENT;
```
evelikov, line 253: Nit: please add a blank line.

Similar to the previous commit, can we get a test. Massive thanks in advance.

shared/strbuf.c
```diff
@@ -102,6 +102,8 @@ size_t strbuf_pushmem(struct strbuf *buf, const char *src, size_t sz)
 
 	if (!strbuf_reserve_extra(buf, sz))
 		return 0;
+	if (sz == 0)
+		return 0;
```
evelikov, line 106: We can have this check before the reserve call.

evelikov (approved):
The changes look great IMHO. Would be amazing to have a test or two but it's a non-blocking request.

mwilck:
I've added tests now that trigger the failures that I saw on #253.

I'm not sure how you deal with failures like this. I chose to add the test with `.expected_fail=true` first, and put the fix on top, reverting the `expected_fail` property.

I also added one more patch ("always use a step size of 4"), as the logic for determining the step size seems strange (I wouldn't have encountered the out-of-bounds memory access problem if the step size wasn't 1 with `n_buckets == 32`. Please have a look.

evelikov (approved):
You are a star, thank you for the tests - loving the `expected_fail` approach. The extra patch also looks good.

Seems like I cannot approve for CI, so it'll be up-to @lucasdemarchi

mwilck:
rebased onto current master

lucasdemarchi:
@evelikov  no idea why CI is not running on this PR.

lucasdemarchi:
Thanks a lot for the fixes and additional tests. I was going to swap the order of commits and avoid using the xfail... but changed my mind - it looks good in this order. I fixed the coding style issue and applied.

evelikov:
@lucasdemarchi some point last year, GitHub changed so that first PR from new people need to be approved by maintainers - [their blog post has the details.](https://github.blog/changelog/2023-06-06-security-enhancements-to-required-approvals-on-pull-requests/)

I suspect that adding me [as reviewer](https://github.com/orgs/community/discussions/23765#discussioncomment-8067102) might also do the trick.

lucasdemarchi:
@evelikov the problem is that the PR itself was not showing it the approval as needed. I had to go to each individual worfklow and approve them individually.  I saw the request when the PR was opened, but it wasn't available anymore after it had been reviewed.

lucasdemarchi:
> I suspect that adding me [as reviewer](https://github.com/orgs/community/discussions/23765#discussioncomment-8067102) might also do the trick.

it doesn't seem that there is such a role in Github. There is "triage" and "write". Anyway I created a "Reviewer" team with write access and invited both you and @stoeckmann. Note that PRs shouldn't be merged though - I manually apply the PRs pushing to both kernel.org and github.

Issue #50, opened 2024-07-17
testsuite: Allow to run tests under a vm


lucasdemarchi:
Instead o mocking the kernel side, allow to run the tests with a real kernel under a vm. This means removing the LD_PRELOAD where needed and integrate the VM setup to be executed under qemu.

gongjun-song:
Could you please clarify our goal? Which of the following meets your expectations?

1. Build QEMU with Meson to facilitate automatic testing of a real kernel in the testsuite.
2. Use QEMU to manually test the testsuite in the vm with a real kernel.

lucasdemarchi:
@gongjun-song it's more around (2). Currently the testsuite mimics the kernel interface by redirecting the calls to init_module/finit_module/delete_module and any path-related call to somewhere else. See `testsuite/{delete_module.c,init_module.c,path.c,uname.c}`. Goal is to be able to disable this part and test it in a real kernel.

gongjun-song:
Kmod Team 2 will work on this issue for the Hackathon, thanks!

PR #258, opened 2024-11-21
RFC: Consolidate test rootfs in single place

evelikov:
This series started off as de-duplicating the macros that we copy-paste across the tests and ultimately paved the way to allowing the tests to run in a VM aka https://github.com/kmod-project/kmod/issues/50.

That goal is not yet complete, but the general premise is that:
 - `DEFINE_TEST` provides the rootfs (_just_ the test sub-folder) and uname
 - any files (output.out, output.err, output.files) are stored and defined as relative to ^^
   - some tests are currently bugged wrt ^^
 - path construction - be that for `output` or the `TESTSUITE_ROOTFS` happens in the shared places - `testsuite.c` and/or `path.c`
 - (to be added as follow-up PR) machinery will copy/mount the sub-test folder into a chroot-like setup and run the test

As you can see the patch series isn't even complete yet. Although since it's churn _heavy_ I thought that sending it out early to feedback, would be a good idea.

Aside: I have another 1-2 series to further reduce the boilerplate, following which the tests will start to flow :crossed_fingers:

### DIFF
--- a/testsuite/test-blacklist.c
+++ b/testsuite/test-blacklist.c
@@ -83,7 +83,7 @@ fail_lookup:
 
 DEFINE_TEST(blacklist_1, .description = "check if modules are correctly blacklisted",
 	    .config = {
-		    [TC_ROOTFS] = TESTSUITE_ROOTFS "test-blacklist/",
+		    [TC_ROOTFS] = "test-blacklist/",
 	    });
 
 TESTSUITE_MAIN();
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -59,7 +59,7 @@ static const char *test_rootfs(const struct test *t)
 	if (t->config[TC_ROOTFS] == NULL)
 		return NULL;
 
-	snprintf(dirname, sizeof(dirname), "%s", t->config[TC_ROOTFS]);
+	snprintf(dirname, sizeof(dirname), TESTSUITE_ROOTFS "%s", t->config[TC_ROOTFS]);
 	return dirname;
 }
 
@@ -69,7 +69,7 @@ static const char *test_actual_filename(const char *fn)
 	if (fn == NULL)
 		return NULL;
 
-	snprintf(filename, sizeof(filename), "%s", fn);
+	snprintf(filename, sizeof(filename), TESTSUITE_ROOTFS "%s", fn);
 	return filename;
 }
 
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -73,6 +73,16 @@ static const char *test_actual_filename(const char *fn)
 	return filename;
 }
 
+static const char *test_stdout(const struct test *t)
+{
+	return test_actual_filename(t->output.out);
+}
+
+static const char *test_stderr(const struct test *t)
+{
+	return test_actual_filename(t->output.err);
+}
+
 static void help(void)
 {
 	const struct option *itr;
@@ -245,7 +255,7 @@ static inline int test_run_child(const struct test *t, int fdout[2], int fderr[2
 	test_export_environ(t);
 
 	/* Close read-fds and redirect std{out,err} to the write-fds */
-	if (t->output.out != NULL) {
+	if (test_stdout(t) != NULL) {
 		close(fdout[0]);
 		if (dup2(fdout[1], STDOUT_FILENO) < 0) {
 			ERR("could not redirect stdout to pipe: %m\n");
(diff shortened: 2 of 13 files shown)


testsuite/testsuite.c
```diff
@@ -53,6 +53,16 @@ static const struct {
 					OVERRIDE_LIBDIR "delete_module.so" },
 };
 
+static const char *test_rootfs(const struct test *t)
+{
+	static char dirname[PATH_MAX];
```
lucasdemarchi, line 58: why is this a static var?

testsuite/testsuite.c
```diff
@@ -63,6 +63,16 @@ static const char *test_rootfs(const struct test *t)
 	return dirname;
 }
 
+static const char *test_actual_filename(const char *fn)
+{
+	static char filename[PATH_MAX];
```
lucasdemarchi, line 68: ditto, I don't really like these static variables.

testsuite/testsuite.c
```diff
@@ -59,7 +59,7 @@ static const char *test_rootfs(const struct test *t)
 	if (t->config[TC_ROOTFS] == NULL)
 		return NULL;
 
-	snprintf(dirname, sizeof(dirname), "%s", t->config[TC_ROOTFS]);
+	snprintf(dirname, sizeof(dirname), TESTSUITE_ROOTFS "%s", t->config[TC_ROOTFS]);
```
lucasdemarchi, line 62: oh... these functions make absolute no sense before this commit. And still it's very odd to save things in this static variable. Can we just open TESTSUITE_ROOTFS in the beginning and switch everything to use openat()?

lucasdemarchi (changes requested):

testsuite/test-blacklist.c
```diff
@@ -83,7 +83,7 @@ static int blacklist_1(const struct test *t)
 
 DEFINE_TEST(blacklist_1, .description = "check if modules are correctly blacklisted",
 	    .config = {
-		    [TC_ROOTFS] = TESTSUITE_ROOTFS "test-blacklist/",
```
lucasdemarchi, line 86: Idea overall looks nice. I like the dedup here. However as I said I´d just do open(TESTSUITE_ROOTFS) and then use openat() to make everything relative to that.

testsuite/testsuite.c
evelikov: The idea is the fall outs of it being static are not (afaict) applicable here and having it dynamically allocated will make `test_export_environ()` even hairier.

Don't have a strong opinion though.

testsuite/testsuite.c
evelikov: It's a good idea which crossed my mind.

Plumbing the `fd` across the dozen+ functions will be quite hairy. Alternatively is to use global variable ... of which we don't have any ATM.

Will go with the global var, unless there are other suggestions.

evelikov:
Pondering a bit, I'm inclined to have v2 roughly as follows. Let me know if anything seems amiss.

```
DEFINE_TEST(foobar,
	.description = "something",
	.config = {
		[TC_UNAME_R] = MODULES_UNAME,
		[TC_ROOTFS] = ROOBAR_ROOTFS,
	},
	.output = {
		.out = ROOBAR_ROOTFS "/correct.txt",
		.files = (const struct keyval[]) {
			{ ROOBAR_ROOTFS "/correct-modules.alias",
			  ROOBAR_ROOTFS "/modules.alias" },
			{ },
		},
	});
```

```
DEFINE_TEST(foobar,
	.description = "something",
	.config = {
		[TC_UNAME_R] = MODULES_UNAME,
		[TC_ROOTFS] = ROOBAR_ROOTFS,
	},
	.compare = {
                .stdout = true,
		.modules-alias = true, // or "alias" or another name really
	});
```

The comparison (aka correct) files will live in "$TC_ROOTFS/compare/" (stdout.txt stderr.txt, modules.alias etc). The test infra will preemptively fail if a) the bool is set yet file is missing or b) when file is present but bool is false.

lucasdemarchi:
@evelikov  that new format looks good. Maybe also keep a `files[]` in that compare struct for the one-off cases?

```
.compare = {
        .files = { "/foo/bar/modules.alias", "/foor/bar/modules.dep" },
}
```

... which expects the same file to be present in $TC_ROOTFS/compare/

lucasdemarchi:
I picked the first commit, "testsuite/test-weakdep: remove custom handling".

lucasdemarchi:
@evelikov are you going to have time to continue on this endeavor? It was looking to be on a right direction.

PR #260, opened 2024-11-23
Fix potential security vulnerability 

GraysonNocera:
Hello, I was running some CodeQL queries on this project when I came across a potential unbounded write that could cause buffer overflow. I have implemented the fix in this PR.

Screenshot 2024-11-23 at 3 09 23 PM

### DIFF
--- a/tools/static-nodes.c
+++ b/tools/static-nodes.c
@@ -144,7 +144,7 @@ static void help(void)
 static int do_static_nodes(int argc, char *argv[])
 {
 	struct utsname kernel;
-	char modules[PATH_MAX], buf[4096];
+	char modules[PATH_MAX], buf[PATH_MAX];
 	const char *output = "/dev/stdout";
 	FILE *in = NULL, *out = NULL;
 	const struct static_nodes_format *format = &static_nodes_format_human;


evelikov:
Please squash the commit and add appropriate commit message. If possible avoid adding pictures in PRs but instead describe in your own words the issue at hand and how the commit addresses it.

Personally I am not worried about any security implications, since to trigger a specially crafted input file which is available in root-only location.

tools/static-nodes.c
```diff
@@ -252,8 +252,10 @@ static int do_static_nodes(int argc, char *argv[])
 		if (buf[0] == '#')
 			continue;
 
+		char format[22];
+		sprintf(format, "%%%ds %%%ds %%c%%u:%%u", PATH_MAX - 1, PATH_MAX - 1);
 		matches =
-			sscanf(buf, "%s %s %c%u:%u", modname, devname, &type, &maj, &min);
+			sscanf(buf, format, modname, devname, &type, &maj, &min);
```
stoeckmann, line 258: The usage of `sscanf(... "%s")` is not a general security issue. The input string must be long enough to overflow the target buffer. This means that `buf` must be larger than `modname` or `devname`.

Realistically, all Linux systems these days have a `PATH_MAX` of 4096. To make this code robust, it would be easiest to declare `buf` with `PATH_MAX` instead of `4096`. That would be and adjustment of exactly one line of code and no binary change.

Proof of Concept about sscanf safety:
```
#include <stdio.h>
#include <string.h>

int
main(int argc, char *argv[])
{
        char buf[4096], buf2[4096];
        memset(buf, 'A', sizeof(buf));
        buf[sizeof(buf) - 1] = '\0';
        sscanf(buf, "%s", &buf2);
        return 0;
}
```

So I would prefer if we apply this patch instead:

```diff
diff --git a/tools/static-nodes.c b/tools/static-nodes.c
index cb8afd3..d7557f4 100644
--- a/tools/static-nodes.c
+++ b/tools/static-nodes.c
@@ -144,7 +144,7 @@ static void help(void)
 static int do_static_nodes(int argc, char *argv[])
 {
 	struct utsname kernel;
-	char modules[PATH_MAX], buf[4096];
+	char modules[PATH_MAX], buf[PATH_MAX];
 	const char *output = "/dev/stdout";
 	FILE *in = NULL, *out = NULL;
 	const struct static_nodes_format *format = &static_nodes_format_human;
```

tools/static-nodes.c
```diff
@@ -252,8 +252,10 @@ static int do_static_nodes(int argc, char *argv[])
 		if (buf[0] == '#')
 			continue;
 
+		char format[22];
+		sprintf(format, "%%%ds %%%ds %%c%%u:%%u", PATH_MAX - 1, PATH_MAX - 1);
```
stoeckmann, line 256: If this commit is about protecting us against a `PATH_MAX` which is not 4096, then this `sprintf` call is actually dangerous. You could overflow the `format` buffer if PATH_MAX is sufficiently large.

Yet it's also just theoretical. A system with such a huge `PATH_MAX` would probably not compile the code at all because the compiler can already figure out that the stack wouldn't be large enough for all these variables which are declared like `char buf[PATH_MAX]` etc.

So let's go with the simple diff I've recommended. Avoids introduction of yet another `sprintf` call.

stoeckmann (changes requested):
Adding `sprintf` and a variable instead of fixed format string rises more red flags than an `sscanf` for all these static code scanners. ;)

stoeckmann (approved):

lucasdemarchi:
@GraysonNocera  a few things:

- It's not appropriate to add a Signed-off-by from someone else. The first Signed-off-by should match the author of the patch.  So I think what you want is:

```
Signed-off-by: Grayson Nocera <gnocera@purdue.edu>
Suggested-by: Tobias Stoeckmann <tobias@stoeckmann.org>
```

- "This fix adresses the issue described in the pull request." is unneeded redirection. :  the commit message should describe the issue rather than point to the pull request. It looks like the rest of the commit message already describes it so you can drop the first part.

lucasdemarchi:
Applied, thanks.

PR #261, opened 2024-11-28
libkmod: tools: Introduce 'module alternative' directory (was: module fallback)

esposem:
This PR is the improvement of https://github.com/kmod-project/kmod/pull/202 opened by @vittyvk .

Below you find a description of the original PR. The main change here is that if `/lib/modules` exists, nothing changes in kmod functionality. If, however, for some reasons it misses, `/run/modules` is used as search path.

The change itself is pretty easy: call twice the same logic for each tool with a different, customizable path (`MODULE_DIRECTORY` and `MODULE_ALTERNATIVE_DIRECTORY`).

I did not implement any test yet.

Original PR description:
```
Currently, traditional Linux systems using initramfs are forced to keep two copies of the modules which may be needed 
both during early boot and for normal system operation. This seems to be inevitable for the case when initramfs is built 
locally but with the rise of distro shipped UKIs the question whether it would be possible to avoid the redundancy arises.

To make it possible to reuse modules from initramfs without copying them to the root filesystem (which may be 
read-only), the suggested approach is to preserve them on tmpfs, e.g. in /run/modules/uname -r but the challenge is to 
make kmod search there.

[...] this part does not apply anymore [...]

Note, no merging of /run/modules/uname -r and /lib/modules/uname -r is performed. It is expected, that in case 
/lib/modules/uname -r exists, it is a superset and modules from initramfs are no longer needed.

To make a real world example of the usage, Fedora/CentOS/RHEL distros now ship 'kernel-uki-virt' package with a UKI 
which is self sufficient in many cases. The package, however, has to have a dependency on 'kernel-modules-core' so 
basically the same set of modules is available post boot. This is needed when e.g. a filesystem module which was not used 
for root filesystem need to be loaded to mount some other volume (and ESP is the most common example). The desire is 
to be able to break this dependency for 'minimal' cloud/VM images but still support the case when people want to install 
''kernel-modules-core'/'kernel-modules'/'kernel-modules-extra' post boot and enjoy additional modules.
```


esposem:
Sure, I thought the commit messages and PR description were self explanatory.
The idea is the following:
* If whatever component logic (say `kmod_new`) for the path MODULE_DIRECTORY (/lib/module) works correctly (no error returned), nothing else is done. The behavior is exactly as it was before this PR.
* If the component logic fails for the path MODULE_DIRECTORY, try MODULE_ALTERNATIVE_DIRECTORY (/run/modules) and see if it works.
This is done to cover the case where /lib/modules is missing but /run/modules is populated, and fall back into /run/modules ONLY IF /lib/modules is missing. The original /lib/modules has priority over the fallback directory.
Does it clear some doubts? Happy to explain if something is not clear

esposem:
Added doc and the last two commits show a possible fix to the confusing error messages that come up when the first call to MODULES_DIRECTORY fails but the second to MODULES_ALTERNATIVE_DIRECTORY is successful.

esposem:
Finished with first pass of fixes (hopefully waiting reviews). 
I modinfo should be completed, meaning it has functionality, docs and test. I fixed (or tried to) all problems described in the PR. Before doing other tests for the other components, I would love to get any feedback on this.

esposem:
Also added the optional build flag `enable-alternative-dir` to make this whole work completely optional at build time, and by default it's disabled.

I put the compile flag together with my solution for logs in separate commits, that can go squashed afterwards. I am waiting for reviews to see whether these options are useful or not :)

esposem:
Hello maintainers, can you please give a look at this PR? I think it's ready for a first pass :) @lucasdemarchi @evelikov

esposem:
Hello maintainers, can you please give a look at this PR? I think it's in shape for a first pass :) @lucasdemarchi @evelikov

esposem:
Hello maintainers, can you please give a look at this PR? I think it's in shape for a first pass :) @lucasdemarchi  @evelikov

esposem:
> In my earlier comment, I requested that the manual pages are updated to summarise all the discussions so far. Basically, how things work and why the order is the way it was proposed.

That's what I did :)
> 
> Some questions from very quickly (literally a few seconds) glancing through the commits:
> 
>     * what's the difference of this PR instead of the one opened previously

The main change here is that if /lib/modules exists, nothing changes in kmod functionality. If, however, for some reasons it misses, /run/modules is used as search path.
The one previously was doing all the way around.
       
>       * why not take that work and fixup as needed - original adds ~200 loc, this one ~500

See above, some commits are reused
 
>     * why are we changing the logger, instead of printing respective debug/error messages as needed

Because otherwise if the modules are in /run you get something like 
modinfo xfat
module not found! # because it is not in default path
<then comes the module info found in /run>

> 
>     * why do we add a build toggle which doesn't do anything
> 
which toggle

>     * have you considered extending `moduledir` to take a list of folders ... NixOS people have something similar IIRC

I can look at that. But when you have time, can you give this PR a more than a few seconds look?


> Overall I am onboard with the idea and would review the patches, once Lucas is happy with how things should work at a high level.

Nice

Issue #48, opened 2024-07-17
dlopen (de)compression libraries


lucasdemarchi:
Load (de)compression libraries ondemand - during normal use they are mostly not needed nowadays since we delegate that to the kernel. However when running some commands like modinfo/depmod we do need them. Distros may want to choose 1 of the decompression method that should be "built-in", but allowing the other libraries to be loaded ondemand.

So I think we also need a switch in the build system to select which library is dlopen'ed and which library is linked directly.

Depends on #47

evelikov:
As part of this work, or perhaps as follow-up, it would be great to add support for [`.note.dlopen`](https://github.com/systemd/systemd/blob/main/docs/ELF_DLOPEN_METADATA.md) and illustrated in this [systemd PR](https://github.com/systemd/systemd/pull/32234).

PR #262, opened 2024-11-30
Allow to dlopen compression libraries

lucasdemarchi:
Allow to dlopen the compression libraries so distros can choose one to match the MODULE_COMPRESS_* option in the kernel, while still allowing to work with other compressed formats.

I'm not very happy with the meson integration, but I gave up while trying to handle the dependencies between the options.

~~Still missing here: the ELF note to declare the implicit dependency.~~

Closes: #48

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -111,9 +111,10 @@ AC_ARG_WITH([zstd],
 	[], [with_zstd=no])
 AS_IF([test "x$with_zstd" != "xno"], [
 	PKG_CHECK_MODULES([libzstd], [libzstd >= 1.4.4], [LIBS="$LIBS $libzstd_LIBS"])
-	AC_DEFINE([ENABLE_ZSTD], [1], [Enable Zstandard for modules.])
+	AC_DEFINE([ENABLE_ZSTD], [1], [Zstandard for modules.])
 	module_compressions="zstd $module_compressions"
 ], [
+	AC_DEFINE([ENABLE_ZSTD], [0], [Zstandard for modules.])
 	AC_MSG_NOTICE([Zstandard support not requested])
 ])
 CC_FEATURE_APPEND([with_features], [with_zstd], [ZSTD])
@@ -124,9 +125,10 @@ AC_ARG_WITH([xz],
 	[], [with_xz=no])
 AS_IF([test "x$with_xz" != "xno"], [
 	PKG_CHECK_MODULES([liblzma], [liblzma >= 4.99], [LIBS="$LIBS $liblzma_LIBS"])
-	AC_DEFINE([ENABLE_XZ], [1], [Enable Xz for modules.])
+	AC_DEFINE([ENABLE_XZ], [1], [xz for modules.])
 	module_compressions="xz $module_compressions"
 ], [
+	AC_DEFINE([ENABLE_XZ], [0], [xz for modules.])
 	AC_MSG_NOTICE([Xz support not requested])
 ])
 CC_FEATURE_APPEND([with_features], [with_xz], [XZ])
@@ -137,9 +139,10 @@ AC_ARG_WITH([zlib],
 	[], [with_zlib=no])
 AS_IF([test "x$with_zlib" != "xno"], [
 	PKG_CHECK_MODULES([zlib], [zlib], [LIBS="$LIBS $zlib_LIBS"])
-	AC_DEFINE([ENABLE_ZLIB], [1], [Enable zlib for modules.])
+	AC_DEFINE([ENABLE_ZLIB], [1], [zlib for modules.])
 	module_compressions="gzip $module_compressions"
 ], [
+	AC_DEFINE([ENABLE_ZLIB], [0], [zlib for modules.])
 	AC_MSG_NOTICE([zlib support not requested])
 ])
 CC_FEATURE_APPEND([with_features], [with_zlib], [ZLIB])
@@ -150,9 +153,10 @@ AC_ARG_WITH([openssl],
 	[], [with_openssl=no])
 AS_IF([test "x$with_openssl" != "xno"], [
 	PKG_CHECK_MODULES([libcrypto], [libcrypto >= 1.1.0], [LIBS="$LIBS $libcrypto_LIBS"])
-	AC_DEFINE([ENABLE_OPENSSL], [1], [Enable openssl for modinfo.])
+	AC_DEFINE([ENABLE_OPENSSL], [1], [openssl for modinfo.])
 	module_signatures="PKCS7 $module_signatures"
 ], [
+	AC_DEFINE([ENABLE_OPENSSL], [0], [openssl for modinfo.])
 	AC_MSG_NOTICE([openssl support not requested])
 ])
 CC_FEATURE_APPEND([with_features], [with_openssl], [LIBCRYPTO])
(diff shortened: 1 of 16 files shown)


configure.ac
```diff
@@ -183,6 +183,11 @@ AC_ARG_WITH([zshcompletiondir],
 	[with_zshcompletiondir=${datadir}/zsh/site-functions])
 AC_SUBST([zshcompletiondir], [$with_zshcompletiondir])
 
+# Never enabled with autotools - let distros switch to meson if feature is wanted
+AC_DEFINE([ENABLE_DLOPEN_ZSTD], [0], [Enable dlopen zstd])
+AC_DEFINE([ENABLE_DLOPEN_XZ], [0], [Enable dlopen xz])
+AC_DEFINE([ENABLE_DLOPEN_ZLIB], [0], [Enable dlopen zlib])
```
evelikov, line 189: What is the reason behind the individual toggles - can we make it a single "use-dlopen-for-all" or alternatively indiscriminately switch to `dlopen` and remove the linking all together?

The systemd team have introduced a nice custom ELF section using which one can describe the (indirect) link dependency. IIRC tools like dracut, mkinitcpio, package managers have been (or in the process of) being update to make use of it.

I can open a follow-up PR adding the ELF section, if that's the blocker for us to always use `dlopen`

meson.build
```diff
@@ -287,23 +287,41 @@ endif
 # Compression support
 #-------------------------------------------------------------------------------
 
-zstd = dependency('libzstd', version : '>= 1.4.4', required : get_option('zstd'))
+zstd_dlopen = get_option('dlopen').contains('zstd')
+if get_option('zstd').disabled() and zstd_dlopen
+  error('Incompatible options: dlopen\'ed zstd needs zstd at build time')
+endif
```
evelikov, line 293: For the `dlopen` case we should be using `foo.partial_dependency(compile_args : true)`. As-is we explicitly pass `-lfoobar` and if we get a symbol snafu (don't properly use the wrapper macros) kmod will be linked against the library.

Ideally we'll go with the "just use dlopen" mentioned above, which will remove the build system complexity. Plus we don't risk having a semi-broken setup for the extra build permutations.

meson.build
```diff
@@ -287,23 +287,41 @@ endif
...
+zstd = zstd_dlopen ? dependency('libzstd', version : '>= 1.4.4', required : true) \
+                   : dependency('libzstd', version : '>= 1.4.4', required : get_option('zstd'))
 if zstd.found()
   cdata.set('ENABLE_ZSTD', true)
+  cdata.set10('ENABLE_DLOPEN_ZSTD', zstd_dlopen)
```
evelikov, line 298: `ENABLE_DLOPEN_ZSTD` is not defined in the `not zstd.found()` case. Which should result in build failure due to the `Wundef` we (rightfully) have.

meson.build
```diff
@@ -372,17 +390,17 @@ libkmod_deps = []
 
 if zstd.found()
   libkmod_files += files('libkmod/libkmod-file-zstd.c')
-  libkmod_deps += zstd
+  libkmod_deps += zstd_dlopen ? [] : zstd
```
evelikov, line 393: This will discard both `-I/usr/include/foobar` and `-Lfoobar`. We want to use `dependency` or `partial_dependency` as mentioned above.

evelikov (changes requested):
Nice, thanks for beating me to it.

configure.ac
lucasdemarchi: The ELF section is in a wip commit on top of these... didn't send that yet as I had to do other things before finalizing it.

My intention is to allow distros to still link directly. At least for a few versions... the ELF section thing is too new to rely on and the whole dlopen thing doesn't work if you statically link  (statically linking was disabled in autotools for other reasons, but I know some downstreams have/had patches to workaround the restriction).

Adding a -Ddlopen=all-enabled seems reasonable.

shared/util.c
```diff
@@ -558,3 +559,46 @@ unsigned long long stat_mstamp(const struct stat *st)
...
+{
+	va_list ap;
+	void *dl;
+	int r;
+
+	if (*dlp)
+		return 0;
+
+	dl = dlopen(filename, RTLD_LAZY);
+	if (!dl)
```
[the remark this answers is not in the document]
lucasdemarchi: I actually changed the return code so they are different and helps debugging if we need. If the library is not found, then it returns -ENOENT. If the symbol is not found then it returns -ENXIO.

meson.build
lucasdemarchi: I guess we should also have `includes : true` in that partial_dependency() call. I'm adding a few more refactors and doing that.

meson.build
lucasdemarchi: Fixed and also changed the rest of the build system to always define the ENABLE_* macros

meson.build
```diff
@@ -287,26 +288,30 @@ endif
...
-zlib = dependency('zlib', required : get_option('zlib'))
-if zlib.found()
-  module_compressions += 'zlib '
-endif
-cdata.set10('ENABLE_ZLIB', zlib.found())
-features += ['@0@ZLIB'.format(zlib.found() ? '+' : '-')]
+  dep = dependency(pkg_dep, version : pkg_dep_version, required : get_option(opt))
+  have = dep.found()
+
+  s = 'ENABLE_' + opt.to_upper()
```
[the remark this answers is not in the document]
lucasdemarchi: how? like this?

```
  cdata.set10('ENABLE_' + opt.to_upper(), have)
  cdata.set10('ENABLE_' + opt.to_upper() + '_DLOPEN', have and dlopen)
```
?

meson.build
```diff
@@ -176,6 +176,9 @@ module_signatures = ''
 features = []
 dep_map = {}
 
+# keep in sync with meson_options.txt
+dlopen_all = get_option('dlopen').contains('all')
```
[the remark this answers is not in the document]
lucasdemarchi: humn... So you mean to have 2 options?
-Ddlopen=[xz,zstd,zlib]
-Duse-dlopen?

I don't think we can mix the boolean and array in the same option, so -Ddlopen=true would be a no go.

If we are doing that, then maybe -Ddlopen-all=true or -Ddlopen-always=true ...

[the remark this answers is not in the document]
lucasdemarchi: But that is not what I want. At least not for now. We can change it once the elf-note becomes more mature or once distros choose just all-or-nothing. I want to allow 1 lib to be linked and the others to be dlopen'ed. With that there's no special handling needed in e.g. dracut/mkinitcpio and modinfo still works, at least for the compression the distro chose for kernel modules.

lucasdemarchi:
merged this by mistake while merging #265

meson.build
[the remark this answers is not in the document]
lucasdemarchi: I'm not convinced about removing the option to individually toggle them. For me it makes sense for distros to just link the 1 library the distro really cares about - that should match the decompression algorithm used by kernel (kernel accepts only 1) and the rest can be kept as dlopen so users can still work with other formats.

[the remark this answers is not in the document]
lucasdemarchi: I'm happy to simplify this later, even before the next release, but let's merge this and work on top.

lucasdemarchi:
> merged this by mistake while merging #264

and now pushed the right version

PR #264, opened 2024-12-01
Add TAKE_PTR()

lucasdemarchi:
Similar to macro in systemd codebase: add a macro that documents we are "leaking" a pointer that would otherwise be cleaned up by the cleanup attribute.

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -146,8 +146,9 @@ static int kmod_config_add_command(struct kmod_config *config, const char *modna
 	if (!l)
 		return -ENOMEM;
 
+	TAKE_PTR(cmd);
 	*list = l;
-	cmd = NULL;
+
 	return 0;
 }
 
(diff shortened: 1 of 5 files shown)


libkmod/libkmod-config.c
```diff
@@ -146,8 +146,9 @@ static int kmod_config_add_command(struct kmod_config *config, const char *modna
 	if (!l)
 		return -ENOMEM;
 
+	TAKE_PTR(cmd);
 	*list = l;
-	cmd = NULL;
```
evelikov, line 150: If we follow the approach used in `kmod_file_load_zlib()` aka `TAKE_PTR` on use, we'll get a memory leak in the above case. Since the annotations are off, this means clang-analyzer won't notice it, where only valgrind (in an error path) will flag it.

Generally I like the idea, although I am wondering if we cannot tweak the magic macro somehow, so we don't end up with issues like ^^

lucasdemarchi: Yeah, we can't do:

```
l = kmod_list_append(*list, TAKE_PTR(cmd));
```

and I didn't see a good alternative other than what I did here.

lucasdemarchi:
Applied, thanks

PR #265, opened 2024-12-02
tools: Unify help messages

stoeckmann:
Unify help messages shown when kmod is called without any arguments to use the same grammar and capitalisation.


lucasdemarchi:
Applied, thanks.

PR #268, opened 2024-12-18
strbuf: Extend strbuf_reserve_extra size check

stoeckmann:
Make sure that alignment to BUF_STEP won't overflow if excessively huge amount of memory is requested to be reserved.


lucasdemarchi:
Applied, thanks.

PR #275, opened 2025-01-06
Followup to LLD fixes

lucasdemarchi:


lucasdemarchi:
Later I also added a CI configuration for that. It was harder to manage with autotools in the tree, but now it's gone.

PR #279, opened 2025-02-15
libkmod: Check correct ELF header size

stoeckmann:
Check if enough bytes are available for Elf32_Ehdr, not Elf32_Shdr. The size of Elf32_Shdr is needed later on.

Since both are 64 bytes, no functional change.


lucasdemarchi:
Applied, thanks.

Issue #277, opened 2025-02-07
Assertion `idx < elf->header.section.count' in elf_get_section_header


yonghong-song:
In Meta, we used Google's PGO patch
   https://patchwork.kernel.org/project/linux-kbuild/patch/20210407211704.367039-1-morbo@google.com/
to build the kernel for performance reason.

Recently when we enabled amdgpu, we found that build failed with the following error message:

```
  INSTALL /root/rpmbuild/BUILDROOT/kernel-6.13.1-0_fbk0_lto_rc5_166_gb4a9a3553c165.x86_64/lib/modules/6.13.1-0_fbk0_lto_rc5_166_gb4a9a3553c165/kernel/net/9p/9pnet_virtio.ko
  DEPMOD  /root/rpmbuild/BUILDROOT/kernel-6.13.1-0_fbk0_lto_rc5_166_gb4a9a3553c165.x86_64/lib/modules/6.13.1-0_fbk0_lto_rc5_166_gb4a9a3553c165
depmod: libkmod/libkmod-elf.c:214: elf_get_section_header: Assertion `idx < elf->header.section.count' failed.
make[2]: *** [scripts/Makefile.modinst:132: depmod] Aborted (core dumped)
make[1]: *** [/rw/kernel/Makefile:1904: modules_install] Error 2
make: *** [Makefile:251: __sub-make] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.3Q0MIr (%install)
```

The depmod version,
```
$ depmod --version
kmod version 28
+ZSTD +XZ +ZLIB +LIBCRYPTO -EXPERIMENTAL
```

Look at amdgpu.ko number of sections:
```
$ llvm-readelf -S amdgpu.ko
...
  [79357] .group         GROUP           0000000000000000 2675e628 000010 04     79383 40006  4
  [79358] __llvm_prf_cnts PROGBITS       0000000000000000 c4bd68 0002c0 00 WAG  0   0  8
  [79359] __llvm_prf_data PROGBITS       0000000000000000 c4c028 000040 00 WAG  0   0  8
  [79360] .rela__llvm_prf_data RELA      0000000000000000 2675e638 000018 18  IG 79383 79359  8
  [79361] .group         GROUP           0000000000000000 2675e650 000010 04     79383 40007  4
  [79362] __llvm_prf_cnts PROGBITS       0000000000000000 c4c068 000010 00 WAG  0   0  8
[...]

PR #280, opened 2025-02-15
Support modules with a lot of sections

stoeckmann:
It seems that it's possible to compile modules with more than 0xFF00 sections. This leads to different fields in ELF files used to handle such a large amount of sections.

Parse proper fields if necessary.

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -52,11 +52,11 @@ struct kmod_elf {
 	struct {
 		struct {
 			uint64_t offset;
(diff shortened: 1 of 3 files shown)


stoeckmann:
I've kept it as a draft because the linux kernel itself cannot support such modules either (as confirmed in https://github.com/kmod-project/kmod/issues/277). I have used this PR for easier testing/reference to the linked issue.

If we want to support possible features of future kernels we could merge this, but right now it's just a possible way of introducing bugs for no practical gain.

PR #281, opened 2025-02-17
Prepare NEWS for v34

lucasdemarchi:


lucasdemarchi:
@evelikov @stoeckmann 

Here are the NEWS entries for v34. I went through all commits and tried to summarize the changes as usual. This time the release is much bigger and I may have missed some important things. Could you take a look and check if there are other things that should be mentioned?

NEWS
```diff
@@ -1,3 +1,196 @@
...
+
+	- Remove depmod_module_directory_override from .pc as the kernel side
+	  is not making use of it and will likely not need it.
+
+	- Improve builtin module listing and retrieving information from its
+	  modinfo index which reduces in up to 90% the amount of syscalls
+	  needed.
+
+	- Improve zstd decompression by using streaming bufferless mode which
+	  reduces the amount of syscalls in 65%.
```
stoeckmann, line 84: in -> by

NEWS
```diff
@@ -1,3 +1,196 @@
...
+
+	- Increase use of pread while parsing ELF and indexes in order to reduce
+	  syscalls and improve performance.
+
+	- Improve module sorting in depmod to speedup the use of the
+	  modules.order index and support duplicate lines in it.
+
+	- Avoid misaligned memory access while reading module signature libkmod.
```
stoeckmann, line 92: in libkmod

NEWS
```diff
@@ -1,3 +1,196 @@
...
+	- Overhaul man pages with multiple clarifications, section rewrites and
+	  additional documentation.
+
+	- Drop --with-rootlibdir as it's seldom used and was partially broken.
+
+	- Drop strndupa() and alloca() for increased libc compatibility.
+
+	- Better handling of LFS for increased compatibility with libc.
+
+	- Protect kmod_get_dirname() and XXXXXX against NULL argument.
```
stoeckmann, line 107: XXXXXX something's missing here, most likely TODO but just pointing it out. :)

lucasdemarchi:
> Do you have a date in mind when to release the next version?

This week. Not decided yet if we are going to have an -rc or not. For fixes are still open (I need to take a look on your recent PR), but I'd rather not take any major refactor or feature for this version.

stoeckmann:
Yeah, definitely not thinking about anything big. More like: Time for re-reading everything again. :D

lucasdemarchi:
The dist for meson vs autotools seem to be a little bit different. I think we can get away by just generating the one from meson, hoping distro will make the switch in this release. For autotools, I think that as long as they call autogen.sh, they should be fine.

eli-schwartz:
You can also run atogen.sh from within [`meson.add_dist_script()`](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonadd_dist_script) if you really like. :)

lucasdemarchi:
@eli-schwartz would it be sufficient to just pass autogen.sh?

lucasdemarchi:
> @eli-schwartz would it be sufficient to just pass autogen.sh?

@eli-schwartz  just added a commit to do that... I **think** that should be sufficient. At least the configure script appears to be there.

```
$ tar -tf /home/ldmartin/p/kmod/build/meson-dist/kmod-33.tar.xz | grep configure
kmod-33/configure
kmod-33/configure.ac
```

eli-schwartz:
Yup, was about to say you'd need to detect MESON_DIST_ROOT. :)

lucasdemarchi:
Applied.

PR #282, opened 2025-02-18
tools/log.c related cleanups

stoeckmann:
Things noticed while looking at `tools/log.c`:

- Use `bool` for `use_syslog` variables since they are used for `log_open(bool)` as argument anyway
- Reduce variable visibility
- Add a missing newline


lucasdemarchi:
Applied, thanks.

PR #284, opened 2025-02-19
kmod: use program_invocation_short_name more often

stoeckmann:
Remove the explicit basename(argv[0]) calls because we already rely on program_invocation_short_name even before reaching these lines. Unifies code and slightly reduces binary size.


lucasdemarchi:
Applied, thanks.

PR #285, opened 2025-02-19
static-nodes: unify return statements

stoeckmann:
The write_* functions have different return statement handling. Unify them by removing "else" if the if-block itself returns and also unify the error handling by checking for the error case, leaving the success return statement at the end of the function.


lucasdemarchi:
Applied, thanks.

PR #287, opened 2025-02-19
Unify manual pages and help output of tools

stoeckmann:
- Long options use `=` between key and value in help output, so manual pages should do the same (common in other projects as well)
- Add missing `=` in modprobe help output for --wait
- Add missing -h description in modinfo


lucasdemarchi:
Applied, thanks.

PR #289, opened 2025-02-21
Drop autotools

lucasdemarchi:
I'm probably going to wait a week or so to drop it. Because if we have fixes to apply to help distros, it's easier if we have autotools in tree. However we should remove this soon.


autogen.sh
```diff
@@ -1,75 +0,0 @@
...
-
-if [ -n "$MESON_DIST_ROOT" ]; then
-    topdir="$MESON_DIST_ROOT"
-else
-    topdir=$(dirname $0)
-fi
-
-cd $topdir
-
-gtkdocize --docdir libkmod/docs || touch libkmod/docs/gtk-doc.make
```
stoeckmann, line 15: This does not work with release tarball if gtk-doc is not installed. `libkmod/docs/gtk-doc.make` is already a symbolic link:
```
libkmod/docs/gtk-doc.make -> /usr/share/gtk-doc/data/gtk-doc.flat.make
```

Same is true for m4/gtk-doc.m4:
```
m4/gtk-doc.m4 -> /usr/share/aclocal/gtk-doc.m4
```

stoeckmann: We would need an `rm -f` call for both mentioned files before touch to make sure that it always works.

stoeckmann (commented):
I've noticed an issue with running `autogen.sh` of release tarball when I wanted to figure out why meson's resulting binary is larger than the previous version (turns out it's `-O2` (autoconf) vs `-O3` (meson).

Since we already say that distributions should switch and since they most likely have gtk-doc in their build environment, I'll keep it as a note on this PR.

autogen.sh
lucasdemarchi: and there's no point fixing it in git since it's only reproducible when building from the tarball, correct?

stoeckmann: Correct, that's why I didn't notice it during tests.

stoeckmann: See https://github.com/kmod-project/kmod/pull/295

.github/workflows/main.yml
```diff
@@ -30,10 +30,10 @@ jobs:
             meson_setup: '-Ddocs=false -Db_sanitize=none'
           - compiler: 'gcc'
             container: 'archlinux:multilib-devel'
-            multilib: 'true'
+            x32: 'true'
```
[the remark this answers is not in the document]
lucasdemarchi: we were building things as 32b because we have multilib. That's not what we want to test - as you said multilib and x32 are not the same thing. Now x32 is explicit and means "we are testing 32b builds".

.github/workflows/main.yml
```diff
@@ -28,9 +28,13 @@ jobs:
           - compiler: 'gcc'
             container: 'alpine:latest'
```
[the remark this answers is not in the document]
lucasdemarchi: I don't think you understood the issue here. We were adding an entry with x32 and also passing a meson_setup option in that entry. However meson_setup was completely ignored and we were rather using hardcoded options. We can rather use the  relevant options in the entry only.

.github/workflows/main.yml
```diff
@@ -104,6 +104,14 @@ jobs:
             container: 'archlinux:multilib-devel'
             linker: 'lld'
 
+          # Variants with moduledir
+          - compiler: 'gcc'
+            container: 'archlinux:multilib-devel'
+            meson_setup: '-Dmoduledir=/usr/lib/modules'
```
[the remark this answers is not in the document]
lucasdemarchi: passing the option or not is handled differently in meson. Previously doing  '-Dmoduledir=/usr/lib/modules' because of the absolute path handling. We can stop testing it once we have too many jobs and it doesn't provide a value anymore.

.github/workflows/main.yml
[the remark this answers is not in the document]
lucasdemarchi: I think splitting it in 2 places is error prone -  the fact that most of our builds were actually testing 32b rather than 64b and on top of that meson_setup from the environment was ignored shows it. That made me move to "explicit is better than implicit" approach for this case.

PR #290, opened 2025-02-22
modinfo: reduce amount of continue statements

stoeckmann:
Clarify that the code does not perform error checks, for which we regularly use checks and continue at the start, but merely checks how to print/filter the data.

Nice benefit: Reduces binary size if compiled with GCC.


lucasdemarchi:
mixed feelings about this...  Before we would have something like:

```
/* filtered output contains no key, just value */
...
```

And we are done: we can process the next one. Now we have to follow the entire loop to see if it's going to do something else.  I don't think the continue statements are only for error handling. Anyway, I'm applying this one.

PR #293, opened 2025-02-23
depmod: handle out of memory condition

stoeckmann:
The strbuf_str call may return NULL in case of out of memory condition. Check its return value before dereferencing the pointer.


stoeckmann:
This PR would be covered by https://github.com/kmod-project/kmod/pull/293 with a different approach.

lucasdemarchi:
> This PR would be covered by #293 with a different approach.

do you mean #296?

lucasdemarchi:
Close in favor of #296

PR #294, opened 2025-02-24
ci: add pahole for Debian unstable

stoeckmann:
The pahole package is required in our test suite.

### DIFF
--- a/.github/actions/setup-debian/action.yml
+++ b/.github/actions/setup-debian/action.yml
@@ -28,6 +28,7 @@ runs:
           libzstd-dev \
           linux-headers-generic \
           meson \
+          pahole \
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
AFAICS is because a kernel config update started to use BTF and then we have these errors now in CI:

```
  BTF [M] /__w/kmod/kmod/build/testsuite/module-playground/mod-simple.ko
/bin/sh: 1: pahole: not found
make[3]: *** [/usr/src/linux-headers-6.12.16-common/scripts/Makefile.modfinal:64: /__w/kmod/kmod/build/testsuite/module-playground/mod-simple.ko] Error 127
```

Can you add that to the commit message?

PR #295, opened 2025-02-24
build: support missing gtkdocize in releases

stoeckmann:
The release tarballs may already contain symbolic links into the file system for gtk-doc specific files.

If gtk-doc is not installed, and thus not desired for building, remove the symbolic links and create empty files instead.


lucasdemarchi:
Applied, thanks.

Issue #298, opened 2025-02-25
kmod 34 linking fails on Debian Bullseye with missing dlclose reference


HiassofT:
When trying to build kmod 34 on a Debian Bullseye system the link step fails with an undefined reference to dlclose:

```
# meson setup -Dzstd=disabled -Dxz=disabled -Dzlib=disabled -Dopenssl=enabled -Dtools=true -Dbuild-tests=false -Ddocs=false -Dmanpages=false /tmp/kmod-34
# meson compile
...
[27/33] Linking target kmod
FAILED: kmod 
cc  -o kmod kmod.p/tools_depmod.c.o kmod.p/tools_insmod.c.o kmod.p/tools_kmod.c.o kmod.p/tools_log.c.o kmod.p/tools_lsmod.c.o kmod.p/tools_modinfo.c.o kmod.p/tools_modprobe.c.o kmod.p/tools_opt.c.o kmod.p/tools_rmmod.c.o kmod.p/tools_static-nodes.c.o -Wl,--as-needed -Wl,--no-undefined -pie -Wl,--gc-sections -Wl,--start-group libshared.a libkmod-internal.a /usr/lib/x86_64-linux-gnu/libcrypto.so -Wl,--end-group
/usr/bin/ld: libshared.a.p/shared_util.c.o: undefined reference to symbol 'dlclose@@GLIBC_2.2.5'
/usr/bin/ld: /lib/x86_64-linux-gnu/libdl.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

The linker line is missing `-ldl` which is needed with the older GNU toolchain.

Building on Debian Bookworm works fine and running `LDFLAGS="-ldl" meson setup ....` on Bullseye works around the issue but it should better be fixed in the build system.

Here's a simple containerfile to reproduce the issue:
```
FROM docker.io/debian:bullseye-slim AS kmod-build
RUN \
    echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list \
    && apt-get update \
    && apt-get -y install build-essential libssl-dev pkg-config curl \
    && apt-get -y -t bullseye-backports install meson ninja-build

FROM kmod-build
RUN \
    curl https://www.kernel.org/pub/linux/utils/kernel/kmod/kmod-34.tar.xz | tar Jxf - -C /tmp/ \
    && mkdir /tmp/kmod-build \
    && cd /tmp/kmod-build \
[...]

PR #299, opened 2025-02-25
meson: Fix build with glibc 2.31

lucasdemarchi:
In order to use dlopen it may be required to link with libdl depending on the libc. Add the proper dependency to fix the build in Debian Bullseye.

Closes: https://github.com/kmod-project/kmod/issues/298


lucasdemarchi:
Applied, closing.

PR #303, opened 2025-03-03
Meson build fixes

hendrikdonner:


hendrikdonner:
Added Signed-off-by tags.

meson.build
```diff
@@ -220,7 +220,7 @@ foreach tuple : _customdirs
           .format(dir_option, customdir))
     endif
     # Strip all leading/trailing and re-add only the leading one.
-    customdir = '' / customdir.strip('/')
+    customdir = '/' / customdir.strip('/')
```
lucasdemarchi, line 223: See a few lines above. We just abort if it's not absolute path

meson.build
```diff
@@ -513,7 +513,7 @@ if get_option('tools')
   mkdir_p = 'mkdir -p "$DESTDIR@0@"'
   meson.add_install_script('sh', '-c', mkdir_p.format(sbindir))
 
-  ln_s = 'ln --symbolic --force --relative "$DESTDIR@0@/kmod" "$DESTDIR@1@"'
+  ln_s = 'ln -sfr "$DESTDIR@0@/kmod" "$DESTDIR@1@"'
```
lucasdemarchi, line 516: what happens if some (other) implementations just implement the long options?

If we are **building** kmod, can't we simply have the complete tool rather than the more capped ones? Is this for supporting it any distro?

meson.build
hendrikdonner: Yes, and than the stripping breaks it. If you currently build with `meson setup -Dmoduledir=/usr/lib/modules` it will turn into this:

```
 Kmod specific
    distconfdir      : /usr/lib
    moduledir        : usr/lib/modules
```

Re-adding the leading / does not work right now.

meson.build
hendrikdonner: > what happens if some (other) implementations just implement the long options?

Than it can already break in https://github.com/kmod-project/kmod/blob/master/meson.build#L527 i guess?

> If we are building kmod, can't we simply have the complete tool rather than the more capped ones? Is this for supporting it any distro?

Breaks for me, we have an in house distro for mostly embedded firmware. We have a slightly patched busybox.

meson.build
lucasdemarchi: can you share your meson version?  meson should handle that...

```
$ meson setup  -Ddistconfdir=/usr/lib -Dmoduledir=/usr/lib/modules build
...
  User defined options
    distconfdir      : /usr/lib
    moduledir        : /usr/lib/modules
$ meson --version
1.7.0
```

Testing the same thing with debian bullseye with meson 1.0.0 (coming in #289 ), it still works. Forcing our min meson version (0.61.0) also works.

Trying to go back to 0.56.2 (which is unsupported and would need a different patch in meson.build):

```
  Kmod specific
          distconfdir: /usr/lib
            moduledir: usr/lib/modules
```

So are you using an older meson?

meson.build
lucasdemarchi: ok... makes sense. Let's get this in.

meson.build
hendrikdonner: No:

```
The Meson build system
Version: 1.7.0
Source dir: /builds/0/pkgbuilds/core/kmod/src/kmod
Build dir: /builds/0/pkgbuilds/core/kmod/src/kmod/build
Build type: native build
Project name: kmod
Project version: 34
....
Build targets in project: 10
kmod 34
  Directories
    prefix           : /usr
    sysconfdir       : /etc
    bindir           : /usr/bin
    sbindir          : /usr/bin
    includedir       : /usr/include
    libdir           : /usr/lib
    datadir          : /usr/share
  Kmod specific
    distconfdir      : /usr/lib
    moduledir        : usr/lib/modules
```

I can also reproduce on Ubuntu 24.04.2 LTS with

```
%:meson --version
1.3.2
```

see:

```
%:meson setup --reconfigure -Dmoduledir=/usr/lib/modules . build
The Meson build system
Version: 1.3.2
...
Build targets in project: 10

kmod 34

  Directories
    prefix           : /usr
    sysconfdir       : /etc
    bindir           : /usr/bin
    sbindir          : /usr/sbin
    includedir       : /usr/include
    libdir           : /usr/lib/x86_64-linux-gnu
    datadir          : /usr/share

  Kmod specific
    distconfdir      : /usr/lib
    moduledir        : usr/lib/modules
```

lucasdemarchi: My example above was showing the "User defined options" rather than "Kmod specific". Looking at the right place, it indeed shows as  relative and indeed causes the testsuite to fail when built with meson rather than autotools.

Looks good indeed. Thanks.

hendrikdonner: Oh right, "User defined options" looks like directly coming from meson and is the unprocessed input.

lucasdemarchi:
Applied both patches. Thanks!

hendrikdonner:
Nice, thanks!

PR #305, opened 2025-03-05
Depmod: Add tmpfile-util to generate temporary file

### LINKED ISSUE #51: depmod: Use O_TMPFILE for temporary file creation
Add abstraction in shared/ for temp file handling and use it in depmod to atomically (from the user POV) deal with the creation of new indexes

gongjun-song:
Closes: #51 

Considering O_TMPFILE is not supported on all linux system (refer by [here)](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=O_TMPFILE%20requires%20support%20by%20the,Linux%20filesystems%20provide%20that%20support.). And mkstemp is a more general function to create temp file. We use mkstemp to refactor this code.

Signed-off-by: Wenjie Wang <wenjie2.wang@intel.com>
Signed-off-by: Gongjun Song <gongjun.song@intel.com>
Signed-off-by: Dan He <dan.h.he@intel.com>
Signed-off-by: Qingqing Li <qingqing.li@intel.com>
Signed-off-by: Yuchi Chen <yuchi.chen@intel.com>

### DIFF
--- a/shared/tmpfile-util.c
+++ b/shared/tmpfile-util.c
@@ -1,0 +1,102 @@
+
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * Copyright (C) 2025  Intel Corporation.
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <libgen.h>
+#include <limits.h>
+#include <stdarg.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "tmpfile-util.h"
+#include "macro.h"
+#include "util.h"
+
+FILE *tmpfile_openat(int dirfd, const char *tmpname_tmpl, mode_t mode,
+		     struct tmpfile *file)
+{
+	const char *tmpname;
+	char tmpfile_path[PATH_MAX];
+	int fd, n, err;
+	_cleanup_free_ char *targetdir;
+	FILE *fp;
+
+	if (file == NULL || tmpname_tmpl == NULL) {
+		return NULL;
+	}
+
+	err = fd_lookup_path(dirfd, &targetdir);
+	if (err < 0) {
+		goto create_fail;
+	}
+
+	n = snprintf(tmpfile_path, PATH_MAX, "%s/%s", targetdir, tmpname_tmpl);
+	if (n < 0 || n >= PATH_MAX) {
+		goto create_fail;
+	}
+
+	fd = mkstemp(tmpfile_path);
+	if (fd < 0) {
+		goto create_fail;
+	}
+
+	if (fchmod(fd, mode) < 0) {
+		goto checkout_fail;
+	}
+
+	fp = fdopen(fd, "wb");
+	if (fp == NULL) {
+		goto checkout_fail;
+	}
+
+	tmpname = basename(tmpfile_path);
+	memset(file->tmpname, 0, PATH_MAX);
+	memcpy(file->tmpname, tmpname, strlen(tmpname));
+
+	file->dirfd = dirfd;
+	file->fd = fd;
+
+	return fp;
+
+checkout_fail:
+	close(fd);
+	remove(tmpfile_path);
+create_fail:
+	return NULL;
+}
+
+int tmpfile_publish(struct tmpfile *file, const char *targetname)
+{
+	if (file == NULL || targetname == NULL) {
+		return -EINVAL;
+	}
+
+	if (renameat(file->dirfd, file->tmpname, file->dirfd, targetname) != 0) {
+		return -errno;
+	}
+
+	file->fd = 0;
+	file->dirfd = 0;
+	memset(file->tmpname, 0, PATH_MAX);
+	return 0;
+}
+
+void tmpfile_release(struct tmpfile *file)
+{
+	if (file == NULL) {
+		return;
+	}
+
+	unlinkat(file->dirfd, file->tmpname, 0);
+
+	file->fd = 0;
+	file->dirfd = 0;
+	memset(file->tmpname, 0, PATH_MAX);
+}
--- a/shared/tmpfile-util.c
+++ b/shared/tmpfile-util.c
@@ -1,4 +1,3 @@
-
 // SPDX-License-Identifier: LGPL-2.1-or-later
 /*
  * Copyright (C) 2025  Intel Corporation.
@@ -15,27 +14,22 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
-#include "tmpfile-util.h"
-#include "macro.h"
-#include "util.h"
+#include <shared/macro.h>
+#include <shared/tmpfile-util.h>
+#include <shared/util.h>
 
-FILE *tmpfile_openat(int dirfd, const char *tmpname_tmpl, mode_t mode,
-		     struct tmpfile *file)
+FILE *tmpfile_openat(int dirfd, mode_t mode, struct tmpfile *file)
 {
+	const char *tmpname_tmpl = "tmpfileXXXXXX";
 	const char *tmpname;
 	char tmpfile_path[PATH_MAX];
-	int fd, n, err;
+	int fd, n;
 	_cleanup_free_ char *targetdir;
 	FILE *fp;
 
-	if (file == NULL || tmpname_tmpl == NULL) {
-		return NULL;
-	}
-
-	err = fd_lookup_path(dirfd, &targetdir);
-	if (err < 0) {
+	targetdir = fd_lookup_path(dirfd);
+	if (targetdir == NULL)
 		goto create_fail;
-	}
 
 	n = snprintf(tmpfile_path, PATH_MAX, "%s/%s", targetdir, tmpname_tmpl);
 	if (n < 0 || n >= PATH_MAX) {
@@ -43,18 +37,15 @@ FILE *tmpfile_openat(int dirfd, const char *tmpname_tmpl, mode_t mode,
 	}
 
 	fd = mkstemp(tmpfile_path);
-	if (fd < 0) {
+	if (fd < 0)
 		goto create_fail;
-	}
 
-	if (fchmod(fd, mode) < 0) {
+	if (fchmod(fd, mode) < 0)
 		goto checkout_fail;
-	}
 
 	fp = fdopen(fd, "wb");
-	if (fp == NULL) {
+	if (fp == NULL)
 		goto checkout_fail;
-	}
 
 	tmpname = basename(tmpfile_path);
 	memset(file->tmpname, 0, PATH_MAX);
@@ -74,13 +65,8 @@ create_fail:
 
 int tmpfile_publish(struct tmpfile *file, const char *targetname)
 {
-	if (file == NULL || targetname == NULL) {
-		return -EINVAL;
-	}
-
-	if (renameat(file->dirfd, file->tmpname, file->dirfd, targetname) != 0) {
+	if (renameat(file->dirfd, file->tmpname, file->dirfd, targetname) != 0)
 		return -errno;
-	}
 
 	file->fd = 0;
 	file->dirfd = 0;
@@ -90,10 +76,6 @@ int tmpfile_publish(struct tmpfile *file, const char *targetname)
 
 void tmpfile_release(struct tmpfile *file)
 {
-	if (file == NULL) {
-		return;
-	}
-
 	unlinkat(file->dirfd, file->tmpname, 0);
 
 	file->fd = 0;
--- a/shared/tmpfile-util.h
+++ b/shared/tmpfile-util.h
@@ -1,7 +1,14 @@
+// SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * Copyright (C) 2025  Intel Corporation.
+ */
+
 #pragma once
 
 #include <stdio.h>
 #include <linux/limits.h>
+
+#include <shared/macro.h>
 
 struct tmpfile {
 	char tmpname[PATH_MAX];
@@ -12,15 +19,15 @@ struct tmpfile {
 /*
  * Create a temporary file at the directory of `dirfd`
 */
-FILE *tmpfile_openat(int dirfd, const char *tmpname_tmpl, mode_t mode,
-		     struct tmpfile *file);
+_must_check_ _nonnull_(3) FILE *tmpfile_openat(int dirfd, mode_t mode,
+					       struct tmpfile *file);
 
 /*
  * Move the temporary file to `targetname`
 */
-int tmpfile_publish(struct tmpfile *file, const char *targetname);
+_nonnull_all_ int tmpfile_publish(struct tmpfile *file, const char *targetname);
 
 /*
  * Delete the temporary file
 */
-void tmpfile_release(struct tmpfile *file);
+_nonnull_all_ void tmpfile_release(struct tmpfile *file);
--- a/shared/tmpfile-util.h
+++ b/shared/tmpfile-util.h
@@ -1,0 +1,26 @@
+#pragma once
+
+#include <stdio.h>
+#include <linux/limits.h>
+
+struct tmpfile {
+	char tmpname[PATH_MAX];
+	int dirfd;
+	int fd;
+};
+
+/*
+ * Create a temporary file at the directory of `dirfd`
+*/
+FILE *tmpfile_openat(int dirfd, const char *tmpname_tmpl, mode_t mode,
+		     struct tmpfile *file);
+
+/*
+ * Move the temporary file to `targetname`
+*/
+int tmpfile_publish(struct tmpfile *file, const char *targetname);
+
+/*
+ * Delete the temporary file
+*/
+void tmpfile_release(struct tmpfile *file);
--- a/shared/util.c
+++ b/shared/util.c
@@ -476,6 +476,31 @@ int mkdir_parents(const char *path, mode_t mode)
 	return mkdir_p(path, end - path, mode);
 }
 
+int fd_lookup_path(int fd, char **ret_path)
+{
+	char proc_path[PATH_MAX];
+	char fd_path[PATH_MAX];
+	ssize_t len;
+
+	len = snprintf(proc_path, sizeof(proc_path), "/proc/self/fd/%d", fd);
+	if (len < 0 || len >= (ssize_t)sizeof(proc_path)) {
+		return -errno;
+	}
+
+	len = readlink(proc_path, fd_path, sizeof(fd_path) - 1);
+	if (len < 0 || len >= (ssize_t)sizeof(fd_path)) {
+		return -errno;
+	}
+
+	fd_path[len] = '\0';
+
+	if (ret_path != NULL) {
+		*ret_path = strdup(fd_path);
+	}
+
+	return 0;
+}
+
 static unsigned long long ts_usec(const struct timespec *ts)
 {
 	return (unsigned long long)ts->tv_sec * USEC_PER_SEC +
--- a/shared/util.c
+++ b/shared/util.c
@@ -476,7 +476,7 @@ int mkdir_parents(const char *path, mode_t mode)
 	return mkdir_p(path, end - path, mode);
 }
 
-int fd_lookup_path(int fd, char **ret_path)
+char *fd_lookup_path(int fd)
 {
 	char proc_path[PATH_MAX];
 	char fd_path[PATH_MAX];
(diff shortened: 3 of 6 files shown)


stoeckmann:
Correct me if I'm wrong but this code does not use `O_TMPFILE` at all. Instead, it's a hand-written version of `mkstemp` or more precisely something like `mkstempat` since it uses a directory handle.

tools/depmod.c
```diff
@@ -2612,13 +2613,15 @@ static int depmod_output(struct depmod *depmod, FILE *out)
...
-		char tmp[NAME_MAX] = "";
+		// char tmp[NAME_MAX] = "";
+        _cleanup_free_ char *tmp = NULL;
 		int r, ferr;
 
 		if (fp == NULL) {
 			int flags = O_CREAT | O_EXCL | O_WRONLY;
 			int mode = 0644;
 			int fd;
+#if 0
```
lucasdemarchi: please don't comment code out. Just remove it. But this should be done in a separate commit, not this one.

The fist commit adding shared/tmpfile-util.c should come with the relevant tests in the testsuite. Then later when making use of it, there should be a commit with `depmod: Use common code for tmpfile handling`.

shared/tmpfile-util.c
```diff
@@ -0,0 +1,187 @@
...
+
+int fopen_temporary_at(int dir_fd, const char *path, int o_flags, int o_mode, int *ret_fd, char **ret_path) {
+    _cleanup_free_ char *tmp_file = NULL;
+    int fd = -EBADF;
+    int r;
+
+    if (dir_fd < 0 || !path)
+        return -EINVAL;
+
+    r = tempfn_random(path, &tmp_file);
```
lucasdemarchi: Apparently you are trying to keep a similar behavior as of we have today, creating a file. But I'm missing the actual use of O_TMPFILE. With that, there's no file visible in the filesystem... you have an fd to an invisible file that later you can atomically make it show up, **after** you finished writing to it.

If you are first adding the API and later converting it to O_TMPFILE, then we need to balance it: shared/tmpfile-util.c should be responsible by both creating it and later linking it on the filesystem:

FILE *tmpfile_fopenat(int dir_fd, int o_flags, int o_mode);
int tmpfile_flinkat(FILE *file, int dirfd, const char *fn);

Makefile.am
```diff
@@ -65,6 +65,8 @@ shared_libshared_la_SOURCES = \
 	shared/missing.h \
 	shared/strbuf.c \
 	shared/strbuf.h \
+	shared/tmpfile-util.c \
+	shared/tmpfile-util.h \
```
lucasdemarchi, line 69: we only care about meson right now.

shared/tmpfile-util.c
```diff
@@ -1,4 +1,7 @@
 // SPDX-License-Identifier: LGPL-2.1-or-later
+/*
+ * Copyright (C) 2025  Wang Wenjie <wenjie2.wang@intel.com>.
```
lucasdemarchi, line 3: this should be in the first patch:
```
/*
 * Copyright © 2025 Intel Corporation
 */
```

lucasdemarchi (changes requested):

wenjiewang-intel:
Considering `O_TMPFILE` is not supported on all linux system (refer by [here)](https://man7.org/linux/man-pages/man2/open.2.html#:~:text=O_TMPFILE%20requires%20support%20by%20the,Linux%20filesystems%20provide%20that%20support. ). And `mkstemp` is a more general function to create temp file. We use `mkstemp` to refactor this code.

shared/tmpfile-util.c
wenjiewang-intel: fixed in new commit

tools/depmod.c
wenjiewang-intel: fixed in new commit

shared/tmpfile-util.c
wenjiewang-intel: In new commit, we use `mkstemp` to generate a temporary file. Is it OK?

Makefile.am
wenjiewang-intel: If we don't add these two files to Makefile.am, Autotools will not successfully compile, and CI will not pass. Are we going to ignore this error?

lucasdemarchi: ok. Just make it part of the commit adding the file rather than as the last commit. If this goes in before #289 , then I rebase it that will drop the entire thing.

wenjiewang-intel:
> Thanks for the help Gongjun Song and team.
> 
> Can you please carefully go through our [contributing guide](https://github.com/kmod-project/kmod/blob/master/CONTRIBUTING.md). At a glance, the first 3 sections seem applicable here. Also out of curiosity - did all 5 individuals participate in creating the patch(es)?
> 
> For this PR, since we are not doing anything with O_TMPFILE let's update the PR title and summary. While the commit message itself, should mention why the new approach is better than the older one.

Hi Emil, we modified the commit message to explain why using `mkstemp`. All 5 of us have participated in discussion and review for the patch.

About the PR title, we don't have permission to update, please @lucasdemarchi help to update the title if `mkstemp` is accepted.

wenjiewang-intel:
> I think it would be nice to get this more into shape, i.e. getting the API closer to shared/ than systemd.
> 
> Something like `struct tmpfile` with a `tmpfile_init` function, which would have a `tmpfile_publish` function or a `tmpfile_release` for either publishing the file or removing it.
> 
> We could use the cleanup traits, too.
> 
> In short: Copy the idea of strbuf for "tmpfile" although tmpfile itself might be a bad name. After all it's just a temporary stage of a file which is supposed to be published.
> 
> And talking about all that: Do we really need that for shared? Which library function do we have that wants to create a file?

Hi Stoeckmann, we don't fully understand your meaning, could you explain more?

What fields should be in struct tmpfile? Should be a struct like following?
```c
struct tmpfile {
    char *parent;              // the parent path of this temporary file
    char *name;               // the name of current temporary file
};

int tmpfile_init(struct *tmpfile, char *parent); // create a temporary under `parent` path.
int tmpfile_publish(struct *tmpfile, char *path);   // move the `tmpfile` to `path` 
int tmpfile_release(struct *tmpfile);     // delete the temporary
```

stoeckmann:
> What fields should be in struct tmpfile? Should be a struct like following?

I'll explain my idea of the API first, then the content of the struct:

> int tmpfile_init(struct *tmpfile, char *parent); // create a temporary under `parent` path.
> int tmpfile_publish(struct *tmpfile, char *path);   // move the `tmpfile` to `path` 
> int tmpfile_release(struct *tmpfile);     // delete the temporary

The API should encapsulate all these details, ideally it would be something like this:

```c
void tmpfile_init(struct tmpfile *, const char *targetname); // prepare a temporary file with given result filename, which eventually should be created
int tmpfile_publish(struct tmpfile *); // create the file with initially given target name
void tmpfile_release(struct tmpfile *); // create any temporary files
int tmpfile_write(struct tmpfile *, char *bytes, size_t count); // write given amount of bytes into temporary file
int tmpfile_printf(struct tmpfile *, char *fmt, ...); // like fprintf, write formatted string into temporary file
```

This would imply that the whole write operations would be agnostic of the details of tmpfile, i.e. no fd or FILE handling, no caring about temporary file name creation etc.

The struct could be something like this:

```c
struct tmpfile {
    char tmpname[PATH_MAX]; // name of temporary file, if needed (O_TMPFILE etc.)
    char *targetname; // name of target file
    FILE *f; // FILE handle of temporary file
};
```

This means that the struct should keep a FILE * handle of a temporary file (O_TMPFILE or not) and the target name.

wenjiewang-intel:
> > What fields should be in struct tmpfile? Should be a struct like following?
> 
> I'll explain my idea of the API first, then the content of the struct:
> 
> > int tmpfile_init(struct *tmpfile, char *parent); // create a temporary under `parent` path.
> > int tmpfile_publish(struct *tmpfile, char *path);   // move the `tmpfile` to `path`
> > int tmpfile_release(struct *tmpfile);     // delete the temporary
> 
> The API should encapsulate all these details, ideally it would be something like this:
> 
> ```c
> void tmpfile_init(struct tmpfile *, const char *targetname); // prepare a temporary file with given result filename, which eventually should be created
> int tmpfile_publish(struct tmpfile *); // create the file with initially given target name
> void tmpfile_release(struct tmpfile *); // create any temporary files
> int tmpfile_write(struct tmpfile *, char *bytes, size_t count); // write given amount of bytes into temporary file
> int tmpfile_printf(struct tmpfile *, char *fmt, ...); // like fprintf, write formatted string into temporary file
> ```
> 
> This would imply that the whole write operations would be agnostic of the details of tmpfile, i.e. no fd or FILE handling, no caring about temporary file name creation etc.
> 
> The struct could be something like this:
> 
> ```c
> struct tmpfile {
>     char tmpname[PATH_MAX]; // name of temporary file, if needed (O_TMPFILE etc.)
>     char *targetname; // name of target file
>     FILE *f; // FILE handle of temporary file
> };
> ```
> 
> This means that the struct should keep a FILE * handle of a temporary file (O_TMPFILE or not) and the target name.

Thanks Stoeckmann for the detailed response, we rewrite the PR according to your comments.

We don't explicitly use the `tmpfile_write` and `tmpfile_printf` in depmod, because in `depmod_output`, it will output the content either to stdout or temporary file. If we pass the `tmpfile` to the callback functions like `output_deps`, this will cause too many modification, but don't make any functional change.

danhe1:
It is not sufficient to check the Linux version only. We found that the O_TMPFILE flag is supported only on certain filesystems (e.g., tmpfs, ext4 with the user_xattr option enabled), which means that it is possible to fail to create the temp file over the ext4 file system equipped with a compatible Linux version. 

But we agree that creating a temp file with O_TMPFILE is more secure than the solution with mktemp. Because of this, we have the following options.

1. Use mktemp only, which is not that secure but is more compatible.
2.  Check the Linux version, and use O_TMPFILE when it is 3.11 or newer, and use mktemp when it is too old.
3. Try  O_TMPFILE first, and switch to mktemp when we get a failure.

We are also considering the struct tmpfile to hold all the information for a temp file. There are a set of functions that we have defined, including tmpfile_init, tmpfile_publish, tmpfile_release, tpmfile_write and tmpfile_print.

- Someone thinks that we should include all of those functions since it is a fundamental library.
- Someone thinks that we should have the basic functions like init, publish and release, while add the new functions in the future when we really need them.

Both opinions are reasonable, but we should make a decision and select one.

@lucasdemarchi @evelikov @stoeckmann Could you provide any suggestions here, so that we can figure out the best solution?

> Since there is a single use-case with a well defined semantics, I wouldn't bother with adding tmpfile abstraction at this point... personally.
> 
> As we get more users we can see what abstraction makes sense and go there.
> 
> Then again, the key point is to use `O_TMPFILE`, so let's explore where/when it is supported:
> 
> * ext{2,3,4}, udf, minix, tmpfs since 3.11 (circa 2013)
> * xfs since 3.15 (circa 2014)
> * btrfs, f2fs since 3.16 (circa 2014)
> * ubifs since 4.9 (circa 2016)
> * bcachefs (with initial merge) since 6.7 (circa 2024)
> 
> From the remaining (based on 6.14-rc7, excluding pseudo fs'), we have:
> 
> * read-only
>   
>   * by design or implementation) - afs, befs, cramfs, efs, erofs, isofs, qnx*, romfs, squashfs, ufs
>   * or experimental write support - adfs
> * third party, storage only afaict
>   
>   * somewhat archaic (affs/amiga, bfs/SCO Unixware, vxfs/SCO Unixware, jfs/IBM, omfs, orangefs, sysv)
>   * MS ones - _fat, ntfs_
>   * Apple ones - hfs*
> * distributed/cluster, storage only afaict - ceph, coda, gfs2, nfs, ocfs
> * virtual, storage only afaict - p9, vboxsf
> * other - jffs2, nilfs2, zonefs
> 
> From those, the RO and "storage only" ones seem out of scope wrt running depmod.
> 
> Looking at my only jffs2 device (openwrt router) it lacks any modules.* files plus it uses kmod tools from busybox. Not sure about the other two.
> 
> In summary:
> 
> It _seems_ that we could mandate `O_TMPFILE` as soon (_if really_) are happy to have linux v3.11 as depmod runtime requirement. Note: runtime, not input files. In the unlikely case of running unsupported combo (xfs before 3.15 or any of the others), users can fallback to `-o` with a `tmpfs` backed mount point.

gongjun-song:
@evelikov @stoeckmann @lucasdemarchi, any suggestion on this PR now? thanks!

lucasdemarchi:
Sorry for the delay on the PRs. On a quick glance they look fine, just have to find some free time to play with them. I plan to do that by EOW

lucasdemarchi:
> In summary:
> 
> It _seems_ that we could mandate `O_TMPFILE` as soon (_if really_) are happy to have linux v3.11 as depmod runtime requirement. Note: runtime, not input files. In the unlikely case of running unsupported combo (xfs before 3.15 or any of the others), users can fallback to `-o` with a `tmpfs` backed mount poin

I don't think so.. there are more scenarios in which it can fail. One that you listed for example is  in the "distributed case". nfs and friends are still used nowadays to mount the /, particularly on CI systems that don't really want to store anything and just fetches the root over the network when they start testing.

I'd be conservative at this point, write the abstraction and then *later* either swap the implementation or add the faster and more secure alternative.

lucasdemarchi:
As for API, I'm more on the init/publish/release idea. I don't think we need to have wrappers over write - we should rather return an fd that is usable for libc. If a FILE is needed, we can just get one with fdopen(). Or if fwrite and friends is currently used (which seems the case in depmod), we could just return a FILE* to avoid the fdopen().

The init() I'd model it around openat. We don't need to create an absolute path. When we tmpfile_openat() we pass the dir where the file will be created and the template filename to be used in the case a mkstemp backend is used. Something like this would be the complete signature:

```
int tmpfile_openat(int dirfd, const char *path_tmpl, int flags, mode_t mode, struct tmpfile *t);
```

Then we have `tmpfile_publish(t, targetname)` that either materializes the file in the filesystem if O_TMPFILE was used, or do the right mv operation. Other possible name is "tmpfile_link()" if we follow more a fs-type of name. Note that you don't need to store the targetname (nor make it an absolute path). It's only important later when we link the file in the filesystem and for the mkstemp approach you can use renameat().

And finally `tmpfile_close()` that possibly discards the file by closing it and free's internal state tracking.

lucasdemarchi:
Also... sorry for taking so long to review. Looking at the implementation you have here seems reasonable and if you don't have time, I can work on some additional changes .

tools/depmod.c
```diff
@@ -2602,6 +2606,11 @@ static int depmod_output(struct depmod *depmod, FILE *out)
 			CRIT("could not open directory %s: %m\n", dname);
 			return err;
 		}
+		if (get_absolute_path(dfd, &dpath) < 0) {
```
lucasdemarchi, line 2609: as I said, I don't think we need this as we don´t need the absolute path... We can just keep using the *at() variants from libc. If we need a function like this, I'd rather name it fd_lookup_path().

tools/depmod.c
```diff
@@ -2610,53 +2619,42 @@ static int depmod_output(struct depmod *depmod, FILE *out)
...
 			break;
 		}
 
-		if (renameat(dfd, tmp, dfd, itr->name) != 0) {
-			err = -errno;
-			CRIT("renameat(%s, %s, %s, %s): %m\n", dname, tmp, dname,
-			     itr->name);
+		err = tmpfile_publish(&file);
+		if (err != 0) {
+			CRIT("publish temporary from %s to %s: %m\n", file.tmpname,
```
lucasdemarchi, line 2656: I don't thinks we can use %m anymore. If the function returned -EINVAL, that's what the should tell the user, not a possible old errno set by libc (note that in your implementation there isn't even a libc call at that point, so it would probably read "SUCCESS" in the error message.

lucasdemarchi (changes requested):

tools/depmod.c
wenjiewang-intel: will fix.

lucasdemarchi:
> Generally, `mkstemp` needs an absolute path to create a temporary file. If we don't want to touch the absolute path, maybe we need to use another method like my first commit version. We can manually replace the XXXXXX with a random number to generate a random filename and not use `mkstemp` anymore.
> first version: https://github.com/gongjun-song/kmod/compare/2f5f67e689cd374e61b9a3ecb5e3d207e35bdbd0..4ebc1b583a871fcf8bac8c0c6064816e872dc2e8

if you replace all that code to generate and check for a random file with the equivalent of what we had in the codebase, it should be fine:

```
			n = snprintf(tmp, sizeof(tmp), "%s.%i.%lli.%lli", itr->name,
				     getpid(), (long long)tv.tv_usec,
				     (long long)tv.tv_sec);
```

i.e. use the pid + timestamp. Because then you don't have a change in behavior, it's just what was already there, called from a different place.

Other possibilities that still use mkstemp:

1) chdir() before calling it, chdir after calling it. Ugly, but should work.
2) still create the absolute path, but don't do that in depmod.c. Do it in the function creating the tmpfile. Once we implement the O_TMPFILE part, it will likely be in a path that is not used anymore (if the fs requirements are met).

shared/tmpfile-util.c
```diff
@@ -0,0 +1,102 @@
...
+	if (err < 0) {
+		goto create_fail;
+	}
+
+	n = snprintf(tmpfile_path, PATH_MAX, "%s/%s", targetdir, tmpname_tmpl);
+	if (n < 0 || n >= PATH_MAX) {
+		goto create_fail;
+	}
+
+	fd = mkstemp(tmpfile_path);
```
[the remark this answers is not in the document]
lucasdemarchi: @evelikov  we already define _GNU_SOURCE, which implies:

```
_POSIX_C_SOURCE with the value 200809L with glibc >= 2.10
```

is this < 2.10 or still relevant? Do you know the behavior for e.g. musl?

shared/util.c
```diff
@@ -476,6 +476,31 @@ int mkdir_parents(const char *path, mode_t mode)
...
 }
 
+int fd_lookup_path(int fd, char **ret_path)
+{
+	char proc_path[PATH_MAX];
+	char fd_path[PATH_MAX];
+	ssize_t len;
+
+	len = snprintf(proc_path, sizeof(proc_path), "/proc/self/fd/%d", fd);
+	if (len < 0 || len >= (ssize_t)sizeof(proc_path)) {
```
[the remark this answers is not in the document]
lucasdemarchi: we have some people building the kernel on MacOS. I wonder where they are getting depmod from. We never really cared about other OSes.

shared/tmpfile-util.c
```diff
@@ -0,0 +1,102 @@
+
```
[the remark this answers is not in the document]
wenjiewang-intel: fixed.

shared/tmpfile-util.c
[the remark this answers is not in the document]
wenjiewang-intel: Should I add some other check?

shared/tmpfile-util.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
```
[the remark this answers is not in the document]
wenjiewang-intel: fixed.

shared/util.c
```diff
@@ -476,6 +476,31 @@ int mkdir_parents(const char *path, mode_t mode)
 	return mkdir_p(path, end - path, mode);
 }
 
+int fd_lookup_path(int fd, char **ret_path)
```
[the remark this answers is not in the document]
wenjiewang-intel: fixed.

shared/util.c
```diff
@@ -476,6 +476,31 @@ int mkdir_parents(const char *path, mode_t mode)
...
+	char proc_path[PATH_MAX];
+	char fd_path[PATH_MAX];
+	ssize_t len;
+
+	len = snprintf(proc_path, sizeof(proc_path), "/proc/self/fd/%d", fd);
+	if (len < 0 || len >= (ssize_t)sizeof(proc_path)) {
+		return -errno;
+	}
+
+	len = readlink(proc_path, fd_path, sizeof(fd_path) - 1);
```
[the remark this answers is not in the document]
wenjiewang-intel: added.

tools/depmod.c
```diff
@@ -2606,34 +2607,16 @@ static int depmod_output(struct depmod *depmod, FILE *out)
...
-			}
-			fd = openat(dfd, tmp, flags, mode);
-			if (fd < 0) {
-				ERR("openat(%s, %s, %o, %o): %m\n", dname, tmp, flags,
-				    mode);
-				continue;
-			}
-			fp = fdopen(fd, "wb");
+			const char *tmpname_tmpl = "tmpfileXXXXXX";
+			mode_t mode = 0644;
```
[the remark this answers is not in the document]
wenjiewang-intel: fixed, move tmpname_tmpl in `tmpfile_openat`.

lucasdemarchi:
I made a few changes:
- squashed the patches since they are doing fixes on top
- Amended a few style changes
- Replaced memset() + memcpy() with memcpy(..., len + 1)  when copying string. As an improvement on top we could probably drop some string copies and just use strbuf for tmpname.
- when releasing/publishing it's better to set fd to -1 - we don't want to inadvertently close fd 0, which is a valid fd

Good job, thanks !

PR #311, opened 2025-03-07
Added the test case of using multiple softdep for a module

gongjun-song:
[Issue: 33] When using multiple softdep for a module, only one of them have effects

We defined two modules mod-softdep-a and mod-softdep-b. The softdep statements are split into two separate modprobe configure files.

The unit test case is supposed to fail, since the libkmod will take only the first softdep statement based on the current implementation.

Signed-off-by: Dan He <dan.h.he@intel.com>
Signed-off-by: Gongjun Song <gongjun.song@intel.com>
Signed-off-by: Yuchi Chen <yuchi.chen@intel.com>
Signed-off-by: Wenjie Wang <wenjie2.wang@intel.com>
Signed-off-by: Qingqing Li <qingqing.li@intel.com>


testsuite/rootfs-pristine/test-softdep-mult/etc/modprobe.d/dumb-softdep.conf
```diff
@@ -0,0 +1,4 @@
+softdep	mod-softdep-a	pre: mod-foo-a
+softdep	mod-softdep-a	pre: mod-foo-b
+softdep	mod-softdep-a	post: mode-foo-c
+softdep	mod-softdep-b	pre: mod-foo-a
```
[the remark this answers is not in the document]
gongjun-song: This is indeed intentional. our test case is designed to cover the following situations:
    1. All the dependencies are included in one single file.
    2.The dependencies are scattered in separate files.
    3.The (pre or post) dependencies are claimed in the same line.
    4.The dependencies are claimed in multiple lines.

gongjun-song:
> Some minor/cosmetic changes suggested. The biggest one being the wrong (?) copyright notice.

Thanks, I have updated copyright, please help review.

gongjun-song:
@evelikov @lucasdemarchi, any further issues before merging?

testsuite/rootfs-pristine/test-multi-softdep/lib/modules/4.4.4/modules.softdep
```diff
@@ -0,0 +1 @@
+# Soft dependencies extracted from modules themselves.
```
lucasdemarchi, line 1: this is weird even for the failure. The test modules have
MODULE_SOFTDEP(...)

Why in the end this file is empty?

testsuite/test-multi-softdep.c
```diff
@@ -0,0 +1,127 @@
...
+	kmod_module_unref_list(pre);
+	kmod_module_unref_list(post);
+	kmod_module_unref(mod);
+	kmod_unref(ctx);
+	return EXIT_FAILURE;
+}
+
+/*
+ * FIXME: This test indicates a bug in kmod. Once fixed, the expected_fail should be removed.
+ * See https://github.com/kmod-project/kmod/issues/33 for more details.
```
lucasdemarchi, line 119: It's not very clear to me this is being handled correctly because you have a single "test dir" that mixes the softdeps coming from the kernel modules and coming from the configuration. Example:

From these files:
testsuite/rootfs-pristine/test-multi-softdep/etc/modprobe.d/dumb-softdep.conf
testsuite/module-playground/mod-softdep-a.c

If we consider that softdep should append, we'd expect a load order chain like:

mod-foo-a -> mod-foo-b -> mod-sofdep-a -> mod-foo-c

But this  info is repeated both in the kernel module and in the conf. Shouldn't we have separate tests (e.g. just having a separate mod-softdep-c or the like should be sufficient) to test the behavior when coming direct from the configuration or from the index file that depmod generates? By index you added here, it seems depmod is at fault too, or the index is not up-to-date.

testsuite/rootfs-pristine/test-multi-softdep/lib/modules/4.4.4/modules.softdep
danhe1: I removed the contents in modules.softdep to avoid depmod from reading the info there. The soft dependencies generated by depmod in modules.softdep are all included in single lines, i.e.

```
mod-softdep-a: pre: mod-foo-a mod-foo-b post: mod-foo-c
mod-softdep-a: pre: mod-foo-a post: mod-foo-b mod-foo-c

```
Since our target is to test the configurations in modprobe.d, we do not want our test script to get the info above.

testsuite/test-multi-softdep.c
danhe1: The modules mod-softdep-a and mod-softdep-b are defined so that the function kmod_module_get_softdeps(mod, &pre, &post); can work correctly. Without those two modules, this function will return empty pre and post lists, which is not what we expect. After we defined those two modules:

For mod-softdep-a, the pre list is mod-foo-a, and the post list is null.
For mod-softdep-b, the pre list is mod-foo-a, and the post list is null.

Which is correct based on the current implementation of libkmod, since only the first line of softdep is active.

lucasdemarchi:
Applied, thanks.

PR #313, opened 2025-03-11
tools: use space for long opts

evelikov:
The getopt_long() API handles both space and = separated long options just fine. On the other hand, supporting both in the shell completions means that we'll have to quadruple them in terms of size/complexity.

So let's try to keep things simple and consistent, by recommending only the space variant.

Make sure to update both the tools' --help and the respective man pages.

### DIFF
--- a/tools/modprobe.c
+++ b/tools/modprobe.c
@@ -102,7 +102,7 @@ static void help(void)
 	       "\t-r, --remove                Remove modules instead of inserting\n"
 	       "\t    --remove-dependencies   Deprecated: use --remove-holders\n"
 	       "\t    --remove-holders        Also remove module holders (use together with -r)\n"
-	       "\t-w, --wait=<MSEC>           When removing a module, wait up to MSEC for\n"
+	       "\t-w, --wait <MSEC>           When removing a module, wait up to MSEC for\n"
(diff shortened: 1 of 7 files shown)


lucasdemarchi:
This doesn't match what is more commonly seen on man pages for some tools.  Both systemd and git use the "=" in their man page.. It matches python's argparse and  rust's clap for --help though.

tools/modprobe.c
```diff
@@ -102,7 +102,7 @@ static void help(void)
 	       "\t-r, --remove                Remove modules instead of inserting\n"
 	       "\t    --remove-dependencies   Deprecated: use --remove-holders\n"
 	       "\t    --remove-holders        Also remove module holders (use together with -r)\n"
-	       "\t-w, --wait=<MSEC>           When removing a module, wait up to MSEC for\n"
```
lucasdemarchi, line 105: on being consistent, mind changing this s/<MSEC>/MSEC/ with a commit on top?

lucasdemarchi (approved):

tools/modprobe.c
lucasdemarchi: that can be done in a new PR, no need to keep this pending.

lucasdemarchi:
Applied, thanks.

PR #314, opened 2025-03-11
ci: remove Debian/pahole dependency

evelikov:


.github/actions/setup-debian/action.yml
```diff
@@ -22,6 +22,7 @@ runs:
           echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list
           backports_pkgs=("meson" "ninja-build")
         else
+          # The pahole dependency should really be resolved via linux-headers-generic.
```
lucasdemarchi, line 25: The background for this is:

- Ubuntu started needing pahole to build their kernel (for BTF handling afair)
- It's not easily installable in debian bullseye and not actually needed - their kernel config require it

```
$ podman run -v $PWD:/src -w /src -it docker.io/library/debian:bullseye-slim
# apt update && apt install pahole
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
E: Unable to locate package pahole

# # this is anyway needed to get a working meson version
# echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list
# apt update && install pahole
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 pahole : Depends: libbpf0 (>= 1:0.4.0) but 1:0.3-2 is to be installed
E: Unable to correct problems, you have held broken packages.
```

I don't think it matches this comment you're adding.

lucasdemarchi: > The comment refers to the _non_ bullseye section of the if block.

ah ok.

> I suspect you meant Debian here, right?

yes

> AFAIU the use of pahole is not specific to any distribution. It's a decision by the kernel devs/build system, where those files are bundled in the `-headers` package.

it depends on the kernel config, which is distro-dependent. If the comment is a "TODO/FIXME" for distros to add pahole as a dependency of the linux-headers packages like Archlinux has, then ok. But the way it's worded it seems more like it's something **we** should be doing in this action.

lucasdemarchi, line 25:
````suggestion
# The pahole dependency should rather be a dependency in linux-headers-generic
# if the kernel config requires that rather than having to manually install it
```

lucasdemarchi: @rfc1036 maybe something you could do?

rfc1036: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1098706 ?

lucasdemarchi: Yes, looks like it. Thanks!

From that thread it seems to be fixed in linux-headers-6.12.17... I think that we can already remove this rather than documenting it.

lucasdemarchi: Just tested it in a debian:unstable container and it works without installing pahole explicitly.

lucasdemarchi:
Applied, thanks.

PR #318, opened 2025-03-11
ci: remove no longer needed libtool

evelikov:
We've dropped libtool across all distros but Fedora. The tool is not needed anymore, so let's remove it.

Fixes: 04520dc9 ("ci: Remove autotools dependencies from containers")


lucasdemarchi:
Applied, thanks.

PR #324, opened 2025-03-21
libkmod: fix buffer-overflow in weakdep_to_char

jspcki:
modprobe -c with any sample weakdep command in modprobe.d will overflow memcpy due to size calculation not incorporating trailing null byte, for example:

==462449==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x50200000067b at pc 0x7a83fe6faf59 bp 0x7ffdf6c25060 sp 0x7ffdf6c24808 WRITE of size 12 at 0x50200000067b thread T0
    0 0x7b687e6faf58 in memcpy /usr/src/debug/gcc/gcc/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115
    1 0x5e22b821235f in weakdep_to_char ../libkmod/libkmod-config.c:623
    2 0x5e22b821235f in weakdep_get_plain_weakdep ../libkmod/libkmod-config.c:1166
    3 0x5e22b821d049 in kmod_config_iter_get_value ../libkmod/libkmod-config.c:1317
    4 0x5e22b81fd5ec in show_config ../tools/modprobe.c:187
    5 0x5e22b81fd5ec in do_modprobe ../tools/modprobe.c:946
    6 0x7b687d635487  (/usr/lib/libc.so.6+0x27487) (BuildId: 0b707b217b15b106c25fe51df3724b25848310c0)
    7 0x7b687d63554b in __libc_start_main (/usr/lib/libc.so.6+0x2754b) (BuildId: 0b707b217b15b106c25fe51df3724b25848310c0)
    8 0x5e22b81da114 in _start ($PWD/kmod+0xaa114) (BuildId: c36444aefc2ca73423765d4ebf017a24e55017ee)

May also appear as:

*** buffer overflow detected ***: terminated
Aborted (core dumped)

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -581,7 +581,7 @@ static char *weakdep_to_char(struct kmod_weakdep *dep)
 	/* Rely on the fact that dep->weak[] is a strv that points to a contiguous buffer */
 	if (dep->n_weak > 0) {
 		start = dep->weak[0];
-		end = dep->weak[dep->n_weak - 1] + strlen(dep->weak[dep->n_weak - 1]);
+		end = dep->weak[dep->n_weak - 1] + strlen(dep->weak[dep->n_weak - 1]) + 1;
 		sz = end - start;
 	} else
 		sz = 0;
@@ -594,8 +594,8 @@ static char *weakdep_to_char(struct kmod_weakdep *dep)
 		char *p;
 
 		/* include last '\0' */
-		memcpy(itr, dep->weak[0], sz + 1);
-		for (p = itr; p < itr + sz; p++) {
+		memcpy(itr, dep->weak[0], sz);
+		for (p = itr; p < itr + sz - 1; p++) {
 			if (*p == '\0')
 				*p = ' ';
 		}
--- a/testsuite/rootfs-pristine/test-weakdep/modprobe-c.txt
+++ b/testsuite/rootfs-pristine/test-weakdep/modprobe-c.txt
@@ -1,0 +1,7 @@
+weakdep mod_loop_b mod-loop-a mod-simple
+weakdep mod_weakdep mod-simple
+
+# End of configuration files. Dumping indexes now:
+
+alias symbol:printA mod_loop_a
+alias symbol:printB mod_loop_b
--- a/testsuite/test-weakdep.c
+++ b/testsuite/test-weakdep.c
@@ -96,7 +96,6 @@ static noreturn int modprobe_config(const struct test *t)
 	exit(EXIT_FAILURE);
 }
 DEFINE_TEST(modprobe_config,
-	.expected_fail = true,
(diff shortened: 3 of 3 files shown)


jspcki:
The "+ 1" can be moved to any appropriate other location. I chose this, because:

* It's next to **strlen**(3) which is known to exclude trailing null byte.
* My C++ brain thinks end should be out of the bounds (although in this case it's not really out).

libkmod/libkmod-config.c
```diff
@@ -581,7 +581,7 @@ static char *weakdep_to_char(struct kmod_weakdep *dep)
 	/* Rely on the fact that dep->weak[] is a strv that points to a contiguous buffer */
 	if (dep->n_weak > 0) {
 		start = dep->weak[0];
-		end = dep->weak[dep->n_weak - 1] + strlen(dep->weak[dep->n_weak - 1]);
+		end = dep->weak[dep->n_weak - 1] + strlen(dep->weak[dep->n_weak - 1]) + 1;
 		sz = end - start;
 	} else
 		sz = 0;
```
stoeckmann, line 587: The else-case has an issue as well. If `dep->n_weak` is 0 (precondition to enter the else-block), then `sz` will be 0 as well and `malloc(0)` will basically on all systems return something that is not NULL. Since `itr = s = malloc(0)` is called and `*itr = '\0'` is called at the end of the function, the `sz` allocation itself has to be 1 in the else-block.

In summary, the malloc argument itself could be adjusted. And maybe it's even better to use the strbuf API here as well to get rid of this type of mistake entirely.

stoeckmann:
Oh and by the way: Nice finding and thanks for the report! Do you have example input which triggers the issue for easier replication?

libkmod/libkmod-config.c
jspcki: That's a good point about implementation defined behaviour in malloc.

It seems like the `weakdep_to_char` here was somewhat based on `softdep_to_char` just above this function, with exception where it has `size_t sz = 1` and later accumulates it from `sz_pre` and `sz_post`. This would explain most of the assumptions about the code past the `malloc`.

I'll ensure at least one byte available, but I'll keep returning NULL in cases where `sz` was previously 0 to keep the behaviour consistent for platforms that return NULL for `malloc(0)`.

jspcki:
Any. With `kmod 34.1-1` on Arch Linux:
```
$ modprobe -c >/dev/null || echo $?
# echo weakdep ice dummy >/etc/modprobe.d/temp.conf
$ modprobe -c >/dev/null || echo $?
*** buffer overflow detected ***: terminated
Aborted (core dumped)
134
```

lucasdemarchi (approved):

lucasdemarchi:
I added a test to our testsuite to make sure we don't regress that. I will merge it once we have CI back.

lucasdemarchi:
thanks for fix and reviews

PR #326, opened 2025-03-26
README: remove autotools instruction

stoeckmann:
The autotools support is gone. Remove it from README.


lucasdemarchi:
Applied, thanks

PR #329, opened 2025-03-30
`ln -r` is not POSIX

firasuke:
Hey there, thanks for your time and effort!

This makes `ln` POSIX compliant, because not all `ln` implementations support `-r`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -513,7 +513,7 @@ if get_option('tools')
   mkdir_p = 'mkdir -p "$DESTDIR@0@"'
   meson.add_install_script('sh', '-c', mkdir_p.format(sbindir))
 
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
> This makes `ln` POSIX compliant, because not all `ln` implementations support `-r`.

I think it would be more useful to add the feature in the `ln` implementation you are using to understand relative symlink. We need the link to be relative, not absolute.

We could implement it in kmod as it's easy, but it doesn't improve the overall ecosystem if each project has to workaround missing features in each lower level command/lib. If it was an obscure option supported by just one implementation, it'd be ok. However this is available almost everywhere. It's even available in toybox since commit 49feb50f22e2 ("Add ln -r and relative_path() to lib, plus test.")

firasuke:
> We need the link to be relative, not absolute.

That answers my question. Closing...

PR #332, opened 2025-04-01
modprobe: s|<MSEC>|MSEC|

evelikov:
Drop the brackets to stay consistent with quoting of other arguments - see --config and friends further down.


lucasdemarchi:
Applied, thanks.

PR #334, opened 2025-04-01
Drop leftover makefiles

lucasdemarchi:
I don't think this was used before (at least I always build from the git rootdir), and now that autotools is removed, they are wrong.


lucasdemarchi:
Applied, thanks.

PR #336, opened 2025-04-05
Introduce KMOD_STRERROR to ensure usage is safer.

crrodriguez:
While recent strerror implementations are thread aware/safe, old ones are not, and the issue of shared library safety remains.

That's why strerrordesc_np() was added that is thread safe, async-signal-safe (does not call malloc) and there is no data race with other functions.

Use that if available, or %m ..else fallback to the closest that is calling strerror_r with a buffer whose scope expires inmediately.

This concerns only libkmod, in the future the toolchain may implement per-dso allocations to make the problem disappear in practice.

Note that strerrordesc_np does not display error in the current locale but only in english, this is an small price to pay for safety.

### DIFF
--- a/libkmod/libkmod-builtin.c
+++ b/libkmod/libkmod-builtin.c
@@ -90,7 +90,7 @@ static ssize_t get_strings(struct kmod_builtin_info *info, const char *modname,
 		if (n == -1) {
 			if (!feof(info->fp)) {
 				count = -errno;
-				ERR(info->ctx, "get_string: %s\n", strerror(errno));
+				ERR(info->ctx, "get_string: %m\n");
 			}
 			break;
 		}
@@ -182,7 +182,7 @@ ssize_t kmod_builtin_get_modinfo(struct kmod_ctx *ctx, const char *modname,
 		*modinfo = strbuf_to_vector(&buf, (size_t)count);
 		if (*modinfo == NULL) {
 			count = -errno;
-			ERR(ctx, "strbuf_to_vector: %s\n", strerror(errno));
+			ERR(ctx, "strbuf_to_vector: %m\n");
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


crrodriguez:
No, strerror is thread safe now, it is not **library** safe though. The buffer is thread local might be invalidated by whatever caller libkmod has no control of, This needs per-dso storage to be really safe but it has not (yet) been implemented.

crrodriguez:
https://sourceware.org/glibc/manual/latest/html_mono/libc.html#Error-Messages-1

"As there is no way to restore the previous state after calling strerror, library code should not call this function because it may interfere with application use of strerror, invalidating the string pointer before the application is done using it. Instead, strerror_r, snprintf with the ‘%m’ or ‘%#m’ specifiers, strerrorname_np, or strerrordesc_np can be used instead."

lucasdemarchi:
IMO the changes to %m are good and could be done without any other changes.

Introducing a KMOD_STRERROR() is very ugly. strerror_r()  would be good, but then there's the compat nightmare to care about. Honestly any application doing something like this is already broken:

```
s = strerror(...)
kmod_xyz()
printf("%s\n",  s);
```
The manpage for strerror() already mentions it:

```
This string must not be modified by the application, and the returned pointer will be
invalidated on a subsequent call to strerror() or strerror_l(), or if the thread that
obtained the string exits.
```

Since the caller and libkmod are sharing libc, the caller has no idea if libkmod is going to use that libc function. And if we didn't, some other library used by kmod could still screw it. I don't think there's any hope for an application to get it right other than stop mixing those calls.

If we do decide to use the safe implementation, then just defining STRERROR() without the  KMOD_ prefix would be better. Then we can probably copy the relevant part in `.github/codeql-queries/PotentiallyDangerousFunction.ql` from systemd.

lucasdemarchi:
Closing since we already converted what we could to %m and not pursuing the rest. Thanks.

PR #341, opened 2025-04-26
testsuite: fix dirty rootfs error message

whythefyouflying:
Running a test binary directly on a dirty rootfs results in the following error message:

        ERR: rootfs is dirty, please run
	'meson compile testsuite/stamp-rootfs' before running this test

Running the command specified by the error message fails:

        $ meson compile testsuite/stamp-rootfs
        ERROR: Can't invoke target `testsuite/stamp-rootfs`:
	target not found

Fix the error message to display the proper target.

Fixes: b85b2a0 ("testsuite: update "make rootfs" error message")


lucasdemarchi:
Applied, thanks.

PR #345, opened 2025-05-07
Some tmpfile inspired changes

evelikov:
As the tmpfile helpers were introduced I've noticed we have a mix of `PATH_MAX` vs `sizeof()` handling across the project. The first two commits address most instances - the remaining ones might need more careful surgery.

The last two commits simplify the helpers wrt LoC and memory usage.

### DIFF
--- a/shared/util.h
+++ b/shared/util.h
@@ -54,7 +54,7 @@ static inline _must_check_ _nonnull_all_ bool path_is_absolute(const char *p)
 int mkdir_p(const char *path, int len, mode_t mode);
 int mkdir_parents(const char *path, mode_t mode);
 unsigned long long stat_mstamp(const struct stat *st);
-char *fd_lookup_path(int fd);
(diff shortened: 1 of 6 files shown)


shared/util.h
```diff
@@ -54,7 +54,7 @@ static inline _must_check_ _nonnull_all_ bool path_is_absolute(const char *p)
 int mkdir_p(const char *path, int len, mode_t mode);
 int mkdir_parents(const char *path, mode_t mode);
 unsigned long long stat_mstamp(const struct stat *st);
-char *fd_lookup_path(int fd);
+_nonnull_all_ int fd_lookup_path(int fd, size_t pathlen, char *path);
```
lucasdemarchi, line 57: even if it's an "out arg", probably better to use `path, pathlen` order like memcpy(), strcpy(), etc.

lucasdemarchi (approved):
Just a nit on order of args in the last patch, other lgtm. Thanks.

lucasdemarchi:
Applied, thanks.

Issue #244, opened 2024-11-15
Audit errno (ab)uses


evelikov:
As mentioned in https://github.com/kmod-project/kmod/pull/240 there are places were we set errno only to never check it. Glancing around we also have instances that set it, where it can be returned directly via the function.

So let try to:
 - list all instances which set errno
 - remove instances which are not used by libkmod nor libc
 - use a simpler way to propagate the error, if available

PR #346, opened 2025-05-10
Reduce strerror and errno usage

evelikov:
Originally inspired by https://github.com/kmod-project/kmod/issues/244, with some commit (fragments) from https://github.com/kmod-project/kmod/pull/336 and https://github.com/kmod-project/kmod/pull/343

Tl:Dr: we no longer manually set `errno` but propagate the error code manually, iff needed. In addition most (but not all) of our `strerror()` usage is replaced by `%m`.

As a nice side-effect the size of libkmod.so (should also be applicable for the tools) has decreased:

 116099    2224       8  118331   1ce3b before//libkmod.so
 115685    2224       8  117917   1cc9d after/libkmod.so

NOTE: the PR also includes a few related bug fixes of varying severity. Happy to flesh them out into separate PR/PRs if that helps.

### DIFF
--- a/libkmod/libkmod-builtin.c
+++ b/libkmod/libkmod-builtin.c
@@ -90,7 +90,7 @@ static ssize_t get_strings(struct kmod_builtin_info *info, const char *modname,
 		if (n == -1) {
 			if (!feof(info->fp)) {
 				count = -errno;
-				ERR(info->ctx, "get_string: %s\n", strerror(errno));
+				ERR(info->ctx, "get_string: %m\n");
 			}
 			break;
 		}
@@ -182,7 +182,7 @@ ssize_t kmod_builtin_get_modinfo(struct kmod_ctx *ctx, const char *modname,
 		*modinfo = strbuf_to_vector(&buf, (size_t)count);
 		if (*modinfo == NULL) {
 			count = -errno;
-			ERR(ctx, "strbuf_to_vector: %s\n", strerror(errno));
+			ERR(ctx, "strbuf_to_vector: %m\n");
 		}
 	}
 
--- a/libkmod/libkmod-builtin.c
+++ b/libkmod/libkmod-builtin.c
@@ -62,19 +62,6 @@ static void kmod_builtin_info_release(struct kmod_builtin_info *info)
 	fclose(info->fp);
 }
 
-static ssize_t get_string(struct kmod_builtin_info *info)
-{
-	ssize_t len;
-
-	len = getdelim(&info->buf, &info->bufsz, '\0', info->fp);
-	if (len > 0 && info->buf[len] != '\0') {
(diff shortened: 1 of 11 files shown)


libkmod/libkmod-builtin.c
```diff
@@ -33,29 +33,27 @@ struct kmod_builtin_info {
...
 {
 	char path[PATH_MAX];
 	FILE *fp = NULL;
 	const char *dirname = kmod_get_dirname(ctx);
 	size_t len = strlen(dirname);
 
-	if ((len + 1 + strlen(MODULES_BUILTIN_MODINFO) + 1) >= sizeof(path)) {
-		errno = ENAMETOOLONG;
+	if ((len + 1 + strlen(MODULES_BUILTIN_MODINFO) + 1) >= sizeof(path))
 		return false;
```
stoeckmann, line 44: Since return value is changed from bool to int, this should be `return -ENAMETOOLONG;`

testsuite/init_module.c
```diff
@@ -234,17 +234,24 @@ long init_module(void *mem, unsigned long len, const char *args)
...
+	}
 
 	err = kmod_elf_get_section(elf, ".gnu.linkonce.this_module", &off, &bufsize);
 	buf = (const char *)kmod_elf_get_memory(elf) + off;
 	kmod_elf_unref(elf);
 
 	/* We couldn't parse the ELF file. Just exit as if it was successful */
 	if (err < 0)
-		return 0;
+		return -1;
```
stoeckmann, line 254: If we really want to return an error now (which I don't mind), the comment has to be adjusted or removed.

stoeckmann (changes requested):

libkmod/libkmod-file.c
```diff
@@ -58,57 +58,54 @@ static const struct comp_type {
...
 	file = calloc(1, sizeof(struct kmod_file));
 	if (file == NULL)
-		return NULL;
+		return -ENOMEM;
 
 	file->fd = open(filename, O_RDONLY | O_CLOEXEC);
 	if (file->fd < 0) {
 		free(file);
-		return NULL;
+		return -ENOMEM;
```
lucasdemarchi, line 96: ENOMEM here doesn't feel right. However using the errno from open() after the free() is also invalid. Options:  a) add a local saved_errno in the error path so you can return it; b) annotate file with _cleanup_free_, and assign it to NULL after setting *out_file.  Approach (b) would allow to remove one more free() call in the error path.

lucasdemarchi: > Not sure I follow - why is it invalid? The free() call should never alter errno... does it?

humn.. true

> Either way, option b) applied. Thanks for the review.

thanks

lucasdemarchi:
Applied, thanks.

PR #348, opened 2025-05-16
libkmod: Fix dependency count in kmod_module_parse_depline()

ytcoode:
The variable n is intended to track the number of dependencies for mod. However, it is only initialized and not incremented during dependency parsing.

Fix it.


lucasdemarchi:
Fixes: be728dd34718 ("libkmod: Simplify module dependency parsing")


Applied, thanks.

stoeckmann:
Thanks for spotting and fixing it!

PR #353, opened 2025-05-19
Modprobe from any for --show-{modversions,exports}

evelikov:
In the odd case of using --show/dump-modversions or --show-exports, I regularly get irked that they take only the full module path.

This PR fixes that, admittedly with zero testing, man/shell-completion updates... Looking at which, it seems like we never added `--show-exports` into the man page - I'm guessing that was not intentional?

If people are on board with the idea I will polish and resubmit. As always - name suggestions (see the helper functions within) are greatly appreciated.

### DIFF
--- a/tools/modprobe.c
+++ b/tools/modprobe.c
@@ -221,20 +221,15 @@ static int module_new_from_any(struct kmod_ctx *ctx, const char *module,
 	return 0;
 }
 
-static int show_modversions(struct kmod_ctx *ctx, const char *filename)
+static int print_modversions(const struct kmod_module *mod)
 {
 	struct kmod_list *l, *list = NULL;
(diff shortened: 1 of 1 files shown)


tools/modprobe.c
```diff
@@ -273,10 +265,32 @@ static int show_exports(struct kmod_ctx *ctx, const char *filename)
...
+	} else {
+		kmod_list_foreach(l, list) {
+			mod = kmod_module_get_module(l);
+			err = print_exports(mod);
+			kmod_module_unref(mod);
+		}
+		kmod_module_unref_list(list);
+	}
+	return err;
+
```
lucasdemarchi, line 307: something wrong here...

PR #354, opened 2025-05-24
Make use of streq/strstartswith and strlen(string-literal)

evelikov:
Does exactly what it says on the tin.

Any `strncmp()` and `memcmp()` instances using a non-literal have not be converted to `strstartswith()` to avoid introducing (albeit negligible) perf changes.


lucasdemarchi:
Applied, thanks.

PR #355, opened 2025-05-24
Test wrapper nitpicks

evelikov:
While looking at re-enabling the tests on 32bit Arch a few nits picks stood up. Namely the 3 LD_PRELOAD libraries are fairly inconsistent around their dlsym handling.

This series resolves that, starting off with the removal of non RTLD_NEXT aware platforms. Our targets - glibc, musl and bionic support it.

If other C runtimes are found to need this we can reintroduce it alongside a CI action to keep things in check.


lucasdemarchi:
Applied, thanks.

PR #356, opened 2025-05-24
Remove ifdef's and other autotools w/a

evelikov:
~~As the title suggests this is a bit of a mix and match PR, where one commit led to the other and ... Let's just say it grew organically, into this unexpected shape.~~

~~See individual commits for details and let me know if it helps to split things into separate PRs.~~

Edit: remove unrelated commits, keep only autotools related workarounds

For a while we've been converging on `if HAVE` - both at pre-coprocessor and compiler level. There were a few instances remaining, required by the autotools macros/helpers that we used. With autotools gone, we can convert those.

While here, we can also remove another autotools workaround - support for in-tree builds. Meson mandates out-of-tree builds.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -27,8 +27,6 @@ cc = meson.get_compiler('c')
 # We rely on the glibc variant of basename, et al.
 cdata.set10('_GNU_SOURCE', true)
 
-# TODO: Once meson-only, adjust all #ifdef X to if X and convert .set to .set10
-
 ################################################################################
 # Function and structure checks
 ################################################################################
@@ -38,9 +36,7 @@ _funcs = [
   'secure_getenv',
 ]
 foreach func : _funcs
-  if cc.has_function(func, args : '-D_GNU_SOURCE')
-    cdata.set('HAVE_@0@'.format(func.to_upper()), true)
-  endif
+  cdata.set10('HAVE_@0@'.format(func.to_upper()), cc.has_function(func, args : '-D_GNU_SOURCE'))
 endforeach
 
 # Meson has some amount of support for finding builtins by passing the symbol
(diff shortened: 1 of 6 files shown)


stoeckmann:
The changes itself look good, but I dislike this mixture in a single PR. Especially dropping dietlibc support in such a mixture.

It's not like I care too much about specific libraries out there, but if you break support for whatever systems out there and it's such a "let's do various things in a single PR with a funny overly long title" which is linked and discussed, it's a recipe for bad spotlight on kmod which it simply doesn't deserve, considering its past development structure.

Same goes for C99. It's important enough as a decision to have its own PR, although I doubt that anyone would mind by this date and time.

stoeckmann (approved):

meson.build
```diff
@@ -457,6 +459,7 @@ kmod_sources = files(
 kmod = executable(
   'kmod',
   kmod_sources,
+  c_args : '-DHAVE_FINIT_MODULE=0',
```
lucasdemarchi, line 462: we need at least a better commit message. It doesn't really make sense to have HAVE_FINIT_MODULE=0 hardcoded. In that case why don't we **drop** it and always use the call via syscall()?

scripts/setup-modules.sh
```diff
@@ -3,14 +3,10 @@
...
-MODULE_PLAYGROUND=$3
+MODULE_PLAYGROUND=$2
 
-# TODO: meson allows only out of tree builds
-if test "$SRCDIR" != "$BUILDDIR"; then
-    mkdir -p "$MODULE_PLAYGROUND"
-    cp --archive "$SRCDIR/$MODULE_PLAYGROUND/"* "$MODULE_PLAYGROUND/"
-fi
+mkdir -p "$MODULE_PLAYGROUND"
+cp --archive "$SRCDIR/$MODULE_PLAYGROUND/"* "$MODULE_PLAYGROUND/"
```
lucasdemarchi, line 9: this relies on $MODULE_PLAYGROUND being relative and $CWD being the builddir... Humn.. we could have dropped the $BUILDDIR arg long ago and just check $SRCDIR vs $CWD?

PR #358, opened 2025-05-29
Annotate variables as static et al

evelikov:
This series annotates a handful of variables as static and removes the unnecessary default initializer. It also includes a couple of benign semi-related changes.

As result the binaries are ~3k smaller:

```
1314006  746968    1168 2062142  1f773e before/kmod
1312318  745456    1160 2058934  1f6ab6 after/kmod

 850358  485152      16 1335526  1460e6 before/libkmod.so
 848694  483640       8 1332342  145476 after/libkmod.so
```


lucasdemarchi:
Applied, thanks.

PR #359, opened 2025-05-29
Standardize compiler/C runtime requirements, use nsec granularity in depmod

evelikov:
This series writes the (unwritten) minimum requirements wrt compiler and C runtime. It also removes a few quirks to be true to what it's written.

As result:
 - uClibc mention was removed from README - EOL for 10+ years 
 - dietlibc workaround was removed - 3 releases last 12 years, site cert is expired, uses CVS
 - with the ^^ w/a removed, we can add nsec granularity in depmod

Note: support for uClibc(-ng), dietlibc and others is pretty much welcome. Since my time is limited (to manually check) a CI action to confirm them working would be appreciated.

### DIFF
--- a/tools/depmod.c
+++ b/tools/depmod.c
@@ -2768,7 +2768,8 @@ invalid_syntax:
 	return 0;
 }
 
-static int depfile_up_to_date_dir(DIR *d, time_t mtime, size_t baselen, char *path)
+static int depfile_up_to_date_dir(DIR *d, const struct timespec *ts, size_t baselen,
+				  char *path)
 {
 	struct dirent *de;
 	int err = 1, dfd = dirfd(d);
@@ -2816,7 +2817,7 @@ static int depfile_up_to_date_dir(DIR *d, time_t mtime, size_t baselen, char *pa
 			}
 			path[baselen + namelen] = '/';
 			path[baselen + namelen + 1] = '\0';
-			err = depfile_up_to_date_dir(subdir, mtime, baselen + namelen + 1,
+			err = depfile_up_to_date_dir(subdir, ts, baselen + namelen + 1,
 						     path);
 			closedir(subdir);
 		} else if (S_ISREG(st.st_mode)) {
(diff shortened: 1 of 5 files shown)


testsuite/test-depmod.c
```diff
@@ -16,7 +16,8 @@
 
 #define EXEC_DEPMOD(...)                     \
 	test_spawn_prog(TOOLS_DIR "/depmod", \
-			(const char *[]){ TOOLS_DIR "/depmod", ##__VA_ARGS__, NULL })
+			(const char *[]){    \
+				TOOLS_DIR "/depmod" __VA_OPT__(, ) __VA_ARGS__, NULL })
```
lucasdemarchi, line 20: > C23 introduces __VA_OPT__, which has been supported by GCC/Clang since
> 2018. Let's use that instead of the GNU extension.

but what's the motivation to use it?
1) we are not going to change to stop using gnu extensions
2) `##__VA_ARGS__` is supported everywhere that matters
3) the standard alternative makes it harder to parse

tools/depmod.c
```diff
@@ -2770,7 +2770,8 @@ static int depmod_load_system_map(struct depmod *depmod, const char *filename)
 	return 0;
 }
 
-static int depfile_up_to_date_dir(DIR *d, time_t mtime, size_t baselen, char *path)
+static int depfile_up_to_date_dir(DIR *d, const struct timespec *mtim, size_t baselen,
```
lucasdemarchi, line 2773: why is this variable called mtim rather than e.g. ts?

lucasdemarchi:
2771c5e testsuite: use C23 __VA_OPT__(,) - don't really like it
1a55b0e README: explicitly document compiler requirements - r-b me, removing first commit
a7c6dc4 README: update C runtime library section - r-b me
6a4c268 Remove dietlibc stat::st_mtim workaround - r-b me
019bf8f tools/depmod: use nsec granularity when checking timestamps - one nit

thanks

PR #360, opened 2025-05-29
Allow mixing code and declarations

evelikov:
C90 doesn't allow mixing code and declarations, while we've been C99+ for over 10 years.

In a handful of places it would be useful interleave the code/declarations, making the scope clearer and potentially avoiding shadowing declarations.


lucasdemarchi:
> Namely, by keeping the declarations where the variables are used, better
> outlines the scope, reduces chances of shadowing and helps us avoid
> "unused variable" compiler warning in some paths.

let's follow what the kernel use. While there are cases in which this improves readability, it's a pain to have things like below:

function()
{
...
...
...
...
...
int foo = ...
}

So: I'm fine removing the warning, but let's not blindly add declaration after statement just because we can.

PR #365, opened 2025-05-30
scripts/sanitizer-env.sh: remove incorrect $CC quotes

evelikov:
On Arch (at least) the shell chokes when the CC variable contains space, for the given construct - $("$CC" foobar).

Since we variable is likely to contain the executable and arguments, the quotation is wrong. Drop it and ultimately resolve the LD_PRELOAD issues... That said, the tests still fail over here, segfault-ing in (or due to) the LD_PRELOAD library init-modules.so.


lucasdemarchi:
Applied, thanks.

PR #373, opened 2025-06-16
ci: remove Alpine -Ddocs=false WA

evelikov:
A while ago the Alpine gtk-doc tool chain was having issues building our docs. That got resolved at some point over the past months, so we can drop this local workaround.


lucasdemarchi:
Applied, thanks.

PR #375, opened 2025-06-16
Rework the assert_return() macro

evelikov:
From the commit introducing it:

```
    The current assert_return() macro has a few short comings:
     - the name implies assert()
     - all users unnecessarily provide EXIT_FAILURE
     - the users cannot provide a message - "Failed assertion:" isn't great

    Tweak it up and in the process bump it to ALL_CAPS, to better indicate
    that it's a macro and not a function.
```

... and the follow-up commit:

```
    Swap the manual checks for the new OK() macro. In the process, we remove
    the cleanup in the error path.

    Many other tests don't bother, plus in the places that do try they don't
    always get it right/fully.
```

Before going forward and porting more tests, would be great to hear if the direction is right.

If* it is, a couple of questions:
 - would people prefer different name - OK one is based on wine, but I'm not married to it ;-)
 - how to handle the conversion - one test at a time, bulk (all assert_return -> OK, all manual -> OK), other?

### DIFF
--- a/testsuite/testsuite.h
+++ b/testsuite/testsuite.h
@@ -119,8 +119,6 @@ int test_run(const struct test *t);
 #define WARN(fmt, ...) _LOG("WARN: ", fmt, ##__VA_ARGS__)
 #define ERR(fmt, ...) _LOG("ERR: ", fmt, ##__VA_ARGS__)
 
-#define assert_return(expr, r) TS_ASSERT(expr)
-
 #define TS_ASSERT(expr)                                                         \
 	do {                                                                    \
 		if ((!(expr))) {                                                \
--- a/testsuite/testsuite.h
+++ b/testsuite/testsuite.h
@@ -119,12 +119,14 @@ int test_run(const struct test *t);
 #define WARN(fmt, ...) _LOG("WARN: ", fmt, ##__VA_ARGS__)
 #define ERR(fmt, ...) _LOG("ERR: ", fmt, ##__VA_ARGS__)
 
-#define assert_return(expr, r)                                                  \
+#define assert_return(expr, r) TS_ASSERT(expr)
+
+#define TS_ASSERT(expr)                                                         \
 	do {                                                                    \
 		if ((!(expr))) {                                                \
-			ERR("Failed assertion: " #expr " %s:%d %s\n", __FILE__, \
-			    __LINE__, __PRETTY_FUNCTION__);                     \
-			return (r);                                             \
+			ERR("ASSERTION FAILED at %s:%d\n", __FILE__, __LINE__); \
+			ERR("\t" #expr "\n");                                   \
+			return EXIT_FAILURE;                                    \
 		}                                                               \
 	} while (false)
 
--- a/testsuite/test-array.c
+++ b/testsuite/test-array.c
@@ -20,8 +20,8 @@ static int test_array_append1(void)
 
 	array_init(&array, 2);
 	array_append(&array, c1);
-	assert_return(array.count == 1, EXIT_FAILURE);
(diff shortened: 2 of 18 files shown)


lucasdemarchi:
>     The current assert_return() macro has a few short comings:
>     - the name implies assert()
>    - all users unnecessarily provide EXIT_FAILURE
>    - the users cannot provide a message - "Failed assertion:" isn't great

I think not needing to provide the same EXIT_FAILURE and being able to provide a message is good. The rename I think is not really needed. What's the problem with the term assert?  IMO it's clearer than "ok". igt for example uses a bunch of igt_assert_*.  ok() IMO is odd, but it maybe because I've never seen it used like this. Rust also has both Ok() and assert!(), which I think will add to the confusion.

> how to handle the conversion - one test at a time, bulk (all assert_return -> OK, all manual -> OK), other?

I think a bulk one. If the message is optional, then the first patch is automated. Additional patches may follow later (maybe per compile unit) to add some good messages.

evelikov:
@lucasdemarchi humble poke

lucasdemarchi:
ts_expect() or ts_assert() would be fine IMO. Other places that use the "expect" terminology:  GoogleTest and kunit... Both of them use semantics like

- KUNIT_EXPECT: Marks the test as failed but continues executing the rest of the test case
- KUNIT_ASSERT: Marks the test as failed and immediately terminates the test case.

So for our uses I think it matches more the "assert" semantics

evelikov:
Changes:
 - open-code `assert_return(... -EINVAL)` instance
 - `OK` -> `TS_ASSERT`
   - based on `KUNIT_ASSERT`
   - slightly different formatting/reporting - "inspired" by `KUNIT_ASSERT`
   - no user-provided messages - can add `TS_ASSERT_MSG` if needed
 - different split - introduce macro, sed the whole tree
 - convert all tests to `TS_ASSERT`
 - `s/return EXIT_SUCCESS/return 0/` - obligatory OCD bikeshed :sweat_smile:

evelikov:
Will check the CI failures later today/tomorrow - the PR should be good for review though.

evelikov:
From the failures:
 - arch multilib (clang -m32 failure) - seems like the latest [`lib32-gcc-libs` package has lost `libgcc_s.so`](https://gitlab.archlinux.org/archlinux/packaging/packages/gcc/-/merge_requests/14)
 - style :facepalm: - fixed
 - codecov - temporary unhappy :shrug:

testsuite/testsuite.h
```diff
@@ -119,12 +119,14 @@ int test_run(const struct test *t);
...
+#define assert_return(expr, r) TS_ASSERT(expr)
+
+#define TS_ASSERT(expr)                                                         \
 	do {                                                                    \
 		if ((!(expr))) {                                                \
-			ERR("Failed assertion: " #expr " %s:%d %s\n", __FILE__, \
-			    __LINE__, __PRETTY_FUNCTION__);                     \
-			return (r);                                             \
+			ERR("ASSERTION FAILED at %s:%d\n", __FILE__, __LINE__); \
+			ERR("\t" #expr "\n");                                   \
```
lucasdemarchi, line 126: From commit message:
   
>  - embeds the expression into the error string, growing the file size

which is still true here, right? Also, for the testsuite the file size due to this would not be of great concern to point out.

lucasdemarchi (approved):
Just a nit on commit message. LGTM.

testsuite/testsuite.h
evelikov: It meant that the expression string is separate from the boilerplate "foobar failed" error message, allowing the latter to be de-duplicated. Nevertheless, dropped since it's a very meh point.

lucasdemarchi:
Applied, thanks

PR #376, opened 2025-06-16
Some __attribute__ inspired changes

evelikov:
The series directly or indirectly reduces our __attribute_ instances... With the most interesting patch being `testsuite: fixup argument handling`.

### DIFF
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -227,8 +227,8 @@ static void test_export_environ(const struct test *t)
 		setenv(env->key, env->val, 1);
 }
 
-static noreturn inline int test_run_child(const struct test *t, int fdout[2],
-					  int fderr[2], int fdmonitor[2])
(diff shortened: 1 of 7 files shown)


testsuite/testsuite.c
```diff
@@ -98,7 +98,16 @@ int test_init(const struct test *start, const struct test *stop, int argc,
...
+			if (oneshot) {
+				ERR("oneshot can be provided only once\n");
+				return -1;
+			}
 			oneshot = 1;
+			/*
+			 * The option requires an argument, thus getopt will return an
+			 * index just past it... Even though we want the argument's index.
+			 */
+			optind--;
```
lucasdemarchi, line 110: not a real problem in this case since the other options stop parsing... but modifying optind like this is error prone if you can do things like `foo --foo -n test --bar`.

probably better to leave the optind alone an check for number of arguments in the different  ways depending on the oneshot option, i.e. the test name is still a a positional arg.
```

PR #379, opened 2025-06-21
testsuite/test-modprobe: reinstate chdir error path

evelikov:
Earlier commit used a global sed to remove the no-longer needed exit(EXIT_FAILURE) calls. In the process it also removed one instance in the error path, which should remain.

Fixes: e09acf2e ("testsuite: remove exit from EXEC_TOOL tests")


lucasdemarchi:
Applied, thanks.

PR #380, opened 2025-06-21
Use common-markdown compatible syntax

evelikov:
I was skimming over the kernel.org cgit repo the other day and noticed the markdown* rendering was off. This PR resolves most of it... 
The remaining bits - different fonts, different sized, etc - are out of our reach.


lucasdemarchi:
Applied, thanks.

PR #384, opened 2025-06-27
Smarter kdir

lucasdemarchi:
Add KDIR=any for my own benefit... bonus point that I *think*  it also allows to use it in CI. If we can't, we can simply remove the second commit.


lucasdemarchi:
> Would it make sense to issue a `$(info Using kernel module directory $foobar)` and/or a `$(warnings KDIR is not defined, assuming $foobar)`.

since we don't build it silently, it's already printed by default:

```
$ KDIR=any meson test -C build 
ninja: Entering directory `/home/ldmartin/p/kmod/build'
[0/3] Generating testsuite/build-module-playground with a custom command
make: Entering directory '/home/ldmartin/p/kmod/build/testsuite/module-playground'
make -C /lib/modules/6.15.4-3-MANJARO/build M=$PWD modules
...
```

PR #387, opened 2025-07-07
testsuite/init_module: Handle __NR_riscv_hwprobe in syscall wrapper

andreas-schwab:
The libcrypto library constructor calls the riscv_hwprobe syscall through
the syscall wrapper.  Forward it to the __riscv_hwprobe function.

Signed-off-by: Andreas Schwab <schwab@suse.de>


lucasdemarchi:
Applied, thanks.

PR #390, opened 2025-07-11
testsuite: Fix build with gcc < 15

lucasdemarchi:
Variable declaration inside a switch case works with gcc 15, but fails pretty much everywhere else with a message like below:

	../testsuite/init_module.c: In function ‘syscall’:
	../testsuite/init_module.c:343:3: error: a label can only be part of a statement and a declaration is not a statement
	  343 |   const char *args;
	      |   ^~~~~

Add an empty statement to fix it.

### DIFF
--- a/testsuite/init_module.c
+++ b/testsuite/init_module.c
@@ -339,7 +339,7 @@ TS_EXPORT long int syscall(long int __sysno, ...)
 #endif
 		errno = ENOSYS;
 		return -1;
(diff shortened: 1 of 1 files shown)


lucasdemarchi:
that was my first option, but the indentation turns out horrible and clang-format indents both closing braces with same level. I think this one is ok too.

Applied, thanks.

PR #393, opened 2025-09-30
modprobe: don't attempt to remove an already removed module

jstancek:
In a scenario like following:
  # lsmod | grep -e bnx2i -e cnic
  bnx2i                  94208  0
  libiscsi               94208  1 bnx2i
  cnic                   90112  1 bnx2i
  uio                    32768  1 cnic
  scsi_transport_iscsi   196608  2 bnx2i,libiscsi

  # modprobe -v --remove --remove-holders cnic
  rmmod bnx2i
  rmmod cnic
  rmmod libiscsi
  rmmod cnic
  modprobe: ERROR: libkmod/libkmod-module.c:856 kmod_module_remove_module()
  could not remove 'cnic': No such file or directory

modprobe attempts to remove cnic module twice and propagates that error to the user with a message as well as an exit code.

Add a check to avoid attempts to remove modules that are already gone.

Signed-off-by: Jan Stancek <jstancek@redhat.com>


jstancek:
Thanks for pointer, I'll have a look if I can add some test.

testsuite/meson.build
```diff
@@ -89,6 +89,7 @@ _testsuite = [
   'test-loaded',
   'test-modinfo',
   'test-modprobe',
+  'test-modprobe-remove-holders',
```
evelikov: Nit: lets move the extra test to the existing `test-modprobe.c`

testsuite/testsuite.c
```diff
@@ -166,6 +166,55 @@ int test_spawn_prog(const char *prog, const char *const args[])
 	return EXIT_FAILURE;
 }
 
+int test_spawn_and_wait4_prog(const char *prog, const char *const args[])
+{
+	int err, pid, ret;
+	const char *msg;
+
+	pid = fork();
+	if (pid < 0) { \
```
evelikov: Stray `\`

testsuite/testsuite.c
```diff
@@ -166,6 +166,55 @@ int test_spawn_prog(const char *prog, const char *const args[])
...
+		}
+		return WTERMSIG(err);
+	}
+
+	return 0;
+
+print_error:
+	ERR("%s %s %d\n", prog, msg, ret);
+	while (*args) {
+		LOG("  args: %s\n", *args);
```
evelikov: Nit: let's make this a small helper function. I think we want ERR for both program and args.

evelikov (approved):
Left a handful of comments - would be great to have them addressed, but are non blocking.

Overall I love the idea of `test_spawn_and_wait4_prog` even if the name is a bit long-winded.

Off the top of my head, we could convert all `test_spawn` users to it, merge the two while keeping the shorter name... Although that's completely orthogonal and should be done as follow-up.

Thanks again for the fix

P.S. While I have commit access to the repo, Lucas does the merge/push to both github and kernel.org. He's busy recently so your patience is appreciated :bow:

evelikov:
> @evelikov I've added a test that goes the way of using real /sys, with a custom check for loaded modules. What do you think about this approach?

About to thinker with it myself, but for the sake of posterity - what happens if we don't have the real /sys?

Note that it's breaking the tests/CI which expect self-contained environment.

Aside: there are some trivial clang-format errors.

jstancek:
> Note that it's breaking the tests/CI which expect self-contained environment.

OK, so that approach is a no-go for CI.

> Personally, I would be fine with landing the fix alone and worrying about the rest/test at a later stage.

I've gone back to just modprobe fix here, and leave tests for a separate PR. Thanks for code comments, and scheding some light on how tests work. From what I've gathered I can't load module "for real" and need test infrastructure enhanced to work with refcnt, holders, etc.

evelikov:
FTR the CI failure is unrelated (Debian bullseye repos have moved) and will be resolved with https://github.com/kmod-project/kmod/pull/395.

lucasdemarchi:
Applied, thanks

PR #395, opened 2025-10-26
ci: swap bullseye for bookworm

evelikov:
The former has been EOL for a bit now and somewhat recently has been (re)moved from the repos -> our CI was failing.

Swap for bookworm, which gives us the "try a slightly older distro" coverage, while still being green in CI.


lucasdemarchi:
Applied, thanks

PR #400, opened 2025-11-10
Minor manpage polish, meson cleanups and man de-duplication

evelikov:
Random collection of patches, while browsing through https://github.com/kmod-project/kmod/pull/397

First patch is a squash (+ drop of kmod bits) from Ricardo, the other 3 minor man page nitpicks.

The next 4 are meson/autotools yak shaving.

With the last two de-duplicating man sections + an open question how to credit the other 20 individuals in the COPYRIGHT section.

Patches can be picked individually or as a whole. Happy to split into separate PRs - let me know if you have any preferences.


testsuite/module-playground/.gitignore
```diff
@@ -1,11 +0,0 @@
...
-.cache.mk
-*.mod.c
-.tmp_versions
-*.mod
-*.a
-*.cmd
-*.o.d
-
-modules.order
-Module.symvers
```
lucasdemarchi: commenting here because github doesn't allow me to comment in the proper place. The testsuite/rootfs-pristine/test-loaded/sys/module/btusb/holders/.gitignore and testsuite/rootfs-pristine/test-rootfs/testdir/.gitignore were there to guarantee the dir existence in rootfs

lucasdemarchi (commented):
A few comments, but overall look good. Thanks and sorry for the delay.

testsuite/module-playground/.gitignore
evelikov: Good point, I'll swap them for `.gitkeep` which seem more common/appropriate.

evelikov:
- Reinstate coreutils in Alpine CI - w/o it `ln -sfr` seems to choke.

We can drop it at misc later point when their busybox/toybox/etc is updated

lucasdemarchi:
Applied, thanks

PR #406, opened 2026-01-14
util: Add s390 31 bit mode support

stoeckmann:
Even though size_t is of the same size as a uint32_t, the s390 architecture is a bit picky about its 31 bit mode with size_t.

Use custom code to fix this issue in a rather architecture independent but unfortunately slower way for s390 31 bit mode.

### DIFF
--- a/shared/util.h
+++ b/shared/util.h
@@ -123,7 +123,15 @@ static inline bool uaddsz_overflow(size_t a, size_t b, size_t *res)
 #if __SIZEOF_SIZE_T__ == 8
 	return uadd64_overflow(a, b, res);
 #elif __SIZEOF_SIZE_T__ == 4
+#ifdef __s390__
+	if (b < SIZE_MAX - a) {
+		*res = a + b;
+		return true;
+	}
+	return false;
+#else
 	return uadd32_overflow(a, b, res);
+#endif
 #else
 #error "Unknown sizeof(size_t)"
 #endif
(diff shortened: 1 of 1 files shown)


shared/util.h
```diff
@@ -123,7 +123,15 @@ static inline bool uaddsz_overflow(size_t a, size_t b, size_t *res)
 #if __SIZEOF_SIZE_T__ == 8
 	return uadd64_overflow(a, b, res);
 #elif __SIZEOF_SIZE_T__ == 4
+#ifdef __s390__
```
lucasdemarchi: instead of checking for the specific arch, could we rely on __builtin_types_compatible_p()? If that's not possible, then this lgtm as a bug fix for s390. Thanks for the fix and @akhuettel for testing

stoeckmann: Just tried to get this into. Unfortunately, the compiler complains that types do not match even if you just put the `__builtin_types_compatible_p()` check in front of it. I've thought the compiler might optimize it away for already knowing that its own builtin function would return false...

So I guess it's this fix.

Simple reproducer:
```
#include <stdint.h>
#include <stdio.h>

void my_function32(uint32_t *pointer) { *pointer = 32; }
void my_function64(uint64_t *pointer) { *pointer = 64; }

int
main(void)
{
        size_t result;

        if (__builtin_types_compatible_p(uint64_t, size_t))
                my_function64(&result);
        else if (__builtin_types_compatible_p(uint32_t, size_t))
                my_function32(&result);

        printf("result is %zu\n", result);
        return 0;
}
```

Compiling on x86_64, which should be able to directly remove the second if-check:
```
poc.c: In function ‘main’:
poc.c:15:31: error: passing argument 1 of ‘my_function32’ from incompatible pointer type [-Wincompatible-pointer-types]
   15 |                 my_function32(&result);
      |                               ^~~~~~~
      |                               |
      |                               size_t * {aka long unsigned int *}
poc.c:4:30: note: expected ‘uint32_t *’ {aka ‘unsigned int *’} but argument is of type ‘size_t *’ {aka ‘long unsigned int *’}
    4 | void my_function32(uint32_t *pointer) { *pointer = 10; }
      |                    ~~~~~~~~~~^~~~~~~
```

stoeckmann: Oh, yes that could be nice. But definitely a larger refactoring.

Since this PR has been tested and is a special case for 32 bit systems, I guess it's acceptable for now to have this s390 check in here.

PR #408, opened 2026-01-16
ci: Fix alpine setup

lucasdemarchi:
linux-edge-dev is failing in alpine:

	ERROR: unable to select packages:
	  linux-edge-dev (no such package):
	    required by: world[linux-edge-dev]
	Error: Process completed with exit code 1.

Use linux-stable instead.


lucasdemarchi:
Applied

PR #418, opened 2026-01-29
Remove a couple of elf related functions

evelikov:
Does what is says on the tin. See respective commit messages for more context.

### DIFF
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -660,12 +660,9 @@ static int do_init_module(struct kmod_module *mod, unsigned int flags, const cha
 	off_t size;
 	int err;
 
-	err = kmod_file_load_contents(mod->file);
+	err = kmod_file_get_contents(mod->file, &mem, &size);
 	if (err)
 		return err;
-
-	mem = kmod_file_get_contents(mod->file);
-	size = kmod_file_get_size(mod->file);
 
 	if (flags & (KMOD_INSERT_FORCE_VERMAGIC | KMOD_INSERT_FORCE_MODVERSION)) {
 		if (mod->elf == NULL) {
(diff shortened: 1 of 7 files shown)


libkmod/libkmod-module.c
```diff
@@ -1735,7 +1740,7 @@ KMOD_EXPORT void kmod_module_section_free_list(struct kmod_list *list)
 	kmod_list_release(list, kmod_module_section_free);
 }
 
-static int kmod_module_get_elf(const struct kmod_module *mod, struct kmod_elf **elf)
+static int kmod_module_get_elf(const struct kmod_module *mod)
```
lucasdemarchi: that name now deviates from the other names. In this context "get" is used to return something from the module whereas now this has more the semantics of kmod_file_**load**_contents

evelikov: Split the inlining and kmod_module_get_elf changes into separate patches. Renaming the latter to `kmod_module_load_elf`, which seems more appropriate I think. Open to other name suggestions though.

evelikov:
Pushed an extra commit on top to fold `kmod_file_load_contents` `kmod_file_get_contents` and `kmod_file_get_size` into one, which seemed the natural next step.

Can drop and/or defer to another PR that helps.

lucasdemarchi:
If memory serves me well, the reason for the split used to be to lazily load the struct, when specific parts were needed. Example: lsmod doesn't need anything related to the elf and it would be a waste of time to make it open the elf. AFAICS this is still the case, even after the refactors, right?

evelikov:
Correct - all the lazy allocation/loading/parsing is preserved. Unless, coffee has failed me :sweat_smile:

lucasdemarchi (approved):

evelikov:
@lucasdemarchi this PR is conflicting with the other crypto ones (dlopen openssl/mbedtls/misc nits).

Do you mind merging it or letting me know which order would be more suitable.

lucasdemarchi:
Applied, thanks

PR #420, opened 2026-02-08
Bump openssl to 3.0 and dlopen it

evelikov:
Bit of a follow-up to https://github.com/kmod-project/kmod/issues/47, this allows us to dlopen the final dependency - openssl.

Due their unique versioning for 1.0/1.1 and extensive use of macros/inline functions in their API, I've bumped the requirement to 3.0 as a prep. This should be fine for anything but the deep maintenance more enterprise linuxes, which are not targeting the latest kmod AFAICT.

With that done, recommend the use of `-D dlopen=all` since it can bring measurable/noticable perf/memory improvements.

NOTE: while the compression libraries are tagged "recommended", libcrypto is only a suggested since the only user that I could find is `modinfo`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ Optional dependencies, required with the default build configuration:
 - ZSTD library
 - OPENSSL library (signature handling in modinfo)
 
-Typical configuration and installation
+All the optional dependencies are loaded on demand via `dlopen` by default. This can
+be changed via the `dlopen` build option. Typical configuration and installation:
 
     meson setup builddir/
     meson compile -C builddir/
--- a/libkmod/libkmod-signature.c
+++ b/libkmod/libkmod-signature.c
@@ -15,6 +15,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <shared/elf-note.h>
 #include <shared/missing.h>
 #include <shared/util.h>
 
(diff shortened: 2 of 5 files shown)


evelikov:
v2: Add/install a valgrind suppression file, to silence (most of) the complaints brought by using dlopen.

evelikov:
v3:
 - don't add "openssl" to kmod.pc's module_compressions
 - remove unreachable `DLSYM_LOCALLY_ENABLED` block

README.md
```diff
@@ -48,6 +48,11 @@ Optional dependencies, required with the default build configuration:
 - ZSTD library
 - OPENSSL library (signature handling in modinfo)
 
+All the optional dependencies, can be used either by default or on demand via
+`dlopen`. The latter is recommended and can be enabled via:
+
+    meson setup -D dlopen=all ...
```
lucasdemarchi: let's switch the default in the build system and mention it can be changed to something more conservative if one wants?

lucasdemarchi (approved):

README.md
evelikov: Sounds good to me, thanks for the rebase and patch.

Might want to remove the incomplete `Use \'all\' to .` sentence from `meson_options.txt` since we're touching it.

lucasdemarchi:
Applied, thanks!

PR #423, opened 2026-02-12
Fix various memory leaks in error paths

evelikov:
Hats off to @jaredmauch for finding and the fixes.

### DIFF
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -168,7 +168,7 @@ int test_spawn_prog(const char *prog, const char *const args[])
 
 static void test_export_environ(const struct test *t)
 {
-	char *preload = NULL;
+	_cleanup_free_ char *preload = NULL;
 	size_t preloadlen = 0;
 	size_t i;
 	const struct keyval *env;
@@ -188,7 +188,6 @@ static void test_export_environ(const struct test *t)
 		tmp = realloc(preload, preloadlen + 2 + ldpreloadlen);
 		if (tmp == NULL) {
 			ERR("oom: test_export_environ()\n");
-			free(preload);
 			return;
 		}
 		preload = tmp;
@@ -209,7 +208,6 @@ static void test_export_environ(const struct test *t)
 			tmp = malloc(preloadlen + 2 + len);
 			if (tmp == NULL) {
 				ERR("oom: test_export_environ()\n");
-				free(preload);
 				return;
 			}
 			memcpy(tmp, existing_preload, len);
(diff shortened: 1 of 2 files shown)


lucasdemarchi (approved):
LGTM. I added a commit on top, let me know if it's ok.

lucasdemarchi (approved):

testsuite/testsuite.c
```diff
@@ -188,7 +188,6 @@ static void test_export_environ(const struct test *t)
 		tmp = realloc(preload, preloadlen + 2 + ldpreloadlen);
 		if (tmp == NULL) {
 			ERR("oom: test_export_environ()\n");
-			free(preload);
```
evelikov: This one is fine, since we use realloc and (re)take ownership just below.

testsuite/testsuite.c
```diff
@@ -209,22 +208,18 @@ static void test_export_environ(const struct test *t)
...
 				ERR("oom: test_export_environ()\n");
-				free(preload);
 				return;
 			}
 			memcpy(tmp, existing_preload, len);
 			tmp[len++] = ' ';
 			memcpy(tmp + len, preload, preloadlen);
 			preloadlen += len;
 			tmp[preloadlen] = '\0';
-			free(preload);
```
evelikov, line 218: This one will leak, since we're using malloc and the old buffer will be disowned.

We can swap for realloc(preload), which means we'll need to manually shift the `preload` contents at the end and copy `existing_preload` at the beginning.

Alternatively, we can swap the order - first get/dup the exising `LD_PRELOAD`, then do a two pass on the test variables. First one to get the size - for a one-off allocation - second to copy the contents.

This should make for tad cleaner, albeit slightly longer, code IMHO. Think I nerd-sniped myself...

evelikov (commented):
Extra commit will cause a leak, alternative incoming shortly.

evelikov:
Pushed a revert for `_cleanup_free_` and a patch/two with the alternative.

Feel free to grab/drop whichever.

evelikov:
Updated the last patch to _append_ the existing LD_PRELOAD, thus preserving the original behaviour.

testsuite/testsuite.c
lucasdemarchi: Thanks for catching. I think we can keep the simplest fix: just leave this one free() here. We are replacing `preload` points to, so it's reasonable to expect that it be free'd before doing so.

I'd left the refactor (that is failing CI) for another PR.

PR #412, opened 2026-01-19
depmod: Parallelize dependency computation and output generation

jaredmauch:
Parallelize depmod to leverage multiple CPU cores. The changes include:

- Parallel module loading: Load symbols and module info concurrently across multiple threads
- Parallel dependency resolution: Resolve module dependencies concurrently
- Parallel dependency caching: Pre-compute transitive dependency caches in parallel before output generation
- Parallel output generation: Generate modules.dep and modules.dep.bin files concurrently

Thread safety is ensured through:
- Mutex protection for shared data structures (symbols, dependencies, output)
- Atomic operations for reference counting
- Memory barriers to ensure visibility of atomic operations
- Fine-grained locking to minimize contention

Example improvements are about 5-6x reduction in wall clock time of depmod in my testing, feedback most welcome!

PR #425, opened 2026-02-13
Threaded depmod, take 2

evelikov:
This is a retake of https://github.com/kmod-project/kmod/pull/412, the slimmed down version in particular.

The changes:
 - split out report error on array_append failure into separate patch
 
 - reuse struct symbol, instead or rolling our own
 - split/reuse symbol creation - 2 times fewer allocations, honour sym_prefix
 - don't allocate the struct symbol(s) twice 
 - consolidate cleanup into a helper function

 - remove unused members in load_modules_work, rename
 - swap modules, start and end for modules(_start) and count
 - store thread id and "work" struct in single place - nproc fewerer calloc

 - properly check for errors on pthread_create/join

In summary, the numbers are as follows:

| master | MR, nproc 1 | MR (nproc 8), CPUs used | compression |
|--------|-------------|--------------|-------------|
| 0.44 s | 0.44 s      | 0.73 s, 1.5  | none        |
| 4.45 s | 4.45 s      | 1.40 s, 4.3  | xz          |
| 1.05 s | 1.04 s      | 0.62 s, 2.9  | zst         |

Which are mostly identical to previously, with a notable exception for uncompressed modules. Unsurprisingly adding extra threads damages performance.

Some nproc numbers from an aging 4 core + HT laptop.

| nproc | none | zst  |
|-------|------|------| 
| 1     | 0.44 | 1.04 |
| 2     | 0.39 | 0.73 |
| 4     | 0.54 | 0.63 |
| 8     | 0.73 | 0.62 |
 
 <details>

<summary>Some `perf stat -r 100` numbers</summary>

Even with the 100 repeat, the numbers are not that stable/reliable.

| compression | stat | master | MR nproc 1 | slim | MR |
| ------------|------|--------|------------|------|----|
| uncompressed | task-clock | 435,749,999 | 435,047,972 | 1,082,318,930 | 1,087,043,546 | 
| uncompressed | instructions | 1,213,446,151 | 1,202,426,481 | 1,193,567,961 | 1,192,806,832 |
| uncompressed | time elapsed | 0.43846 +- 0.00177 | 0.43802 +- 0.00184 | 0.72881 +- 0.00417 | 0.73283 +- 0.00318 |
| ------------|------|--------|------------|------|----|
| xz | task-clock | 4,450,100,429 | 4,444,117,416 | 6,063,143,589 | 6,057,782,176 |
| xz | instructions | 27,467,626,437 | 27,439,935,695 | 27,414,581,685 | 27,409,414,470|
| xz | time elapsed | 4.45702 +- 0.00579 | 4.45167 +- 0.00663 | 1.40124 +- 0.00259 | 1.40324 +- 0.00255 |
| ------------|------|--------|------------|------|----|
| zst | task-clock | 1,047,500,349 | 1,038,654,591 | 1,836,694,522 | 1,810,041,398|
| zst | instructions | 6,941,232,755 | 6,976,108,698 | 6,904,710,582 | 6,904,091,683 |
| zst | time elapsed | 1.05134 +- 0.00459 | 1.04285 +- 0.00338 | 0.63994 +- 0.00215 | 0.62171 +- 0.00400 |

</details>

<details>

<summary>Valgrind memory stats</summary>

| allocations | bytes | compression | build |
|-------------|-------|-------------|-------|
| 1,295,300 | 175,206,192 | uncompressed | master |
| 1,295,407 | 180,905,088 | uncompressed | MR nproc 1 |
| 1,322,139 | 176,862,003 | uncompressed | slim |
| 1,295,462 | 176,150,040 | uncompressed | MR |
|-----------|-------------|--------------|----|
| 1,404,387 | 53,370,466,853 | xz | master |
| 1,404,494 | 53,376,165,749 | xz | MR nproc 1 |
| 1,431,226 | 53,372,122,664 | xz | slim |
| 1,404,549 | 53,371,410,701 | xz | MR |
|-----------|-------------|--------------|----|
| 1,306,364 | 1,169,580,231 | zst | master |
| 1,306,471 | 1,175,279,127 | zst | MR nproc 1 |
| 1,333,203 | 1,171,236,042 | zst | slim |
| 1,306,526 | 1,170,524,079 | zst | MR |

Summary, relative to master
- MR, nproc 1 - extra ~100 allocs, ~5.6M bytes in total
- Slim - extra ~27k allocs, ~1.5M bytes in total
- MR - extra ~160 allocs, ~1M bytes in total

The 27k -> 160 changes are expected since we reuse `struct symbols` although why we get an extra 5MB memory for nproc is beyond me :shrug: 

</details>

### DIFF
--- a/tools/depmod.c
+++ b/tools/depmod.c
@@ -11,6 +11,7 @@
 #include <errno.h>
 #include <getopt.h>
 #include <limits.h>
+#include <pthread.h>
 #include <regex.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -1597,8 +1598,10 @@ struct module_symbols {
 	struct array symbols; /* struct symbol */
 };
 
-static int resolve_module_symbols(struct module_symbols *mod_syms)
+static void *resolve_module_symbols(void *arg)
 {
+	struct module_symbols *mod_syms = arg;
+
 	array_init(&mod_syms->symbols, 128);
 
 	for (size_t i = 0; i < mod_syms->modules_count; i++) {
@@ -1622,14 +1625,14 @@ static int resolve_module_symbols(struct module_symbols *mod_syms)
 						   mod);
 			if (sym == NULL) {
 				kmod_module_symbols_free_list(list);
-				return -ENOMEM;
+				return (void *)(intptr_t)-ENOMEM;
 			}
 
 			err = array_append(&mod_syms->symbols, sym);
 			if (err < 0) {
 				free(sym);
 				kmod_module_symbols_free_list(list);
-				return -ENOMEM;
+				return (void *)(intptr_t)-ENOMEM;
 			}
 		}
 		kmod_module_symbols_free_list(list);
@@ -1644,7 +1647,7 @@ load_info:
 
 				err = array_append(&mod->alias_values, value);
 				if (err < 0)
-					return err;
+					return (void *)(intptr_t)err;
 				continue;
 			}
 			if (streq(key, "softdep")) {
@@ -1652,7 +1655,7 @@ load_info:
 
 				err = array_append(&mod->softdep_values, value);
 				if (err < 0)
-					return err;
+					return (void *)(intptr_t)err;
 				continue;
 			}
 			if (streq(key, "weakdep")) {
@@ -1660,7 +1663,7 @@ load_info:
 
 				err = array_append(&mod->weakdep_values, value);
 				if (err < 0)
-					return err;
+					return (void *)(intptr_t)err;
 				continue;
 			}
 		}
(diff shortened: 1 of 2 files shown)


evelikov:
Seems like the CI is failing, even if it's working fine locally. Will check at some point tomorrow/day after.


In the meanwhile,  do people have a preference about:
 - exposing the number of threads to users
 - adjusting the default to nproc/2
 - keeping/nuking/wrapping the `(void *)(intptr_t)` casts

evelikov:
Kicking off helgrind reminded me that we mutate `kmod_ctx` for each new `kmod_module` instance. In that the former has a pool (hash map) with all the modules.

We can resolve, at least some of, the problem by deferring the hash-entry removal(s) aka `kmod_module_unref` to the main thread, which means we don't free until the end of the program - easily reaching 0.5GB of memory. Something we cannot do in the general case.

So as a whole, we could look on making hash implementation lock-free... And as a small aside, to convert the refcounting to use atomics.

To make things even more fun, helgrind is also unhappy with OpenSSL - haven't looked why/how to be honest. It can be side-stepped by a) having a more granular `kmod_module_get_info` variant which doesn't parse the crypto bits and/or b) adding an alternative code-path to OpenSSL.


Since I don't know if I'll be looking at it, I'm going to park this as draft.

evelikov:
If OpenSSL/helgrind proves to be a blocking factor, it can _potentially_ be side-stepped by https://github.com/kmod-project/kmod/pull/426.

tools/depmod.c
```diff
@@ -1644,54 +1647,117 @@ static int resolve_module_symbols(struct module_symbols *mod_syms)
...
 {
-	struct module_symbols mod_syms = {
-		.depmod = depmod,
-		.modules = (struct mod **)depmod->modules.array,
-		.modules_count = depmod->modules.count,
-	};
+	struct thread_info {
+		pthread_t tid;
+		struct module_symbols mod_syms;
+	} *tinfo;
```
lucasdemarchi, line 1688: I was thinking about making mod_syms a thread_local, but... we we will already need to allocate the array for keeping the tid

tools/depmod.c
```diff
@@ -1644,54 +1647,117 @@ static int resolve_module_symbols(struct module_symbols *mod_syms)
...
+	for (unsigned int i = 0; i < n_threads; i++) {
+		int local_err;
+		void *res;
+
+		local_err = pthread_join(tinfo[i].tid, &res);
+		if (err == 0) {
+			if (local_err != 0)
+				err = -local_err;
+			if ((int)(intptr_t)res != 0)
+				err = (int)(intptr_t)res;
```
lucasdemarchi, line 1735: should we worry about the other threads continuing even if one of them already falied? I guess we can keep it simple, otherwise I think we'd need to add a cond var just for this.

tools/depmod.c
```diff
@@ -1561,20 +1561,22 @@ static int depmod_symbol_add(struct depmod *depmod, const char *name, bool prefi
...
+	return sym;
+}
 
-	err = hash_add(depmod->symbols, sym->name, sym);
-	if (err < 0) {
-		free(sym);
+static int depmod_symbol_add(struct depmod *depmod, struct symbol *sym)
+{
+	int err = hash_add(depmod->symbols, sym->name, sym);
+	if (err < 0)
```
lucasdemarchi, line 1576: should we keep the `free(sym)` here to avoid leaking it?

tools/depmod.c
```diff
@@ -1644,54 +1647,117 @@ static int resolve_module_symbols(struct module_symbols *mod_syms)
...
+	if (n_threads > depmod->modules.count)
+		n_threads = depmod->modules.count;
+
+	tinfo = calloc(n_threads, sizeof(*tinfo));
+	if (tinfo == NULL)
+		return -ENOMEM;
+
+	modules_per_thread = (depmod->modules.count + n_threads - 1) / n_threads;
+	last_modules_per_thread =
+		modules_per_thread - (depmod->modules.count % n_threads);
```
lucasdemarchi, line 1705: 11 modules, 4 threads:

```
modules_per_thread == (11 + 3) / 4 == 3
last_modules_per_thread = 3 -  (11 % 4) = 0
```

I think you wanted:

```
last_modules_per_thread = depmod->modules.count - n_threads * modules_per_thread;
```

so the last thread has potentially fewer modules to calculate:

tinfo[0].mod_syms.modules_count = 3
tinfo[1].mod_syms.modules_count = 3
tinfo[2].mod_syms.modules_count = 3
tinfo[3].mod_syms.modules_count = 2

tools/depmod.c
```diff
@@ -1644,54 +1647,117 @@ static int resolve_module_symbols(struct module_symbols *mod_syms)
...
+
+	for (unsigned int i = 0; i < n_threads; i++) {
+		struct module_symbols *mod_syms = &tinfo[i].mod_syms;
+
+		mod_syms->depmod = depmod;
+		mod_syms->modules =
+			(struct mod **)depmod->modules.array + (i * modules_per_thread);
+		mod_syms->modules_count = modules_per_thread;
```
lucasdemarchi, line 1713: humn... assuming the threads are working on different modules, and thus have access to different kmod_mod, I think it should be fine to call into kmod from a concurrency perspective. One place I'm not so sure is when freeing the module - because that puts it back in the pool and there may be a synchronization issue with another thread creating the module.

The other issue would be logging, but that I think we could ignore.

Maybe it would be better from a correctness perspective to clone the kmod_ctx per thread?

lucasdemarchi (changes requested):

lucasdemarchi:
> Kicking off helgrind reminded me that we mutate `kmod_ctx` for each new `kmod_module` instance. In that the former has a pool (hash map) with all the modules.
> 
> We can resolve, at least some of, the problem by deferring the hash-entry removal(s) aka `kmod_module_unref` to the main thread, which means we don't free until the end of the program - easily reaching 0.5GB of memory. Something we cannot do in the general case.
> 
> So as a whole, we could look on making hash implementation lock-free... And as a small aside, to convert the refcounting to use atomics.

sorry, just saw this comment after reviewing... you had already caught that issue.  But this solution still seems fragile - we are breaking the rule of not calling into libkmod from multiple threads. I think just cloning the ctx would be better. This is what e.g. udev does to handle multithread.

tools/depmod.c
evelikov: I don't think we need to worry if a failed execution does not immediately exit/terminate all threads. As a whole it should be faster (tm) than the original non-threaded instance.

tools/depmod.c
evelikov: The commit introducing the split tries to highlight that we have a (somewhat) serious issue with missing error handling - admittedly this isn't making it better. A free here, seems to mess up with the ownership model IMHO.

Perhaps the better options is to use the split API only as needed and have a wrapper (`depmod_symbol_create_n_add`?) for the rest - preserving the current state.

tools/depmod.c
evelikov: Indeed - not sure what I was smoking, while typing that up. Thanks

evelikov:
> I think just cloning the ctx would be better. This is what e.g. udev does to handle multithread.

I am struggling to find a "cloning" libkmod API and the systemd codebase effectively uses a global ctx - one for each program that uses kmod... Do you have any concrete pointers?

PR #427, opened 2026-02-15
Crypto bikeshedding ;-)

evelikov:
Nearly every time I skim through `libkmod-signature.c` it doesn't sit well with how different it is from the rest of the codebase. Which is not surprising, since it was written and then updated by people different from the main contributors.

This doesn't mean it's bad or wrong - simply it doesn't fit with the design/style with the rest of kmod.

This PR drops a few of the small papercuts. Where the notable surgery done in `libkmod/libkmod-signature: rework struct kmod_signature_info`.

Most of the patches are self-contained, so they can be picked individually - in case I've overdone it ;-)

### DIFF
--- a/libkmod/libkmod-internal.h
+++ b/libkmod/libkmod-internal.h
@@ -177,11 +177,8 @@ struct kmod_signature_info {
 	const char *hash_algo, *id_type;
 	const char *sig;
 	size_t sig_len;
-	void (*free)(void *);
-	void *private;
 };
-_must_check_ _nonnull_all_ bool kmod_module_signature_info(const struct kmod_file *file, struct kmod_signature_info *sig_info);
-_nonnull_all_ void kmod_module_signature_info_free(struct kmod_signature_info *sig_info);
+_must_check_ _nonnull_all_ bool kmod_module_signature_info(const struct kmod_file *file, struct kmod_signature_info **sig_info);
 
 /* libkmod-builtin.c */
 _nonnull_all_ ssize_t kmod_builtin_get_modinfo(struct kmod_ctx *ctx, const char *modname, char ***modinfo);
--- a/libkmod/libkmod-internal.h
+++ b/libkmod/libkmod-internal.h
@@ -174,7 +174,7 @@ struct kmod_signature_info {
 	size_t signer_len;
 	const char *key_id;
 	size_t key_id_len;
-	const char *algo, *hash_algo, *id_type;
+	const char *hash_algo, *id_type;
 	const char *sig;
 	size_t sig_len;
 	void (*free)(void *);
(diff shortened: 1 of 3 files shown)


evelikov:
v2: fix the `-D openssl=disabled` build

... will look at the 32bit builds, shortly.

libkmod/libkmod-signature.c
```diff
@@ -314,19 +323,19 @@ bool kmod_module_signature_info(const struct kmod_file *file,
...
+		ret = fill_default(mem, size, &modsig, sig_len, sig_info);
+		break;
 	}
-}
 
-void kmod_module_signature_info_free(struct kmod_signature_info *sig_info)
-{
-	if (sig_info->free)
-		sig_info->free(sig_info);
+	if (ret == true) {
```
lucasdemarchi, line 335: nit: `if (ret)`

libkmod/libkmod-internal.h
```diff
@@ -174,8 +174,11 @@ struct kmod_signature_info {
 	size_t signer_len;
 	const char *key_id;
 	size_t key_id_len;
-	const char *algo, *hash_algo, *id_type; /* XXX: algo is currently unused. See
-						   kmod_module_get_info */
+	/*
```
lucasdemarchi, line 177: why do we have both commit 97387469d37d ("libkmod/libkmod-signature: comment kmod_signature_info::algo") and 78a694936120 ("libkmod/libkmod-signature: comment out kmod_signature_info::algo")? I think we should have just one, just dropping what is unused instead of adding comments.

lucasdemarchi (approved):
Overall LGTM. I left 2 comments on minor things. Thanks

libkmod/libkmod-internal.h
evelikov: It's the organic order as I was browsing through - don't mind either way though.

Dropped the 3 commits for the one, which removes the unused bits.

libkmod/libkmod-module.c
```diff
@@ -1944,7 +1944,7 @@ KMOD_EXPORT int kmod_module_get_info(const struct kmod_module *mod,
 
 list_error:
 	/* aux structures freed in normal case also */
-	kmod_module_signature_info_free(&sig_info);
+	free(sig_info);
```
lucasdemarchi, line 1947: a good candidate for _cleanup_free_ since it's unconditionally freed.

libkmod/libkmod-signature.c
```diff
@@ -191,66 +174,86 @@ static bool fill_pkcs7(const char *mem, off_t size, size_t sig_len,
...
+		goto err1;
 
-	issuer_str = x509_name_to_str(is->issuer);
+	total_len += hash_algo_len + 1;
+
+	/* sig */
+	sig = si->enc_digest;
+	total_len += ASN1_STRING_length(sig);
+
+	sig_info = calloc(1, total_len);
```
lucasdemarchi, line 206: as you mentioned in the commit message, in future we could do: check the total_len in the normal case (i.e. dump it by callinrg this for all modules), then use  DECLARE_STRBUF_WITH_STACK() with a reasonable size.

lucasdemarchi (approved):

lucasdemarchi:
Applied, thanks

PR #429, opened 2026-02-18
Memory leaks follow up to pr 423

lucasdemarchi:
@evelikov moving the refactors from #423 here.


evelikov:
Turns out (as it always does) that the original commit was fine and the last second push broke it :facepalm: 

It should be all good now o/

lucasdemarchi:
Applied, thanks

PR #432, opened 2026-03-25
Add build-id to modinfo

lucasdemarchi:
Kernel uses -Wl,--build-id in the build system, which causes a .notes.gnu.build-id section to be created containing. This can then be used by people developing modules to make sure the running module is the one that was last built.

I used to rely on srcversion for that purpose, but it has 2 issues:
1) It doesn't change if the compilation options change
2) It may be [going away in a future kernel version](https://lore.kernel.org/all/2026031303-prelaunch-creation-3fce@gregkh/)

As per last commit message, this can already be used as is, but is rather cumbersome. I will add more commits later to probably have a `modinfo -F raw-build-id` that simply dumps the section, which would allow to compare with the sysfs with something like:

```
cmp <(cat /sys/module/foobar/notes/.note.gnu.build-id) <(modinfo -F raw-build-id foobar)
```

I'm on the fence about that because we could simply leave modinfo alone for that and add built-in tool:

```
$ kmod build-id --check --verbose foobar
live: a:b:c:...
file: a:b:c:...
match: ok
$ echo $?
0
```

### DIFF
--- a/libkmod/libkmod-elf.c
+++ b/libkmod/libkmod-elf.c
@@ -14,6 +14,7 @@
 
 #include "libkmod.h"
 #include "libkmod-internal.h"
+#include <wchar.h>
 
 /* as defined in module-init-tools */
 struct kmod_modversion32 {
@@ -32,6 +33,7 @@ enum kmod_elf_section {
 	KMOD_ELF_SECTION_STRTAB,
 	KMOD_ELF_SECTION_SYMTAB,
 	KMOD_ELF_SECTION_VERSIONS,
+	KMOD_ELF_SECTION_BUILD_ID,
 	KMOD_ELF_SECTION_MAX,
 };
 
@@ -41,6 +43,7 @@ static const char *const section_name_map[] = {
 	[KMOD_ELF_SECTION_STRTAB] = ".strtab",
 	[KMOD_ELF_SECTION_SYMTAB] = ".symtab",
 	[KMOD_ELF_SECTION_VERSIONS] = "__versions",
+	[KMOD_ELF_SECTION_BUILD_ID] = ".note.gnu.build-id",
 };
 
 struct kmod_elf {
@@ -1251,4 +1254,55 @@ int kmod_elf_get_dependency_symbols(const struct kmod_elf *elf,
 	}
 	free(visited_versions);
 	return count;
+}
+
+#ifndef NT_GNU_BUILD_ID
+#define NT_GNU_BUILD_ID 3
+#endif
+
+/*
+ * Read the .notes.gnu.build-id section that is added by the linker via --build-id. The
+ * section format follows the one defined in shared/elf-note.h, with name == "GNU\0",
+ * type == NT_GNU_BUILD_ID and desc being the hash.
+ */
+int kmod_elf_get_build_id(const struct kmod_elf *elf, const void **hash, size_t *hash_len)
+{
+	uint64_t off, size;
+	uint32_t namesz, descsz, type;
+	const char *name;
+
+	*hash = NULL;
+	*hash_len = 0;
+
+	off = elf->sections[KMOD_ELF_SECTION_BUILD_ID].offset;
+	size = elf->sections[KMOD_ELF_SECTION_BUILD_ID].size;
+	if (off == 0 || size == 0)
+		return -ENODATA;
+
+	if (size < 12) {
+		ELFDBG(elf, "build-id section is too small: %" PRIu64 "\n", size);
+		return -EINVAL;
+	}
+
+	namesz = elf_get_uint32(elf, off);
+	descsz = elf_get_uint32(elf, off + 4);
+	type = elf_get_uint32(elf, off + 8);
+
+	if (size < 12 + namesz + descsz) {
+		ELFDBG(elf, "build-id section is too small: %" PRIu64 "\n", size);
+		return -EINVAL;
+	}
+
+	name = elf_get_mem(elf, off + 12);
+
+	if (type != NT_GNU_BUILD_ID || namesz != 4 || memcmp(name, "GNU\0", 4)) {
+		ELFDBG(elf, "Invalid buid-id section: type %" PRIu32 ", name %.3s\n",
+		       type, name);
+		return -EINVAL;
+	}
+
+	*hash = elf_get_mem(elf, off + 16);
+	*hash_len = descsz;
+
+	return 0;
 }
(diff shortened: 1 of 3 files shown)


libkmod/libkmod-elf.c
```diff
@@ -162,6 +162,11 @@ static uint64_t elf_get_uint(const struct kmod_elf *elf, uint64_t offset, uint16
 	return ret;
 }
 
+static inline uint32_t elf_get_uint32(const struct kmod_elf *elf, uint64_t offset)
```
evelikov, line 165: Previous commit removes a few inline instances, where this commits adds another.

Would be great if we can get a brief note what's the cut-off. Be that a soft or hard rule - as commit message, or in CONTRIBUTING.

libkmod/libkmod-elf.c
```diff
@@ -162,6 +162,11 @@ static uint64_t elf_get_uint(const struct kmod_elf *elf, uint64_t offset, uint16
...
+{
+	return elf_get_uint(elf, offset, 4);
```
evelikov, line 167: `s/4/sizeof(uint32_t)/`

libkmod/libkmod-elf.c
```diff
@@ -14,6 +14,7 @@
 
 #include "libkmod.h"
 #include "libkmod-internal.h"
+#include <wchar.h>
```
evelikov, line 17: Might be missing something - why do we need this include?

libkmod/libkmod-elf.c
```diff
@@ -1252,3 +1255,54 @@ int kmod_elf_get_dependency_symbols(const struct kmod_elf *elf,
...
+ * type == NT_GNU_BUILD_ID and desc being the hash.
+ */
+int kmod_elf_get_build_id(const struct kmod_elf *elf, const void **hash, size_t *hash_len)
+{
+	uint64_t off, size;
+	uint32_t namesz, descsz, type;
+	const char *name;
+
+	*hash = NULL;
+	*hash_len = 0;
```
evelikov, line 1275: No need to preemptively zero initialize the output variables - we don't do that in most/nearly all of the project.

libkmod/libkmod-elf.c
```diff
@@ -1252,3 +1255,54 @@ int kmod_elf_get_dependency_symbols(const struct kmod_elf *elf,
...
+
+	off = elf->sections[KMOD_ELF_SECTION_BUILD_ID].offset;
+	size = elf->sections[KMOD_ELF_SECTION_BUILD_ID].size;
+	if (off == 0 || size == 0)
+		return -ENODATA;
+
+	if (size < 12) {
```
evelikov, line 1282: `s/12/3 * sizeof(uint32_t)/` here and for the second size comparison.

libkmod/libkmod-elf.c
```diff
@@ -1252,3 +1255,54 @@ int kmod_elf_get_dependency_symbols(const struct kmod_elf *elf,
...
+		ELFDBG(elf, "build-id section is too small: %" PRIu64 "\n", size);
+		return -EINVAL;
+	}
+
+	namesz = elf_get_uint32(elf, off);
+	descsz = elf_get_uint32(elf, off + 4);
```
evelikov, line 1288: IIRC we are usually incrementing the offset in kmod:

```
	namesz = elf_get_uint32(elf, off);
    off += sizeof(uint32_t);
	descsz = elf_get_uint32(elf, off);
    ...
```

libkmod/libkmod-elf.c
```diff
@@ -1252,3 +1255,54 @@ int kmod_elf_get_dependency_symbols(const struct kmod_elf *elf,
...
+	type = elf_get_uint32(elf, off + 8);
+
+	if (size < 12 + namesz + descsz) {
+		ELFDBG(elf, "build-id section is too small: %" PRIu64 "\n", size);
+		return -EINVAL;
+	}
+
+	name = elf_get_mem(elf, off + 12);
+
+	if (type != NT_GNU_BUILD_ID || namesz != 4 || memcmp(name, "GNU\0", 4)) {
```
evelikov, line 1298:
```
	if (type != NT_GNU_BUILD_ID || !streq(name, "GNU")) {
```

evelikov (approved):
Code-wise this looks good, left some stylistic suggestions - nothing too crazy I hope 😅

lucasdemarchi:
@evelikov any thoughts about a new build-id subcommand?

evelikov:
> @evelikov any thoughts about a new build-id subcommand?

Without having many other sub-command to compare against - the proposed syntax seems a bit odd IMHO.

For better or worse I appreciate the style and consistency of the systemd "ctl" tools. Would be in favour if we have a proposal of two how we can structure the existing options (modprobe et al) and see how those will align with the build-id bits. 


Aside: please add a note - here in PR, in a commit message, or in-code... more or less anywhere - which kernel version introduced support for `-Wl,--build-id`. Eg. circa 4.10

libkmod/libkmod-elf.c
lucasdemarchi: leftover from interim implementation that wasn't.

libkmod/libkmod-elf.c
lucasdemarchi: humn... namesz is a field we read from the section and I was expecting it to be strlen + NUL. But indeed, let's not error out if for some reason the linker decides to add more \0 to align things.

lucasdemarchi:
> Without having many other sub-command to compare against - the proposed syntax seems a bit odd IMHO.
> 
> For better or worse I appreciate the style and consistency of the systemd "ctl" tools. Would be in favour if we have a proposal of two how we can structure the existing options (modprobe et al) and see how those will align with the build-id bits.

Note that the ctl tools in systemd are targeted at talking to a daemon. Once upon a time I wanted to deprecate the standalone modprobe/insmod/... etc and have a unified "kmod <cmd>". I'm not pursuing that anymore as I don't really see pressuring need to break people's workflow/scripts. However for new things that don't fit the interface of these old tools, I'd rather use a subcommand for kmod.

Here what I want is to give a quick command for kernel module developers to a) check the build-id of a filename/module and b) check if it matches what is loaded in the kernel. I don't think this fits any of modinfo/modprobe/etc.

evelikov:
> Once upon a time I wanted to deprecate the standalone modprobe/insmod/... etc and have a unified "kmod ". I'm not pursuing that anymore as I don't really see pressuring need to break people's workflow/scripts.

That explains things.

Looking at the existing kmod invocations, the following seems more natural fit. Although I don't have a strong opinion either way.

```
$ kmod check build-id --verbose foobar
```

lucasdemarchi:
Applied, thanks!

Issue #428, opened 2026-02-17
Add coverage for more hashalgo types, fix the non-openssl case


evelikov:
It seems that kmod's crypto/hashalgo (and thus tests) have seen limited attention since ~2013.

In that, we're still using the `enum pkey_hash_algo` from kernel v3.12. Upstream kernels have added an UAPI enum with ee08997fee16f10be23c9748d609dbdf3baab8e4 and the kernel tree was updated in 3fe78ca2fb1d61ea598e63fcbf38aec76b36b3a8.

Since then, kmod has gained SM3 support with some of caveats:
 - our local copy has deviated from upstream, thus the non-openssl path was broken
 - the openssl-side relied on compile-time guard 

Follow-up commits upstream (v6.7) added SHA3 support, promted the openssl path of kmod to be updated \o/ Although the enum and non-openssl path fell deeper behind.

Rough ideas/tasks:
- update and ensure we don't deviate again

Namely, update the local enum copy and respective code-base. Add a test to check the size and contents of our vs upstream.
 
- add some sha3 tests

Reuse the style used for the sha1/256 ones.

- Potentially, revert 510c8b7f7455c6613dd1706e5e41ec7b09cf6703 - aka go back to using NID to `enum pkey_hash_algo` mapping...

Not sure to what extend the output of `OBJ_obj2txt()` is guaranteed to be stable. Also the openssl/crypto code-path is quite heavy, so we can drop the extra malloc and 2 function calls... Plus others like mbedtls, don't seems to have equivalent API.

Blaze420420:
> It seems that kmod's crypto/hashalgo (and thus tests) have seen limited attention since ~2013.
> 
> In that, we're still using the `enum pkey_hash_algo` from kernel v3.12. Upstream kernels have added an UAPI enum with ee08997fee16f10be23c9748d609dbdf3baab8e4 and the kernel tree was updated in 3fe78ca2fb1d61ea598e63fcbf38aec76b36b3a8.
> 
> Since then, kmod has gained SM3 support with some of caveats:
>  - our local copy has deviated from upstream, thus the non-openssl path was broken
>  - the openssl-side relied on compile-time guard 
> 
> Follow-up commits upstream (v6.7) added SHA3 support, promted the openssl path of kmod to be updated \o/ Although the enum and non-openssl path fell deeper behind.
> 
> Rough ideas/tasks:
> - update and ensure we don't deviate again
> 
[...]

PR #443, opened 2026-06-01
Some missing.h polish

evelikov:
While moving the openssl code, I've noticed that the `missing.h` handling could be flattened/improved a bit. So here it is:
 - moving the `MODULE_INIT_*` defines to their sole user
 - moving the `missing.h` inclusion where it's used and fixing the fallout

Aside: some v7.1-rc5 commits promote more of the module data/structs that we use to UAPI headers... Which means we could reuse it - either via `missing.h`, pulling a local copy or otherwise - in the not too distant future. Aka somewhat related to https://github.com/kmod-project/kmod/issues/428

### DIFF
--- a/libkmod/libkmod-config.c
+++ b/libkmod/libkmod-config.c
@@ -16,6 +16,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#include <shared/missing.h> // For basename
 #include <shared/util.h>
 
 #include "libkmod.h"
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -22,6 +22,7 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#include <shared/missing.h> // For finit_module
 #include <shared/strbuf.h>
 #include <shared/util.h>
 
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -620,6 +620,16 @@ KMOD_EXPORT int kmod_module_remove_module(struct kmod_module *mod, unsigned int 
 
 extern long init_module(const void *mem, unsigned long len, const char *args);
 
+/*
+ * (Re-)define some linux/module.h macros.
+ *
+ * In practice, most people/distros do not have the relevant package/header during the
+ * build. The values are UAPI and cannot change, so we might as well have them locally.
+ */
+#define MODULE_INIT_IGNORE_MODVERSIONS 1
+#define MODULE_INIT_IGNORE_VERMAGIC 2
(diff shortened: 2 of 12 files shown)


libkmod/libkmod-module.c
```diff
@@ -620,6 +620,16 @@ KMOD_EXPORT int kmod_module_remove_module(struct kmod_module *mod, unsigned int
 
 extern long init_module(const void *mem, unsigned long len, const char *args);
 
+/*
+ * (Re-)define some linux/module.h macros.
+ *
+ * In practice, most people/distros do not have the relevant package/header during the
```
lucasdemarchi, line 627: I'm not sure this comment is true anymore... we could probably just require the header

libkmod/libkmod-config.c
```diff
@@ -16,6 +16,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+#include <shared/missing.h> // For basename
```
lucasdemarchi, line 19: I don't understand the reason for doing this. The "missing.h" is exactly the thing you **don't** want to include explicitly : the idea is that you only include the headers you actually need and if the libc is missing something, it gets resolved by this header.

Also, this patch does something I don't really like: add a comment "// For xyz...". This is not really maintainable  - what happens when this file also needs something else from missing.h?

There's another approach I like for this kind of stuff that spreads the definitions to the headers they are expected on in an additional include like this:


```
// shared/missing/linux/module.h
#include_next <linux/module.h>

#ifndef ....
#define ...
#endif
```

for basename:

```
// shared/missing/libgen.h
#error ...

// shared/missing/string.h
#include_next <string.h>

#if !HAVE_DECL_BASENAME                           
static inline const char *basename(const char *s) 
{
        const char *p = strrchr(s, '/');
        return p ? p + 1 : s;
}
#endif
```

And then add shared/missing/ early in the include path. But I'm not sure it's worth it for our small use missing definitions.

evelikov: Fwiw looking at `missing.h` reminds be of gnulib and some interesting PTSD, of sorts :sweat_smile: 

IMHO missing.h focuses on the wrong aspect - availability - instead of the platform (OS or c runtime), standard or the header which provides the functionality. In addition, it actively masks missing includes and makes things more fragile.

In particular, I've been itching to have macOS/windows build of `depmod` and `modinfo`. Plus would be an interesting exercise to have the Android build also covered - AOSP is lagging at kmod v32.


Having standalone headers was my other option, let me try that. Thanks

affinanti333 (commented):
I appreciate your assistance in recognizing the issue

PR #445, opened 2026-06-19
shared: replace packed-struct get/put_unaligned with memcpy

### LINKED ISSUE #338: Bus error when running kmod on architecture that does not support unaligned access
I encountered problem with kmod when working with bpif3 board. I used gcc 14.2.0, flags: -O2 (or -O3) -march=rv64gv.
Crash looks like this:
```
modprobe[266]: unhandled signal 7 code 0x1 at 0x0000002aba6dc9d6 in kmod[2aba6cd000+22000]

Backtrace:
#0  read_long_mm (p=<synthetic pointer>) at libkmod/libkmod-index.c:632
#1  index_mm_read_node (idx=0x2ad4e953c0, offset=536871449) at libkmod/libkmod-index.c:679
...

Instraction:
v = get_unaligned((uint32_t *) addr);
f9d6           vle32.v v2, (a2)     <-- crash here, load in 32-bit vector register 

Register:
a2             0x3f893a321b     272885232155   <-- unaligned address
```

Maybe this error is related to gcc and vector optimization, but why does kmod use get_unaligned  in this form?
```
#define get_unaligned(ptr)                       \
	({                                       \
		struct __attribute__((packed)) { \
			typeof(*(ptr)) __v;      \
		} *__p = (typeof(__p))(ptr);     \
		__p->__v;                        \
	})
```

There is no memory allocation at aligned address, and "packed" is not part of the C standard, so there is no clear description that it should help in this situation.

I wrote my own implementation of get_unaligned and put_unaligned where there is memory allocation at aligned address: 
```
#define get_unaligned(src) \
    *(typeof(*(src))*)memcpy( \
        (union {char alias[1]; typeof(*(src)) value;}){.value = 0}.alias, \
        (src), sizeof(*(src)))

#define put_unaligned(val, dst) \
    ((void)sizeof(0 ? (typeof(val)*)0 : (dst)), \    /* <-- check for same type val and (*dst)*/
     (void)memcpy((dst), \
        (union {char alias[1]; __typeof(*(dst)) value;}){.value = (val)}.alias, \
        sizeof(val)))
```
With the same compiler flags program stopped crashing.
Here are some examples where program crashes with your macros  (error is visible even in assembler code), but doesn't crash with mine:
Example 1: https://godbolt.org/z/5d1qEMbz5
Example 2: https://godbolt.org/z/cv349fY1K

I would like to know your opinion, is it possible to change implementation of get_unaligned and put_unaligned?

rawrmonster17:
Fixes #338.

## Problem

The current `get_unaligned`/`put_unaligned` macros in `shared/util.h` use a packed-struct pointer-cast pattern:

```c
#define get_unaligned(ptr)                       \
	({                                       \
		struct __attribute__((packed)) { \
			typeof(*(ptr)) __v;      \
		} *__p = (typeof(__p))(ptr);     \
		__p->__v;                        \
	})
```

This is not guaranteed portable. With GCC 14 at `-O2` on RISC-V (and historically on MIPS/SPARC strict-alignment targets), the auto-vectoriser can emit **aligned vector loads** for the surrounding loop in `hash_superfast()`, causing **SIGBUS** on hardware that enforces natural alignment (reported in #338).

The bug affects `shared/hash.c` (4 call sites in the main and fall-through loops) and `libkmod/libkmod-index.c` (1 call site reading a `uint32_t` from a potentially-unaligned address). `put_unaligned` is defined but currently unused.

## Fix

Replace both macros with `memcpy`-based versions:

```c
#define get_unaligned(ptr)                                               \
	__extension__({                                                  \
		typeof(*(ptr)) __v;                                      \
		memcpy(&__v, (ptr), sizeof(__v));                        \
		__v;                                                     \
	})

#define put_unaligned(val, ptr)                                          \
	do {                                                             \
		typeof(*(ptr)) __v = (val);                              \
		memcpy((ptr), &__v, sizeof(__v));                        \
	} while (0)
```

`memcpy` with a compile-time-constant size is recognised by GCC, clang, and ICC and lowered to the same efficient inline sequence as the pointer cast on x86/x86-64, while remaining correct and standards-compliant on all architectures. This is the same approach used by the Linux kernel (`include/linux/unaligned.h`).

`<string.h>` is already included by `shared/util.h`. The macro signatures are unchanged, so all call sites in `hash.c` and `libkmod-index.c` are unaffected.

## Testing

- Built with `meson setup build --buildtype=plain && ninja -C build` — **zero errors, zero warnings** on all changed translation units.
- Ran a dedicated self-check covering all 5 call-site patterns (uint16_t at +1 offset, uint16_t at +3 offset, uint32_t at +1 offset, put uint16_t round-trip, put uint32_t round-trip) with intentionally unaligned pointers — **all pass**.

### DIFF
--- a/shared/util.h
+++ b/shared/util.h
@@ -72,20 +72,17 @@ unsigned long long get_backoff_delta_msec(unsigned long long tend,
 
 /* endianness and alignments                                                */
 /* ************************************************************************ */
-#define get_unaligned(ptr)                       \
-	({                                       \
-		struct __attribute__((packed)) { \
-			typeof(*(ptr)) __v;      \
-		} *__p = (typeof(__p))(ptr);     \
-		__p->__v;                        \
+#define get_unaligned(ptr)                                                   \
+	__extension__({                                                      \
+		__typeof_unqual__(*(ptr)) __v;                               \
+		__builtin_memcpy((void *)&__v, (const void *)(ptr), sizeof(__v)); \
+		__v;                                                          \
 	})
 
-#define put_unaligned(val, ptr)                  \
-	do {                                     \
-		struct __attribute__((packed)) { \
-			typeof(*(ptr)) __v;      \
-		} *__p = (typeof(__p))(ptr);     \
-		__p->__v = (val);                \
+#define put_unaligned(val, ptr)                                              \
+	do {                                                                 \
+		__typeof_unqual__(*(ptr)) __v = (val);                       \
+		__builtin_memcpy((void *)(ptr), (const void *)&__v, sizeof(__v)); \
 	} while (0)
 
 static inline unsigned int align_power2(unsigned int u)


lucasdemarchi:
> This is the same approach used by the Linux kernel (`include/linux/unaligned.h`).

Relevant kernel commit: a339671db64b ("vdso: Switch get/put_unaligned() from packed struct to memcpy()")

lucasdemarchi:
Please use a real user for commit author and add your s-o-b.

It doesn't pass a warning-free build:

```
../shared/util.h:78:24: error: passing argument 1 of ‘memcpy’ discards ‘const’ qualifier from pointer target type [-Werror=discarded-qualifiers]
   78 |                 memcpy(&__v, (ptr), sizeof(__v));                        \
      |                        ^~~~

```

and also triggers array out of bounds:

```
In function ‘memcpy’,
    inlined from ‘read_u32_mm’ at ../libkmod/libkmod-index.c:662:6,
    inlined from ‘index_mm_open’ at ../libkmod/libkmod-index.c:791:16:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:29:10: error: array subscript 0 is outside array bounds of ‘const uint8_t[0]’ {aka ‘const unsigned char[]’} [-Werror=array-bounds=]
   29 |   return __builtin___memcpy_chk (__dest, __src, __len,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   30 |                                  __glibc_objsize0 (__dest));
      |                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~
In function ‘index_mm_open’:
cc1: note: source object is likely at address zero
cc1: all warnings being treated as errors
[25/34] Compiling C object kmod.p/tools_depmod.c.o
ninja: build stopped: subcommand failed.
```

I think adopting a exact same macro as the kernel uses should be fine. This simplified version doesn't work as expected.

rawrmonster17:
Updated the PR branch.

- Rewrote the commit with a real author and `Signed-off-by: Chris Beck <beck.chris88@outlook.com>`.
- Switched the helpers to the kernel-style pattern using `__typeof_unqual__` plus `__builtin_memcpy()` with `void *` casts, rather than the simplified `memcpy()` form.
- Rebuilt on Linux with `gcc` using `-O2 -Werror`; the build now passes warning-free.
- Also rechecked with a 32-bit build and a small targeted self-check for the affected unaligned read/write patterns.

No call sites changed; the update is still localized to `shared/util.h`.

evelikov:
Thanks for the PR @rawrmonster17 - few questions/suggestions, in random order:

- Please double-check our readme and contributing docs - namely the "Hacking" and "Commit trailers" sections.

- Do share the self-check program and methodology used

- Are the new extensions/built-ins available on our minimum gcc/clang versions, should we update the readme? I think we can omit the `__extensions__` keyword.

- In the commit message let's reference the kernel commit and highlight on the strict-aliasing issue, listing the RISC-V improvements as a "bonus".

lucasdemarchi:
__typeof_unqual__ is gcc >= 14 only. Also, even after that, it doesn't seem the generated code is really the same.

PR #448, opened 2026-07-03
Ci updates

evelikov:
Few commits inspired while looking at the code coverage action, namely:

- we can add/bump ubuntu container/runners to 26.04 - much new, so shiny - also mbedtls v3
- ~~codecov action warns about missing python-coverage, even though it should (tm) be using gcov~~ explicitly specify `gcov` as the only plugin - moar codecov v5 "deprecation" snafu
- codecov v5 removed support for `file`, which they listed as "changed" and "deprecated" :stuck_out_tongue: 

Hope this is enough to get it working reliably :crossed_fingers: since the alternative might be a bug in their server code.


lucasdemarchi:
Applied, thanks

PR #449, opened 2026-07-16
Add whitelist feature

### LINKED ISSUE #444: Whitelist support
Several recent Kernel vulnerabilities (like Copy Fail, Dirty Frag, Fragnesia, ...) can be mitigated by blacklisting kernel modules.

It would be great to be able to whitelist modules instead of blacklisting.

This was already proposed in Fedora in 2010, see links below, including a patch.

Prior art:

- https://fedoraproject.org/wiki/Features/ModprobeWhitelist
- https://bugzilla.redhat.com/show_bug.cgi?id=560084
- https://bugzilla.redhat.com/attachment.cgi?id=638907

paulmenzel:
Resolves: https://github.com/kmod-project/kmod/issues/444

### DIFF
--- a/libkmod/libkmod-module.c
+++ b/libkmod/libkmod-module.c
@@ -1001,6 +1001,8 @@ static int kmod_module_get_probe_list(struct kmod_module *mod, bool ignorecmd,
 	return err;
 }
 
+bool whitelist_allowed(const char *modname);
+
 KMOD_EXPORT int kmod_module_probe_insert_module(
 	struct kmod_module *mod, unsigned int flags, const char *extra_options,
 	int (*run_install)(struct kmod_module *m, const char *cmd, void *data),
@@ -1031,6 +1033,9 @@ KMOD_EXPORT int kmod_module_probe_insert_module(
 		if (flags & KMOD_PROBE_APPLY_BLACKLIST)
 			return KMOD_PROBE_APPLY_BLACKLIST;
 	}
+
+	if (!whitelist_allowed(mod->name))
+		return -EPERM;
 
 	err = kmod_module_get_probe_list(mod, !!(flags & KMOD_PROBE_IGNORE_COMMAND),
 					 &list);
--- a/libkmod/libkmod-whitelist.c
+++ b/libkmod/libkmod-whitelist.c
@@ -43,6 +43,7 @@
  */
 static char  **wl_entries = NULL;
 static size_t  wl_count   = 0;
+static bool    wl_testmode = false;
 
 /*
  * 0 = not yet loaded
@@ -114,6 +115,11 @@ static void whitelist_load(void)
 		if (line[0] == '\0')
 			continue;
 
+		if (strcmp(line, "whitelist-test-mode") == 0) {
+			wl_testmode = true;
+			continue;
+		}
+
 		/* Expect: "whitelist <modname>" */
 		if (!(strncmp(line, "whitelist", 9) == 0 &&
 		      isspace((unsigned char)line[9]))) {
@@ -183,6 +189,13 @@ bool whitelist_allowed(const char *modname)
 	for (i = 0; i < wl_count; i++) {
 		if (strcmp(wl_entries[i], modname) == 0)
 			return true;
+	}
+
+	if (wl_testmode) {
+		syslog(LOG_NOTICE,
+		       "kmod whitelist: module '%s' would be denied"
+		       " (test mode active, load permitted)\n", modname);
+		return true;
 	}
 
 	syslog(LOG_NOTICE,
--- a/libkmod/libkmod-whitelist.c
+++ b/libkmod/libkmod-whitelist.c
@@ -1,0 +1,192 @@
+/*
+ * libkmod-whitelist.c - module loading whitelist
+ *
+ * Out-of-tree addition. Intentionally self-contained: no kmod internal
+ * headers, no callbacks into library infrastructure. Only libc.
+ *
+ * Semantics:
+ *   - If WHITELIST_CONF does not exist, all modules are allowed (opt-in).
+ *   - If the file exists but is empty, all modules are denied.
+ *   - If the file has entries, only listed modules are allowed.
+ *
+ * Config file format (/etc/kmod/whitelist.conf):
+ *   # comment lines start with '#'
+ *   whitelist <modulename>
+ *   whitelist <modulename>
+ *   ...
+ *
+ * Module names must be in normalised form (underscores, not hyphens),
+ * both in this file and as passed by the caller. kmod guarantees that
+ * mod->name is always normalised before reaching whitelist_allowed().
+ */
+
+#include <ctype.h>
+#include <errno.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <syslog.h>
+
+#define WHITELIST_CONF "/etc/kmod/whitelist.conf"
+
+/*
+ * Internal state - loaded once per process lifetime.
+ *
+ * Note on thread safety: libkmod is designed around an explicit kmod_ctx
+ * passed through every call, which would make it thread-safe (one context
+ * per thread, no shared state) were it not for this file. These statics are
+ * intentionally process-global and are NOT thread-safe. This is acceptable
+ * because this implementation is scoped to the single-threaded kmod CLI
+ * tools (modprobe, insmod, etc.) and is not intended for use in a
+ * multithreaded embedding context.
+ */
+static char  **wl_entries = NULL;
+static size_t  wl_count   = 0;
+
+/*
+ * 0 = not yet loaded
+ * 1 = loaded, whitelist active (file was present)
+ * 2 = file absent, allow everything
+ */
+static int wl_state = 0;
+
+static void trim_trailing_comment(char *s)
+{
+	char *comment = strchr(s, '#');
+	if (comment)
+		*comment = '\0';
+}
+
+static void trim_trailing_whitespace(char *s)
+{
+	char *end;
+
+	if (*s == '\0')
+		return;
+
+	end = s + strlen(s) - 1;
+	while (end > s && isspace((unsigned char)*end))
+		end--;
+	*(end + 1) = '\0';
+}
+
+static char *skip_whitespace(char *p)
+{
+	while (isspace((unsigned char)*p))
+		p++;
+	return p;
+}
+
+static void whitelist_load(void)
+{
+	FILE *fp;
+	char buf[512];
+	char *line, *modname, *entry;
+	size_t cap = 0;
+	unsigned int linenum = 0;
+
+	fp = fopen(WHITELIST_CONF, "r");
+	if (fp == NULL) {
+		if (errno == ENOENT) {
+			/* File absent: whitelist inactive, allow everything. */
+			wl_state = 2;
+		} else {
+			syslog(LOG_ERR,
+			       "kmod whitelist: failed to open " WHITELIST_CONF
+			       ": %m\n");
+			/* Fail safe: deny all if config is unreadable. */
+			wl_state = 1;
+		}
+		return;
+	}
+
+	wl_state = 1;
+
+	while (fgets(buf, sizeof(buf), fp) != NULL) {
+		line = buf;
+
+		linenum++;
+		trim_trailing_comment(line);
+		trim_trailing_whitespace(line);
+		line = skip_whitespace(line);
+
+		if (line[0] == '\0')
+			continue;
+
+		/* Expect: "whitelist <modname>" */
+		if (!(strncmp(line, "whitelist", 9) == 0 &&
+		      isspace((unsigned char)line[9]))) {
+			syslog(LOG_WARNING,
+			       "kmod whitelist: " WHITELIST_CONF
+			       ":%u: ignoring unrecognised line\n", linenum);
+			continue;
+		}
+
+		modname = skip_whitespace(line + 9);
+		if (modname[0] == '\0') {
+			syslog(LOG_WARNING,
+			       "kmod whitelist: " WHITELIST_CONF
+			       ":%u: missing module name\n", linenum);
+			continue;
+		}
+
+		/* Grow the entry array. */
+		if (wl_count >= cap) {
+			size_t newcap = (cap == 0) ? 16 : cap * 2;
+			char **tmp = realloc(wl_entries,
+					     newcap * sizeof(*wl_entries));
+			if (tmp == NULL) {
+				syslog(LOG_ERR,
+				       "kmod whitelist: out of memory\n");
+				break;
+			}
+			wl_entries = tmp;
+			cap = newcap;
+		}
+
+		entry = strdup(modname);
+		if (entry == NULL) {
+			syslog(LOG_ERR, "kmod whitelist: out of memory\n");
+			break;
+		}
+
+		wl_entries[wl_count++] = entry;
+	}
+
+	fclose(fp);
+}
+
+bool whitelist_allowed(const char *modname);
+
+/*
+ * whitelist_allowed - check whether a module may be loaded.
+ *
+ * Returns true  if the module is permitted (or whitelist is inactive).
+ * Returns false if the whitelist is active and the module is not listed.
+ *
+ * This is the sole public symbol in this file. It is declared with a
+ * forward declaration directly in libkmod-module.c to avoid polluting
+ * any header.
+ */
+bool whitelist_allowed(const char *modname)
+{
+	size_t i;
+
+	if (wl_state == 0)
+		whitelist_load();
+
+	/* File was absent: allow everything. */
+	if (wl_state == 2)
+		return true;
+
+	for (i = 0; i < wl_count; i++) {
+		if (strcmp(wl_entries[i], modname) == 0)
+			return true;
+	}
+
+	syslog(LOG_NOTICE,
+	       "kmod whitelist: module '%s' not in whitelist, denied\n",
+	       modname);
+	return false;
+}
--- a/meson.build
+++ b/meson.build
@@ -364,6 +364,7 @@ libkmod_files = files(
   'libkmod/libkmod-list.c',
   'libkmod/libkmod-module.c',
   'libkmod/libkmod-signature.c',
+  'libkmod/libkmod-whitelist.c',
 )
 
 libkmod_deps = []


paulmenzel:
\cc @sathieu, @buczek

evelikov:
Hi @paulmenzel a few high-level notes:
 - change needs to be properly integrated with kmod and the "out of tree" note effectively resolved
 - the interaction with soft and weak modules needs to be spelled out
 - the manual page need to be updated
 - test coverage is needed for the new feature

The above might be a bit more work than updating the original patch linked in the issue. Although it is doable  - just wondering if it will save you some time.

Out of curiosity: is there any public reference (mailing list, bugzilla, PR, etc) where the original author submitted the change? If so, please add it as a `Link:` trailer.

Thanks o/

buczek:
> Hi @paulmenzel a few high-level notes:
>
> * change needs to be properly integrated with kmod and the "out of tree" note effectively resolved
> * the interaction with soft and weak modules needs to be spelled out
> * the manual page need to be updated
> * test coverage is needed for the new feature
>
> The above might be a bit more work than updating the original patch linked in the issue. Although it is doable - just wondering if it will save you some time.

We could certainly implement all of this and prepare the changes properly for integration into upstream. I'd just like to know beforehand whether the feature would be welcome in principle. It would be a waste to invest time in it if it isn't going to be merged in the end. I was surprised that such an obvious feature doesn't exist, so I'm concerned that there might be fundamental objections to it.

> Out of curiosity: is there any public reference (mailing list, bugzilla, PR, etc) where the original author submitted the change? If so, please add it as a `Link:` trailer.

Original author here.

This was initially intended for our internal distribution only. Accordingly, there was no public discussion.

Here is the original history:

* https://github.molgen.mpg.de/mariux64/bee-files/pull/3358
* https://github.molgen.mpg.de/mariux64/mxtools/pull/533

Best
Donald

> Thanks o/

lucasdemarchi:
> We could certainly implement all of this and prepare the changes properly for integration into upstream. I'd just like to know beforehand whether the feature would be welcome in principle. It would be a waste to invest time in it if it isn't going to be merged in the end. I was surprised that such an obvious feature doesn't exist, so I'm concerned that there might be fundamental objections to it.

I don't think there's a fundamental objection. It makes sense IMO.  I was thinking that maybe it would make more sense in the kernel instead of kmod.... but after reading the PRs you linke it seems more a sysadmin conf, so userspace should be ok.

evelikov:
Same here - as mentioned in the issue, I am in favour.

One question that immediately comes to mind is should the whitelist be exposed via likmod similar to the original patch or not? Would love to see some arguments why/why not from the authors.

buczek:
@paulmenzel: I've added commits on top of your PR branch which mangles it into something better integrated into upstream. Can you pull it from [add-whitelist-feature](https://github.molgen.mpg.de/mariux64/kmod/tree/add-whitelist-feature)  and push it here or give me write access to your repository/branch so that I can push to it. I think we should keep the discussion/changes in this PR.

The commits are RFC and probably need to be rebased at a later time. Maybe the maintainers here don't want to pollute their history with the diversion over our out-of-tree code evolving into something else and would prefer commits which go directly from their master to the end result. Anyway, we can rebase/rewrite to whatever is wanted later.

Also I'm not happy with the logging yet: The specific "not in whitelist" message are probably lost in many environments because of log level LOG_NOTICE and unspecific messages ("could not insert %: EPERM") from modprobe are not really self-explanatory. Upgrading LOG_NOTICE to LOG_ERROR would emit two messages per denied module, which is ugly. 

I think some basics concepts still need to be discussed, though: In the code change, whitelists are built to match blacklists. But this means, that only toplevel modules are checked, dependencies are not. There is `KMOD_PROBE_APPLY_BLACKLIST_ALL` but it looks like it is not used by modprobe. We want to use whitelists as a feature to reduce attack surface and I would say, that it would be safer if all modules would need to be whitelisted explicitly, not just the toplevel module requested by the kernel. But if whitelists don't semantically match blacklists, the functions would be more difficult to explain and understand.

@evelikov , I think I addressed your points in the new commits.  I just wasn't sure about "the interaction with soft and weak modules needs to be spelled out".  I think hard and soft dependencies are handled the same (maybe wrong, see above) and "weak" dependencies are out of scope. As far as I understand it  (but I may by wrong, we don't use that), weak dependency are only used by external tooling, e.g. to decide what goes into a initramfs? When the modules are actually used at a later time, black/whitelists would apply.

paulmenzel:
> @paulmenzel: I've added commits on top of your PR branch which mangles it into something better integrated into upstream. Can you pull it from [add-whitelist-feature](https://github.molgen.mpg.de/mariux64/kmod/tree/add-whitelist-feature) and push it here or give me write access to your repository/branch so that I can push to it. I think we should keep the discussion/changes in this PR.

@buczek, awesome that you found time to wok on this.I pushed your changes, and invited you as a collaborator.

buczek:
AI assistance disclosure: Portions of this patch series (implementation, tests, and documentation) were drafted with AI assistance under my direction and review. All design decisions, verification, and final review were performed by me.

evelikov:
Don't have a strong opinion on the LLM/AI use and in this case there is nothing ground breaking that would warrant licensing and/or other concerns, IMHO.

Would be great if @lucas can put together some rough notes in our contributing guide, for posterity sake.

Glancing through the patches (proper review will come in a few days), they all seem ok. Some questions come to mind:

 - some duplication wrt the "blacklist" equivalents - should be able to flesh out and reuse some helpers
 - what is the current means of obtaining the complete list - set flag, reboot, plug all devices, grep the logs? can we do something easier via say `modprobe` or `depmod`
 - if we stick with the original ^^ approach, can we merge `whitelist-enable` and `whitelist-test-mode` one - a tristate `whitelist` -> `enable` `disable` `test`
 - wrt the module interactions (hard, soft and/or weak), have you considered:
   - what should be reported if a required module is loaded, prior to having the "test-mode" enabled? eg. early initrd without the config
   - should all dependencies be tracked - including soft & weak ones?
   - how can admins ensure the list is complete? eg. is it uncommon/not allowed for a person to come with their own keyboard/mouse and expect it to work?

I suspect some of these might be covered in the patches already. If so, feel free to ignore ;-)

lucasdemarchi:
> Don't have a strong opinion on the LLM/AI use and in this case there is nothing ground breaking that would warrant licensing and/or other concerns, IMHO.
> 
> Would be great if @lucas can put together some rough notes in our contributing guide, for posterity sake.

I guess you referred to me, even though that's not my @ handle :)

"AI is a tool, just like other tools we use.  And it's clearly a useful one." - https://lore.kernel.org/linux-media/CAHk-=wi4zC+Ze8e+p3tMv8TtG_80KzsZ1syL9anBtmEh5Z40vg@mail.gmail.com/

buczek:
>     * some duplication wrt the "blacklist" equivalents - should be able to flesh out and reuse some helpers

Yes, the whitelist helpers are direct copies of the blacklist helpers and I've considered deduplication. In the end I decided to avoid changing existing code at this stage. Deduplication can be done with simple and easy to verify separate patches on top of these changes, either in this PR or independently. It's only a minor source code style issue, I guess the compiler would inline the short static functions anyway, thereby removing a "list" argument and using the caller's constants directly.

Someone just needs to tell me whether I should do it right away.

>     * what is the current means of obtaining the complete list - set flag, reboot, plug all devices, grep the logs? can we do something easier via say `modprobe` or `depmod`

We used "lsmod" across all our live machines and created a whitelist for our fleet from that. Then we let it run a few days in "test mode" to see if anything not yet covered is sometimes plugged in dynamically. Then we enabled it. Sysadmins need to be aware of it, because that can be relevant when new hardware is provisioned or kernel is updated. For that reason, a reliable and understandable log is important, see my note about that. In the out-of-tree variant we used syslog() with specific messages but this has been changed now to libkmod standard and I don't even know how other distributions configure their logs.

>     * if we stick with the original ^^ approach, can we merge `whitelist-enable` and `whitelist-test-mode` one - a tristate `whitelist` -> `enable` `disable` `test`

Yes, of course. Can someone decide, please?

>     * wrt the module interactions (hard, soft and/or weak), have you considered:
>       
>       * what should be reported if a required module is loaded, prior to having the "test-mode" enabled? eg. early initrd without the config

Hmmm. I wouldn't use whitelists in initrd at all. Our usage is kind of a lockdown mode. "no new modules without explicit approval". This is to prevent unprivileged attack code from triggering autoload of kernel modules with vulnerabilities. 

>       * should all dependencies be tracked - including soft & weak ones?

The lists don't apply to hard dependencies and I regard weak dependencies as out of scope (see my previous reply). ~~The lists do apply to soft dependencies, though, because these are separately requested by the kernel~~ (edit: The lists don't apply to soft dependencies). This is the current state with blacklists and the proposed whitelists follow along. Please correct me if I am wrong.

For us, that would definitely be fine. Is somebody requesting a change?

>       * how can admins ensure the list is complete? eg. is it uncommon/not allowed for a person to come with their own keyboard/mouse and expect it to work?

I can only answer for our environment: No, they are not expected to connect their own hardware. Nevertheless most simple hardware like keyboards/mice works with the generic drivers which are already available. If they connect something which would require a new kernel module to be loaded, that would fail and the user would contact our helpdesk for assistance. This is a good thing, because we can review what the user is trying to do and either allow it or otherwise assist in resolving the underlying problem in a correct way. 

I think a fleet of multi-user systems is uncommon nowadays. But a security-aware self-admin might use "lsmod" on his notebook and generate a whitelist from that with the same intent.

> I suspect some of these might be covered in the patches already. If so, feel free to ignore ;-)

Thanks for your thoughts!

buczek:
Should we copy the KMOD_PROBE_APPLY_BLACKLIST* API for whitelist? I hesitate, because its somehow strange with KMOD_PROBE_APPLY_BLACKLIST_ALIAS_ONLY defaulted in modprobe.c.

q1sh101:
On the soft dependency question, measured on kmod 34.2 in an isolated config dir so it reproduces anywhere:

```sh
d=$(mktemp -d)
printf 'softdep xfs pre: e1000e\ninstall e1000e /bin/false\n' > "$d/t.conf"
modprobe -C "$d" -n -v xfs
```

For a module reached through a softdep, the two deny mechanisms behave differently. An `install` command on `e1000e` is applied and the plan shows `install /bin/false`. A `blacklist e1000e` line is not, and `e1000e` is inserted anyway. So `blacklist` does not apply to soft dependencies, but `install` does.

The other direction is less obvious. If the requested module itself declares a `pre:` or `post:` softdep, its own `install` command is dropped and it is inserted directly:

```sh
printf 'install xfs /bin/false\nsoftdep xfs pre: e1000e\n' > "$d/t.conf"
modprobe -C "$d" -n -v xfs
```

This follows `mod->ignorecmd = (pre != NULL || post != NULL)` in `__kmod_module_fill_softdep()`, and modprobe.d(5) notes that softdep takes precedence over install. A `softdep` line with neither `pre:` nor `post:` leaves the command intact.

I cannot say much about `weakdep`. `modules.weakdep` is empty on this kernel, so I could only exercise the config directive, where the command is applied.

Since the two directions differ, it would help to state which one `whitelist` follows, in particular whether a `pre:`/`post:` softdep can pull in a module that `whitelist-enable` does not permit.

libkmod/libkmod-module.c
```diff
@@ -1002,6 +1018,24 @@ static int kmod_module_get_probe_list(struct kmod_module *mod, bool ignorecmd,
...
+static bool module_whitelist_check(struct kmod_module *mod)
+{
+	const struct kmod_config *config = kmod_get_config(mod->ctx);
+
+	if (!config->whitelist_active || module_is_whitelisted(mod))
+		return true;
+
+	if (config->whitelist_test_mode) {
+		NOTICE(mod->ctx,
+		       "module '%s' would be denied (test mode active, load permitted)\n",
```
evelikov: Let's have a clear/unique prefix like the original patch. Eg. `whitelist: module %s....`

Here and throughout.

libkmod/libkmod-module.c
```diff
@@ -1033,6 +1067,9 @@ KMOD_EXPORT int kmod_module_probe_insert_module(
 			return KMOD_PROBE_APPLY_BLACKLIST;
 	}
 
+	if (!module_whitelist_check(mod))
+		return -EPERM;
+
```
evelikov, line 1079: Curious about the train of thought here - do we want the check prior to any of the others - already_loaded, blacklist - or why not?

man/modprobe.d.5.scd
```diff
@@ -156,6 +156,27 @@ weakdep _modulename_ _modules_
 	required/desired at runtime. When c is loaded and is being probed, it
 	may issue calls to request_module() causing a or b to also be loaded.
 
+whitelist _modulename_
+	Adds _modulename_ to the whitelist of modules that are permitted to
+	load. This command only has an effect once *whitelist-enable* is
+	also given (in this or any other configuration file); by itself, a
+	*whitelist* entry is parsed but has no effect.
```
evelikov, line 163: Not strictly related to this PR: Ideally we would have a way to validate the modulename(s) listed in the config. Think typos, people accidentally adding multiple on the same line (space, comma, other-separated), etc.

None of this is a blocker for this feature, but if you feel like sending separate PR that would be really appreciated.

testsuite/test-whitelist.c
```diff
@@ -0,0 +1,66 @@
...
+static int whitelist_deny_all(void)
+{
+	return EXEC_TOOL(modprobe, "mod-simple");
+}
+DEFINE_TEST(whitelist_deny_all,
+	.description = "check that modules are denied when whitelist-enable is set with no entries",
+	.config = {
+		[TC_UNAME_R] = "3.3.3",
+		[TC_ROOTFS] = TESTSUITE_ROOTFS "test-whitelist/deny-all",
+		[TC_INIT_MODULE_RETCODES] = "",
```
evelikov: Do we need the empty TC_INIT_MODULE_RETCODES here?

testsuite/test-whitelist.c
```diff
@@ -0,0 +1,66 @@
...
+	},
+	.expected_fail = true,
```
evelikov: Not related to your work: we should get a exit-status field, since expected_fail effectively masks any failure - crash, sanitizers, etc.

man/modprobe.d.5.scd
```diff
@@ -156,6 +156,27 @@ weakdep _modulename_ _modules_
...
+
+whitelist-enable
+	Enables whitelist enforcement: once this directive appears in any
+	configuration file, only modules listed via *whitelist* commands
+	(anywhere in the configuration) are permitted to load, and all other
+	modules are denied. If *whitelist-enable* is given but no
+	*whitelist* entries exist, every module is denied. Without
+	*whitelist-enable*, no whitelist is enforced regardless of any
+	*whitelist* entries present, and all modules may load as usual.
```
evelikov: Let's shorten this up a bit - my somewhat ADHD brain stopped reading past "configuration file" the first time.

	Enables whitelist enforcement, where only modules listed via
    *whitelist* commands are permitted to load, and all other
	modules are denied.

evelikov (commented):
Was LLM/AI also used to replying to the review questions? They seems overly verbose and somewhat cautions not to make a statement.

Eg.
> Yes, the whitelist helpers are direct copies of the blacklist helpers and I've considered deduplication. In the end I decided to avoid changing existing code at this stage. Deduplication can be done with simple and easy to verify separate patches on top of these changes, either in this PR or independently. It's only a minor source code style issue, I guess the compiler would inline the short static functions anyway, thereby removing a "list" argument and using the caller's constants directly.

> Someone just needs to tell me whether I should do it right away.

Is usually `I've considered it, but decided against for now. Can fixup or send a follow-up PR if people prefer`

Let me see if I understood correctly, the usual/recommended workflow is:
 - construct a list from lsmod - PR should include a patch to make that process easier
 - enable test mode, gather logs/data, analyse and amend whitelist config - love the use of NOTICE (we don't use it yet), but we could use a phrase (like the original patch) to make process easier
 - no whitelisting - active or test-mode is present in initrd
 - for any issues - driver not loading, need keyboard, etc - sysadmin is involved

Overall I like the idea, although it brings a few questions - not relevant for kmod, but something to consider potentially in your org/setup.
 - does the initrd contain full-blown kernel module set - aka people can bypass the whitelist all together, the module isn't listed during test-mode phase. 
 - in case of keyboard failures (infamous coffee spills), is the recovery process "remove the storage drive" (think soldered ssd/nvme) or there are sufficient backup devices


Wrt weak dependencies - they cannot be out of scope. The way they work is:
 - kernel module calls `request_module(foobar)`, kernel executes `modprobe -q -- foobar`
 - kernel module adds `MODULE_WEAKDEP` blurb, to alert userspace  - eg. make sure the module is available in initrd/rootfs

So as a whole, I think we don't want/care about the dependency type - hard/soft/weak/etc. Similarly, I don't see the point in adding `KMOD_PROBE_APPLY_` - if anything, that would defeat the purpose of the option.

That said, I don't mind being proven wrong :smile: 


Left a few small inline comments. Let's rebase & squash this up (1-2 patches) and pump the commit message to include, some brief about:
 - the why
 - how-to-use recommendation

buczek:
> Was LLM/AI also used to replying to the review questions? They seems overly verbose and somewhat cautions not to make a statement.
> [...]

It was not. Natural and limited intelligence and sorry for being too verbose. I used AI as a spell checker, though, because I tend to make a lot of typos, use wrong wording and overlook careless mistakes after editing my text over and over.  <https://claude.ai/share/7f18f71d-8fdc-43ea-9335-2bdbcf5e46ee>

> Let me see if I understood correctly, the usual/recommended workflow is:
> 
>     * construct a list from lsmod - PR should include a patch to make that process easier
>     * enable test mode, gather logs/data, analyse and amend whitelist config - love the use of NOTICE (we don't use it yet), but we could use a phrase (like the original patch) to make process easier
>     * no whitelisting - active or test-mode is present in initrd
>     * for any issues - driver not loading, need keyboard, etc - sysadmin is involved

This is just what we did. Not sure if that is the way to go in other environments. 

> Overall I like the idea, although it brings a few questions - not relevant for kmod, but something to consider potentially in your org/setup.
> 
>     * does the initrd contain full-blown kernel module set - aka people can bypass the whitelist all together, the module isn't listed during test-mode phase.

Our initrd doesn't contain any dynamic kernel modules or kmod. All modules required to read the root filesystem are static.  So people can't bypass the whitelist by having hardware plugged in during boot. When the system switches to the real rootfs, which provides dynamic kernel modules, the whitelist configuration is there at the same time.

>     * in case of keyboard failures (infamous coffee spills), is the recovery process "remove the storage drive" (think soldered ssd/nvme) or there are sufficient backup devices

Sorry, I don't get that scenario. 

> Wrt weak dependencies - they cannot be out of scope. The way they work is:
> 
>     * kernel module calls `request_module(foobar)`, kernel executes `modprobe -q -- foobar`
>     * kernel module adds `MODULE_WEAKDEP` blurb, to alert userspace  - eg. make sure the module is available in initrd/rootfs
> 
> So as a whole, I think we don't want/care about the dependency type - hard/soft/weak/etc. Similarly, I don't see the point in adding `KMOD_PROBE_APPLY_` - if anything, that would defeat the purpose of the option.

Okay, thanks for the opinion. I need that, because we never managed dependencies manually so I have no experience on my own with all that. We just configure our kernels with a lot of dynamic modules so that we don't need to compile a new kernel or reboot when we want to play with a new feature, then `make modules_install`. Aside from a few modules which had security problems and were blacklisted before we had the whitelist feature and a few options for some modules, our modprobe.d has nothing. I had to look up "soft" and "hard" and "weak" dependencies when it was mentioned here. 

> That said, I don't mind being proven wrong 😄
> 
> Left a few small inline comments.
>
> Let's rebase & squash this up (1-2 patches) and pump the commit message to include, some brief about:
> 
>     * the why
> 
>     * how-to-use recommendation

Thanks! I will look into your comments and rebase. I hope I find time next week. Can't promise, tough.

libkmod/libkmod-module.c
buczek: The genesis of this was that we didn't want our kernel to dynamically load modules not explicitly whitelisted by us to reduce attack surface. So the simple logic was that modprobe when used by the kernel to load a module should work like this: "requested module not whitelisted? Reject". Already loaded modules are already loaded at that stage, nothing we can do about that . A module on both, blacklist and whitelist, would not be loaded, which seems to be the better of the two options (less surprising, less change of existing behavior, whitelist required is on top of the existing blacklist feature, this is a security feature so default to the fail-save position to not load a module)

libkmod/libkmod-module.c
buczek: Good point, will do.

man/modprobe.d.5.scd
buczek: Well, yes, we had a stricter parsing and error messages and features (dangling comments) in the original code reading `/etc/kmod/whitelist.conf`. Now the idea was to do proper upstream integration, so we use the existing functions to read `modprobe.d` files. So the parsing of "whitelist XXX" is now identical to the existing "blacklist XXX".

If this should be approved, I think that should go into a separate PR.

testsuite/test-whitelist.c
buczek: I just followed the current code style. See `git grep TC_INIT_MODULE_RETCODES`.

testsuite/test-whitelist.c
buczek: Okay, so no change requested here.

man/modprobe.d.5.scd
buczek: Agreed, will shorten that as suggested.

buczek:
The builtin logger ( `NOTICE()` ) doesn't work at all, because the kernel defaults to `modprobe -q` ( `kernel/module/kmod.c`) and `-q` sets `verbose = LOG_EMERG` and calls `log_setup_kmod_log(ctx, verbose);`  (`tools/modprobe.c`). I have to rethink the logging, which is essential for this feature to be useful.

buczek:
>     * if we stick with the original ^^ approach, can we merge `whitelist-enable` and `whitelist-test-mode` one - a tristate `whitelist` -> `enable` `disable` `test`

I hesitate. One advantage of a separate `whitelist-test-mode` is that you can leave it in the file but commented outfor easy switching:

```
whitelist-enable
# whitelist-test-mode
whitelist MODULE
whitelist MODULE
whitelist MODULE
...
```

If you make it tristate, you probably need the same number of lines to remind the admin of the correct setting for test mode.but the required in-line replacement to switch is easier to do wrong:

```
# replace "whitelist enable" with "whitelist test" for test mode
whitelist enable
whitelist MODULE
whitelist MODULE
whitelist MODULE
...
```

Plus this would clash with kernel modules "enable" or "test" or we'd need to use another keyword...

buczek:
Rebased to upstream/master and squashed into one commit. Changes to
previous code:

- Prefix the two whitelist NOTICE() log messages with "whitelist: "
  for easier identification in logs.
- Shorten the whitelist-enable description in modprobe.d.5.scd.
- Fold test-whitelist into test-modprobe, matching how blacklist's
  modprobe-level tests are organized upstream (test-blacklist.c no
  longer exists as a standalone binary either).
- Make the whitelist deny/test-mode messages bypass modprobe's usual
  log verbosity gate. The kernel always invokes the autoload helper
  as "modprobe -q", which sets verbosity to LOG_EMERG and silences
  these messages entirely, defeating whitelist auditing.

buczek:
>     * construct a list from lsmod - PR should include a patch to make that process easier

Instructions or a script? Where should I put it?

Issue #381, opened 2025-06-26
RFC tests reorg


evelikov:
While going through the earlier test PRs (https://github.com/kmod-project/kmod/pull/258 and https://github.com/kmod-project/kmod/pull/375), a few things of varying annoyance stood out. Each of these is mostly independent, so one can thumbs up/down individually or as a whole:

 - inconsistent cleanups in error path - mostly addressed by latter PR, finish and merge
 - custom checks vs `assert_return()` - ditto
 - test names (as per `test-foo --list`) starting with `test_` (61) vs not (55) - remove the prefix
 - arbitrary/non-descriptive test names (`alias_1`, `from_alias`, etc) - use `util_alias`, `libkmod_new_from_alias`, aka `$top-level-dir_testname` or `$tool_testname`
 - ^^ for filenames (`test-blacklist.c`, `test-dependencies.c`, `test-loaded.c`, etc) - use `test-libkmod-foobar.c`
 - multiple-files-per-tool (`test-modprobe`, `test-weakdep`, etc) - merge into one
 - one-test-per-file (mostly likmod API) - merge, unless it causes problems
 - dozen+ test binaries, non-unique testnames (practically identical/duplicated `test_remove`), manually running tests (tab completion doesn't work in `sh -c 'source ....`) - fold into single binary

@lucasdemarchi @stoeckmann do let me know what you think. All of these are a net positive IMHO, with the* minus being the temporary churn. If you see any item as a potential deterrent from adding more test, please let me know.

evelikov:
@lucasdemarchi any preferences on the above?

Would prefer to have things somewhat well-defined and consistent before we expand on the tests - hint the recently opened softdep bug and the other one from 2024.

lucasdemarchi:
> * inconsistent cleanups in error path - mostly addressed by latter PR, finish and merge
> * custom checks vs `assert_return()` - ditto

Both replied in #375 

> * test names (as per `test-foo --list`) starting with `test_` (61) vs not (55) - remove the prefix

Fine, but you will need some changes in the DEFINE_TEST() macro. Example:

`static int test_strbuf_pushchar(void)`

This is a test for the strbuf_pushchar() internal API. So probably:

-#define DEFINE_TEST(_name, ...) DEFINE_TEST_WITH_FUNC(_name, _name, __VA_ARGS__)
+#define DEFINE_TEST(_name, ...) DEFINE_TEST_WITH_FUNC(_name, test_ ## _name, __VA_ARGS__)

[...]

Issue #352, opened 2025-05-18
Check/remove testsuite retcode


evelikov:
AFAICT having a retcode in the testsuite is redundant and error-prone. If we have an errno, then the syscall should return -1...

In addition, looking at the instances we have in-tree, we have a retcode of 1, 0 and -1 - something doesn't feel right.

evelikov:
While in here, we should probably error out early if `TC_*_MODULE_RETCODES` is malformed in any way.

PR #450, opened 2026-07-18
Merge/group individual tests

evelikov:
As mentioned in https://github.com/kmod-project/kmod/issues/381 this MR groups the tests, during which a pre-existing bug stood out... yay?

In practice it does a little bit more than just grouping... As always things grew organically - happy to split in other PRs if that helps.
 - fixes the loop order so `test-foo --list` is no longer the inverse of that in the source
 - removes the no longer needed UNIQ macro
 - removes/converts some `LOG` instances and `TS_` prefixes the (tests) logging macros - removing the `undef ERR` hack
 - renames `tmpfile-util.[ch]` to `tmpfile.[ch]` - OCD to the "rescue"
 - folds the array/hash/strbuf/util tests into `test-shared`
 - folds the respective (public API) libkmod tests into `test-libkmod`
 - renames the (internal) kmod_list tests (test-list) to `test-libkmod-list`
 - adds a missing Copyright year
 - drop double-space instances in Copyright blurb

As result, we get the exact same (sub)tests executed _yet_ grouped into 8 tests (instead of 19). If people prefer a long tests list printed by the runner, we can add the notion of (sub)groups aka `test-shared --group array` to achieve that.

### DIFF
--- a/testsuite/init_module.c
+++ b/testsuite/init_module.c
@@ -32,8 +32,6 @@
 /* kmod_elf_get_section() is not exported, we need the private header */
 #include <libkmod/libkmod-internal.h>
 
-/* FIXME: hack, change name so we don't clash */
-#undef ERR
 #include "testsuite.h"
 #include "stripped-module.h"
 
--- a/testsuite/test-blacklist.c
+++ b/testsuite/test-blacklist.c
@@ -18,8 +18,6 @@
  * function rather pointless */
 #include <libkmod/libkmod-internal.h>
 
-/* FIXME: hack, change name so we don't clash */
-#undef ERR
 #include "testsuite.h"
 
 static int blacklist_1(void)
--- a/testsuite/test-blacklist.c
+++ b/testsuite/test-blacklist.c
@@ -16,6 +16,7 @@
 
 /* good luck building a kmod_list outside of the library... makes this blacklist
  * function rather pointless */
+/* FIXME: convert the test so we don't need the internal header */
 #include <libkmod/libkmod-internal.h>
 
 #include "testsuite.h"
--- a/testsuite/testsuite.c
+++ b/testsuite/testsuite.c
@@ -84,7 +84,7 @@ int test_init(const struct test *start, const struct test *stop, int argc,
 
 	/* An empty testsuite is not likely intended */
 	if (start == stop) {
-		ERR("Empty testsuite found. Did the compiler toolchain discard them?\n");
+		TS_ERR("Empty testsuite found. Did the compiler toolchain discard them?\n");
 		return -EINVAL;
 	}
 
@@ -106,19 +106,19 @@ int test_init(const struct test *start, const struct test *stop, int argc,
 		case '?':
 			return -1;
 		default:
-			ERR("unexpected getopt_long() value %c\n", c);
+			TS_ERR("unexpected getopt_long() value %c\n", c);
 			return -1;
 		}
 	}
 
 	if (oneshot) {
 		if ((optind + 1) != argc) {
-			ERR("Too many arguments provided. Only one is supported - the testname\n");
+			TS_ERR("Too many arguments provided. Only one is supported - the testname\n");
 			return -1;
 		}
 	} else {
 		if (optind != argc && (optind + 1) != argc) {
-			ERR("Invalid number of arguments provided. One optional can be provided - the testname\n");
+			TS_ERR("Invalid number of arguments provided. One optional can be provided - the testname\n");
 			return -1;
 		}
 	}
@@ -151,7 +151,7 @@ static int test_spawn_test(const struct test *t)
 
 	execv(progname, (char *const *)args);
 
-	ERR("failed to spawn %s for %s: %m\n", progname, t->name);
+	TS_ERR("failed to spawn %s for %s: %m\n", progname, t->name);
 	return EXIT_FAILURE;
 }
 
@@ -164,8 +164,8 @@ int test_spawn_prog(const char *prog, const char *const args[])
 {
 	execv(prog, (char *const *)args);
 
-	ERR("failed to spawn %s\n", prog);
-	ERR("did you forget to build tools?\n");
+	TS_ERR("failed to spawn %s\n", prog);
+	TS_ERR("did you forget to build tools?\n");
 	return EXIT_FAILURE;
 }
 
@@ -192,7 +192,7 @@ static void test_export_environ(const struct test *t)
 
 		preload = malloc(total_len + sizeof(char));
 		if (preload == NULL) {
-			ERR("oom: test_export_environ()\n");
+			TS_ERR("oom: test_export_environ()\n");
 			return;
 		}
 
@@ -242,7 +242,7 @@ static inline int test_run_child(const struct test *t, int fdout[2], int fderr[2
 	if (t->output.out != NULL) {
 		close(fdout[0]);
 		if (dup2(fdout[1], STDOUT_FILENO) < 0) {
-			ERR("could not redirect stdout to pipe: %m\n");
+			TS_ERR("could not redirect stdout to pipe: %m\n");
 			return EXIT_FAILURE;
 		}
 	}
@@ -250,7 +250,7 @@ static inline int test_run_child(const struct test *t, int fdout[2], int fderr[2
 	if (t->output.err != NULL) {
 		close(fderr[0]);
 		if (dup2(fderr[1], STDERR_FILENO) < 0) {
-			ERR("could not redirect stderr to pipe: %m\n");
+			TS_ERR("could not redirect stderr to pipe: %m\n");
 			return EXIT_FAILURE;
 		}
 	}
@@ -263,18 +263,18 @@ static inline int test_run_child(const struct test *t, int fdout[2], int fderr[2
 		struct stat rootfsst, stampst;
 
 		if (stat(stamp, &stampst) != 0) {
-			ERR("could not stat %s\n - %m", stamp);
+			TS_ERR("could not stat %s\n - %m", stamp);
 			return EXIT_FAILURE;
 		}
 
 		if (stat(rootfs, &rootfsst) != 0) {
-			ERR("could not stat %s\n - %m", rootfs);
+			TS_ERR("could not stat %s\n - %m", rootfs);
 			return EXIT_FAILURE;
 		}
 
 		if (stat_mstamp(&rootfsst) > stat_mstamp(&stampst)) {
-			ERR("rootfs %s is dirty, please run 'meson compile testsuite/create-rootfs' before running this test\n",
-			    rootfs);
+			TS_ERR("rootfs %s is dirty, please run 'meson compile testsuite/create-rootfs' before running this test\n",
+			       rootfs);
 			return EXIT_FAILURE;
 		}
 	}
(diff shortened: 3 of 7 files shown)


evelikov:
As a much sought after bonus, the process of building and running the tests should be few seconds faster and couple of MB more efficient... If using RPi1 era hardware ;-)

/jk

tools/depmod.c
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 /*
- * Copyright (C) 2008  Alan Jenkins <alan-jenkins@tuffmail.co.uk>.
- * Copyright (C) 2011-2013  ProFUSION embedded systems
+ * Copyright (C) 2008 lan Jenkins <alan-jenkins@tuffmail.co.uk>.
```
lucasdemarchi: typo here

testsuite/init_module.c
```diff
@@ -32,8 +32,6 @@
 /* kmod_elf_get_section() is not exported, we need the private header */
 #include <libkmod/libkmod-internal.h>
 
-/* FIXME: hack, change name so we don't clash */
-#undef ERR
```
lucasdemarchi, line 36: I wonder if we should move the undef inside testsuite.h while it's defining TS_ERR() so ERR() is not used.

testsuite/test-blacklist.c
```diff
@@ -16,6 +16,7 @@
 
 /* good luck building a kmod_list outside of the library... makes this blacklist
  * function rather pointless */
```
lucasdemarchi: not that pointless since the assumption is that the list is coming from a probe or the like, not manually crafted as it's needed for a testsuite. Which brings the question: does it make sense to add the FIXME below?

testsuite/testsuite.h
```diff
@@ -136,8 +136,8 @@ int test_run(const struct test *t);
 #define DEFINE_TEST(_name, ...) DEFINE_TEST_WITH_FUNC(_name, _name, __VA_ARGS__)
 
 #define TESTSUITE_MAIN()                                                                 \
-	extern struct test __start_kmod_tests[] __attribute__((visibility("hidden")));   \
-	extern struct test __stop_kmod_tests[] __attribute__((visibility("hidden")));    \
+	extern const struct test __start_kmod_tests[];                                   \
```
lucasdemarchi, line 150: don't really remember why visibility hidden was there in the first place. It seems it was added by commit c5798fea9376 ("testsuite: use a section to put tests in instead of array"). Do you know why they are not needed anymore?

testsuite/testsuite.c
```diff
@@ -137,7 +137,7 @@ const struct test *test_find(const struct test *start, const struct test *stop,
 {
 	const struct test *t;
 
-	for (t = start; t < stop; t++) {
+	for (t = stop - 1; t >= start; t--) {
```
lucasdemarchi: not needed since it will only find 1 test anyway. Instead, I think we should reverse the order in TESTSUITE_MAIN() so when running the all the tests they are not in reversed order.

Another thing to consider: the order is only stable in the same compilation unit and may vary from compiler to compiler although afair both gcc and clang use  the same order (reversed defined order). See https://lore.kernel.org/all/fsleccud43leoioli6bj4xoydwbdcjb5qr4denuf6morhgtqki@t6xya2mygon5/

testsuite/test-shared.c
```diff
@@ -186,4 +189,275 @@ static int test_array_pop(void)
 
 DEFINE_TEST(test_array_pop, .description = "test array pop");
 
+/* hash sub-group */
```
lucasdemarchi, line 200: the fact that you need this probably means it should be separate ;). I think the grouping/visibility would be accomplished by naming them test-shared-hash.c, test-shared-array.c, etc.

But.... I don't care much about this and if you think it's better this way, so be it.

testsuite/test-shared.c
```diff
@@ -650,4 +657,283 @@ static int xfail_strbuf_shrink_to(void)
...
+			printf("len     %zu\n", len);
+			printf("output  %s\n", buf);
+		}
+
+		printf("\n");
+	}
+
+	return 0;
+}
+DEFINE_TEST(alias_1,
```
lucasdemarchi, line 694: should we have a util_ prefix?

testsuite/meson.build
```diff
@@ -80,8 +80,8 @@ _testsuite = [
   'test-blacklist',
   'test-dependencies',
   'test-depmod',
-  'test-init',
   'test-initstate',
+  'test-libkmod',
```
lucasdemarchi, line 81: should we also rename the TC_ROOTFS paths for them?

lucasdemarchi (commented):
Started reviewing this... done until commit "testsuite: rename test-init -> test-libkmod". Thanks.

tools/depmod.c
lucasdemarchi: it also looks like there are some double spaces in testsuite copyright notes

scripts/setup-rootfs.sh
```diff
@@ -78,6 +78,7 @@ map=(
     ["test-dependencies$MODULE_DIRECTORY/4.0.20-kmod/kernel/fs/"]="mod-foo.ko"
     ["test-init/"]="mod-simple.ko"
     ["test-remove/"]="mod-simple.ko"
+    ["test-remove2/"]="mod-simple.ko"
```
lucasdemarchi: As in previous comment, should it be test-libkmod-remove/?

testsuite/test-libkmod.c
```diff
@@ -132,4 +135,39 @@ DEFINE_TEST(
...
+	kmod_unref(ctx);
+
+	return 0;
+}
+DEFINE_TEST(test_remove2,
+	    .description = "test if libkmod's delete_module removes module directory",
+	    .config = {
+		    [TC_ROOTFS] = TESTSUITE_ROOTFS "test-remove2/",
+		    [TC_INIT_MODULE_RETCODES] = "",
+		    [TC_DELETE_MODULE_RETCODES] = "mod_simple:0:0" STRINGIFY(ENOENT),
```
lucasdemarchi, line 179: humn.... pre-existing bug? should be `"mod_simple:0:0:" STRINGIFY(ENOENT)`, with the extra `:`, like in test-init.c

From a quick look it seems this STRINGIFY() is just wrong though as it doens't seem to be used in parse_retcodes() and will rather parse that as a module name :-o. In that function errno is parsed as number, not as string.

tools/depmod.c
evelikov: As-in more than the ones fixed - do you have a pointer if so?

I've intentionally left the code motion changes on their own, so that git diff/log nicely flags them as such.

testsuite/init_module.c
evelikov: It feels a bit odd to do so, if they're no longer clashing. Plus is we resolve the blacklist `include ...libkmod-internal.h` instance, we can convert linking to shared libkmod (kmod-list will still need static one).

Hence, there will be nobody providing the macro in the first place.

testsuite/test-blacklist.c
evelikov: The comment is moved so wether it makes sense, it's not up-to me.

Although, as mentioned above - it would be good to have the internal symbol dropped and linkage resolved. The test is supposed to cover the public API and as such should be using `libkmod.so`.

testsuite/testsuite.h
evelikov: No idea honestly... It may be that an older compiler/linker version wasn't flagging them as hidden (or protected) due to the special section handling?

I know there was an issue(?) with weak symbols being exported, regardless of `-fvisibilidy=hidden`.

testsuite/testsuite.c
evelikov: Hmm flipping the order in the caller means that we're passing `stop, start` named variables, while the functions themselves use `start, stop`. Guess I'll update both places for consistency sake.

And yes, the order is compiler/linker dependant but in practise it's consistent. Will add a note in the commit message.

testsuite/test-shared.c
evelikov: While moving things over, my initial thought was about adding test groups to the runner. Although that means extra code, hence maintenance, etc.

Upon glancing at the shared tests themselves - they seem to be redundantly extensive and somewhat opposite to the rest. Quite seriously itching to chop about 1/3, without affecting the coverage. Although I've kept that for another PR, since this is long enough as-is.

testsuite/test-shared.c
evelikov: Generally yes, but I've left the renaming for another PR.

Don't want to make git (and hence future archaeologists) sad* when it's not a simple code motion.

testsuite/meson.build
evelikov: :facepalm: not sure how I missed that - hell yes.

scripts/setup-rootfs.sh
evelikov: Ideally not, because:
 - both tests have bugs - one you're already spotted
 - both should really be merged

testsuite/test-libkmod.c
evelikov: Yes, although the whole thing should probably be made more robust - see https://github.com/kmod-project/kmod/issues/352.

Edit: planning to fix both remove tests and merge them into one as a follow-up.

tools/depmod.c
lucasdemarchi: to be clear, first comment was about the inadvertent `s/Alan/lan/`.

For the second comment, TL;DR: ignore :). My bad... I started reviewing this in github web UI and thought this was one of the first commits, not one of the last commits. Then while reviewing the the intermediary commits I noticed the double spaces.

testsuite/test-blacklist.c
lucasdemarchi: The real point is:

1) We don't want to export kmod_list as there's no reason for a real user of libkmod to craft its own list
2) Testsuite needs to craft a list just to check if the result is correct in a more "unit" testing way

tools/depmod.c
evelikov: No worries and thanks for spotting the (A)lan mistake. Will fix.

testsuite/test-blacklist.c
evelikov: Fully agreed on the first point.

On the second... I hope we can use some other API which gives us a list to (re)use. If that's not possible, that brings a bigger question - how is the API usable for anyone outside of kmod itself. But overall not something to worry about, yet.

lucasdemarchi: the list comes from one of the APIs. You then compose it with apply_filter() action an get a new list:

```
kmod_module_new_from_lookup(ctx, "foobar", &modlist);
kmod_module_apply_filter(ctx, KMOD_FILTER_BLACKLIST, modlist, &filtered_modlist);
kmod_list_foreach(itr, filtered_modlist) {
    mod = kmod_module_get_module(itr);
    printf("%s\n", kmod_module_get_name(ctx, mod);
    kmod_module_unref(mod);
}
```

testsuite/testsuite.c
lucasdemarchi: For TESTSUITE_MAIN() the point is that it **executes** the tests in the same order given by --list. Inverting all of them should be fine IMO.

lucasdemarchi:
Just finished reviewing the rest and they lgtm. Thanks

lucasdemarchi:
Applied, thanks

PR #451, opened 2026-07-26
Modinfo rework

### LINKED ISSUE #434: modinfo uses mixed delimiters in --null mode
If you use `modinfo --null` in an attempt to use it as a machine-readable output there are multiple delimiters in use, specifically whilst for most fields the output changes to a concise `field=value\0` output, the first line is still `filename:        foo.mk`.

evelikov:
Here's a series addressing most of the formatting issues raised in https://github.com/kmod-project/kmod/issues/434. In particular the formatting code is unified and beaten into shape to use consistent formatting. With the outstanding bits being:

- the module description string is free-form and can have misc whitespace characters - `\t`, `\n` et al
- the `signature` value is still has multi-line/wrapped/tabbed formatting

In all fairness, I'm inclined to leave those as-is and let other handle them in a future PR.

The first 3 commits are something we'd probably want for the stable 34.x branch assuming there will be another release. Although, in that case we should probably check the other PRs for similar non-invasive fixes.

For the PR itself - it mostly small prep and fixup patches with the most sizable change being `modinfo: rework parm handling`.

Resolves: https://github.com/kmod-project/kmod/issues/434

### DIFF
--- a/shared/macro.h
+++ b/shared/macro.h
@@ -31,6 +31,9 @@
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]) + _array_size_chk(arr))
 
+#define _ALIGN_MASK(x, _MASK) (((x) + (_MASK)) & ~(_MASK))
+#define ALIGN(x, B) _ALIGN_MASK(x, (typeof(x))(B) - 1)
+
 #define XSTRINGIFY(x) #x
 #define STRINGIFY(x) XSTRINGIFY(x)
 
--- a/tools/modinfo.c
+++ b/tools/modinfo.c
@@ -187,16 +187,18 @@ static int modinfo_do(struct kmod_module *mod)
 	if (is_builtin) {
 		if (field == NULL)
 			print_line("name", kmod_module_get_name(mod));
-		else if (field != NULL && streq(field, "name"))
+		else if (streq(field, "name")) {
 			print_line(NULL, kmod_module_get_name(mod));
+			return 0;
+		}
 		filename = "(builtin)";
 	}
 
-	if (field != NULL && streq(field, "filename")) {
+	if (field == NULL)
+		print_line("filename", filename);
+	else if (streq(field, "filename")) {
 		print_line(NULL, filename);
 		return 0;
-	} else if (field == NULL) {
-		print_line("filename", filename);
 	}
 
 	err = kmod_module_get_info(mod, &list);
@@ -223,10 +225,11 @@ static int modinfo_do(struct kmod_module *mod)
 		const char *value = kmod_module_info_get_value(l);
 
 		if (field != NULL) {
-			if (streq(field, key)) {
+			if (streq(field, key))
 				print_line(NULL, value);
-			}
-		} else if (streq(key, "parm")) {
+			continue;
+		}
+		if (streq(key, "parm")) {
 			err = process_parm(parm_desc, value, &params);
 			if (err < 0)
 				goto end;
--- a/tools/modinfo.c
+++ b/tools/modinfo.c
@@ -28,19 +28,19 @@ static const char *field;
 struct param {
 	struct param *next;
 	const char *name;
-	const char *param;
+	const char *desc;
 	const char *type;
 	int namelen;
-	int paramlen;
+	int desclen;
 	int typelen;
 };
 
-static int add_param(const char *name, size_t namelen, const char *param, size_t paramlen,
+static int add_param(const char *name, size_t namelen, const char *desc, size_t desclen,
 		     const char *type, size_t typelen, struct param **list)
 {
 	struct param *it;
 
-	if (namelen > INT_MAX || paramlen > INT_MAX || typelen > INT_MAX) {
+	if (namelen > INT_MAX || desclen > INT_MAX || typelen > INT_MAX) {
 		return -EINVAL;
 	}
 
@@ -57,15 +57,15 @@ static int add_param(const char *name, size_t namelen, const char *param, size_t
 		*list = it;
 		it->name = name;
 		it->namelen = namelen;
-		it->param = NULL;
+		it->desc = NULL;
 		it->type = NULL;
-		it->paramlen = 0;
+		it->desclen = 0;
 		it->typelen = 0;
 	}
 
-	if (param != NULL) {
-		it->param = param;
-		it->paramlen = paramlen;
+	if (desc != NULL) {
+		it->desc = desc;
+		it->desclen = desclen;
 	}
 
 	if (type != NULL) {
@@ -78,8 +78,8 @@ static int add_param(const char *name, size_t namelen, const char *param, size_t
 
 static int process_parm(const char *key, const char *value, struct param **params)
 {
-	const char *name, *param, *type;
-	size_t namelen, paramlen, typelen;
+	const char *name, *desc, *type;
+	size_t namelen, desclen, typelen;
 	const char *colon = strchr(value, ':');
 	int ret;
 
@@ -96,18 +96,18 @@ static int process_parm(const char *key, const char *value, struct param **param
 	name = value;
 	namelen = colon - value;
 	if (streq(key, "parm")) {
-		param = colon + 1;
-		paramlen = strlen(param);
+		desc = colon + 1;
+		desclen = strlen(desc);
 		type = NULL;
 		typelen = 0;
 	} else {
-		param = NULL;
-		paramlen = 0;
+		desc = NULL;
+		desclen = 0;
 		type = colon + 1;
 		typelen = strlen(type);
 	}
 
-	ret = add_param(name, namelen, param, paramlen, type, typelen, params);
+	ret = add_param(name, namelen, desc, desclen, type, typelen, params);
 	if (ret < 0) {
 		ERR("Unable to add parameter: %s\n", strerror(-ret));
 		return -ENOMEM;
@@ -137,14 +137,14 @@ static int modinfo_params_do(const struct kmod_list *list)
 		struct param *p = params;
 		params = p->next;
 
-		if (p->param == NULL)
+		if (p->desc == NULL)
 			printf("%.*s: (%.*s)%c", p->namelen, p->name, p->typelen, p->type,
 			       separator);
 		else if (p->type != NULL)
-			printf("%.*s:%.*s (%.*s)%c", p->namelen, p->name, p->paramlen,
-			       p->param, p->typelen, p->type, separator);
+			printf("%.*s:%.*s (%.*s)%c", p->namelen, p->name, p->desclen,
+			       p->desc, p->typelen, p->type, separator);
 		else
-			printf("%.*s:%.*s%c", p->namelen, p->name, p->paramlen, p->param,
+			printf("%.*s:%.*s%c", p->namelen, p->name, p->desclen, p->desc,
 			       separator);
 
 		free(p);
@@ -234,15 +234,15 @@ static int modinfo_do(struct kmod_module *mod)
 		struct param *p = params;
 		params = p->next;
 
-		if (p->param == NULL)
+		if (p->desc == NULL)
 			printf("%-16s%.*s: (%.*s)%c", "parm:", p->namelen, p->name,
 			       p->typelen, p->type, separator);
 		else if (p->type != NULL)
 			printf("%-16s%.*s:%.*s (%.*s)%c", "parm:", p->namelen, p->name,
-			       p->paramlen, p->param, p->typelen, p->type, separator);
+			       p->desclen, p->desc, p->typelen, p->type, separator);
 		else
 			printf("%-16s%.*s:%.*s%c", "parm:", p->namelen, p->name,
-			       p->paramlen, p->param, separator);
+			       p->desclen, p->desc, separator);
 
 		free(p);
 	}
--- a/tools/modinfo.c
+++ b/tools/modinfo.c
@@ -176,6 +176,8 @@ end:
 
 static int modinfo_do(struct kmod_module *mod)
 {
+	const bool print_all = field == NULL;
+	const bool print_parm = !print_all && streq(field, "parm");
 	struct kmod_list *l, *list = NULL;
 	struct param *params = NULL;
 	int err, is_builtin;
@@ -185,7 +187,7 @@ static int modinfo_do(struct kmod_module *mod)
 
 	/* TODO: align builtin vs not wrt listing "name:" via kmod_module_get_info() */
 	if (is_builtin) {
-		if (field == NULL)
+		if (print_all)
 			print_line("name", kmod_module_get_name(mod));
 		else if (streq(field, "name")) {
 			print_line(NULL, kmod_module_get_name(mod));
@@ -194,7 +196,7 @@ static int modinfo_do(struct kmod_module *mod)
 		filename = "(builtin)";
 	}
 
-	if (field == NULL)
+	if (print_all)
 		print_line("filename", filename);
 	else if (streq(field, "filename")) {
 		print_line(NULL, filename);
@@ -215,7 +217,7 @@ static int modinfo_do(struct kmod_module *mod)
 		return err;
 	}
 
-	if (field != NULL && streq(field, "parm")) {
+	if (print_parm) {
 		err = modinfo_params_do(list);
 		goto end;
 	}
@@ -224,7 +226,7 @@ static int modinfo_do(struct kmod_module *mod)
 		const char *key = kmod_module_info_get_key(l);
 		const char *value = kmod_module_info_get_value(l);
 
-		if (field != NULL) {
+		if (!print_all) {
 			if (streq(field, key))
 				print_line(NULL, value);
 			continue;
@@ -242,7 +244,7 @@ static int modinfo_do(struct kmod_module *mod)
 		}
 	}
 
-	if (field != NULL)
+	if (!print_all)
 		goto end;
 
 	while (params != NULL) {
--- a/tools/modinfo.c
+++ b/tools/modinfo.c
@@ -88,6 +88,11 @@ static int process_parm(const char *key, const char *value, struct param **param
 		return 0;
 	}
 
+	if (colon == value) {
+		ERR("Missing param name in value \"%s\"\n", value);
+		return 0;
+	}
+
 	name = value;
 	namelen = colon - value;
 	if (streq(key, "parm")) {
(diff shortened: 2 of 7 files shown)


evelikov:
The latest revision has all the yaks shaved on my end ;-)

v2:
 - drop calloc commit
 - enum names `p_parm{desc,type}` -> `parm_{desc,type]`
 - find matching or empty entry in a single loop (add_param)
 - match up-to UINT_MAX parm/parmtype entries
 - drop unneeded INT_MAX checks for strlen(value)
 - don't return success if we fail to [cm]alloc
 - kill off struct param::{desc,type}len
 - rework parm string length validation

Edit: this time for real - aka rebase + clang-format nitpick

evelikov:
NOTE: this effectively prints the "parm" entries in inverse order... AFAICT that should not be an issue, since the output order was never *meant to be* deterministic.

tools/modinfo.c
```diff
@@ -213,58 +166,96 @@ static int modinfo_do(struct kmod_module *mod)
...
+
+		/* XXX: do we want to emit a warning/error here? */
+		params_count = (params_count > UINT_MAX) ? UINT_MAX : (unsigned int)count;
+
+		params = calloc(params_count, sizeof(*params));
+		/*
+		 * The "name:" exists in both parm&parmtype, so don't worry if we
+		 * overallocate.
+		 */
+		sbuf = malloc(longest_desc + longest_type + strlen(" ()"));
```
lucasdemarchi: missing + 1 here for the '\0'. Also, I'd use a real sbuf here with size like 256.

lucasdemarchi: I mean... loop to the get the biggest size and then strbuf_reserve().

tools/modinfo.c
```diff
@@ -213,58 +166,96 @@ static int modinfo_do(struct kmod_module *mod)
...
-		params = p->next;
+	for (unsigned int i = 0; i < params_count; i++) {
+		struct param *p = &params[i];
+		if (p->name == NULL)
+			continue;
 
 		if (p->type != NULL)
-			printf("%-16s%.*s:%.*s (%.*s)%c", "parm:", p->namelen, p->name,
-			       p->desclen, p->desc, p->typelen, p->type, separator);
+			sprintf(sbuf, "%.*s:%.*s (%.*s)", p->namelen, p->name, p->desclen,
```
lucasdemarchi: as above... here I think it would miss the +1 space for \0 for the longest option. That would only occur of course if the longest one is also one without a description.

tools/modinfo.c
```diff
@@ -156,19 +156,18 @@ static int modinfo_params_do(const struct kmod_list *list)
 
 static int modinfo_do(struct kmod_module *mod)
 {
+	const enum kmod_module_initstate state = kmod_module_get_initstate(mod);
+	const bool is_builtin = state == KMOD_MODULE_BUILTIN;
+	const char *filename = is_builtin ? "(builtin)" : kmod_module_get_path(mod);
```
lucasdemarchi: I think we need to be careful here... we may be running modinfo() into a module that is builtin for the current kernel, but we are actually doing modinfo for something else.

Example: ext4 builtin on my distro, but I'm doing `modinfo ./ext4.ko`  for a kernel I'm developing for another system. Previously it would notice that `mod` got constructed from a module and use that as filename. Here you'd wrongly report that it is builtin.

tools/modinfo.c
```diff
@@ -192,60 +149,100 @@
...
+
+				count++;
+
+				if (len > longest_type)
+					longest_type = len;
+			}
+		}
+
+		/* XXX: do we want to emit a warning/error here? */
+		params_count = (params_count > UINT_MAX) ? UINT_MAX : (unsigned int)count;
```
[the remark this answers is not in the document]
lucasdemarchi: agreed. You did `unsigned int params_count = 0;`  above and there's nothing possibly changing it above.

tools/modinfo.c
```diff
@@ -213,58 +166,96 @@ static int modinfo_do(struct kmod_module *mod)
...
+
+		params = calloc(params_count, sizeof(*params));
```
lucasdemarchi: params_count is possibly 0.

tools/modinfo.c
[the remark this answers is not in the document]
evelikov: Silly typo there. Fixed.

tools/modinfo.c
evelikov: Converted to reserve, which takes care of the `\0` internally ... modulo a small bug or two. Fix included, might be worth picking individually for 34.x.

tools/modinfo.c
evelikov: Valid point - dropped for now.

As a whole, a few functions rely on parsing /sys/modules and friends even when a different kernel version is explicitly provided to `kmod_ctx_new`. The current behaviour seems suboptimal/buggy.

IMHO we should probably error out, although I could be missing something?

tools/modinfo.c
evelikov: Fixed, thanks.

evelikov:
v3:
 - scope variable declaration
 - code dedup in parm/parmtype paths
 - use correct counter
 - avoid allocations on 0 parm(s)
 - remove param::{desc,type}len
 - move parm* len validation further up the call stack

 - make strbuf_reserve_extra public again + reintroduce tests
 - fix overflow in ^^
 - use ALIGN in ^^
 - remove strbuf_init -

tools/modinfo.c
evelikov: Somewhat related:

It looks like we have different filename (path really) in database vs filesystem... without ever flagging it as an issue. Observed this while copying/adding more modinfo tests. To make it more interesting, `setup-rootfs.sh` manually changes the path in `.dep` files but not in `.dep.bin` and co.

One possible (perhaps overly comprehensive) solution is:
 - have the proper depmod artefacts checked-in
 - on test prep, re-generate ^^ into a throwaway location and error on miss-match
 - ensure correct path is used across .dep/.dep.bin/etc - might need a patch to libkmod/depmod/
 - on test closure, flag files with unchanged atime - ignoring the depmod regen/check

What do you think?

tools/modinfo.c
```diff
@@ -192,60 +149,109 @@ static int modinfo_do(struct kmod_module *mod)
...
+		struct param *p = &params[i];
+		if (p->name == NULL)
+			continue;
 
-	while (params != NULL) {
-		struct param *p = params;
-		params = p->next;
+		strbuf_pushmem(&buf, p->name, p->namelen);
+		strbuf_pushchar(&buf, ':');
+		strbuf_pushchars(&buf, p->desc);
```
evelikov: This should be `strbuf_pushchars(&buf, p->desc ? p->desc : "");`

libkmod/libkmod-builtin.c
```diff
@@ -130,13 +130,14 @@ static char **strbuf_to_vector(struct strbuf *buf, size_t count)
 	    uaddsz_overflow(strbuf_used(buf), vec_size, &total_size))
 		return NULL;
 
+	/* XXX: Do not handle bytes behind the object's back */
```
lucasdemarchi: and we had strbuf_steal() for this kind of situation that got removed.  IMO let's not drop these helpers if they may be used in future?

tools/modinfo.c
```diff
@@ -213,58 +167,106 @@ static int modinfo_do(struct kmod_module *mod)
...
-			printf("%-16s%.*s:%.*s%c", "parm:", p->namelen, p->name,
-			       p->desclen, p->desc, separator);
+		if (p->type != NULL) {
+			strbuf_pushchars(&buf, " (");
+			strbuf_pushmem(&buf, p->type, p->typelen);
+			strbuf_pushchars(&buf, ")");
+		}
 
-		free(p);
+		print_line(print_parm ? NULL : "parm", strbuf_str(&buf));
```
lucasdemarchi, line 251: missing `strbuf_clear()` here or at he beginning of the loop? Otherwise parm is just accumulated and I got several errors in the output.

```
$ ./build/modinfo xe
...
parm:           wedged_mode:Module's default policy for the wedged mode (0=never, 1=upon-critical-error, 2=upon-any-hang-no-reset [default=upon-critical-error]) (uint)max_vfs:Limit number of Virtual Functions (VFs) that could be managed. (0=no VFs; N=allow up to N VFs [default=unlimited]) (uint)force_probe:Force probe options for specified devices. See CONFIG_DRM_XE_FORCE_PROBE for details [default=]) (charp)gsc_firmware_path:GSC firmware path to use instead of the default one - empty string disables (charp)huc_firmware_path:HuC firmware path to use instead of the default one - empty string disables (charp)guc_firmware_path:GuC firmware path to use instead of the default one (charp)guc_log_level:GuC firmware logging level (0=disable, 1=normal, 2..5=verbose-levels [default=1]) (int)vram_bar_size:Set the vram bar size in MiB (<0=disable-resize, 0=max-needed-size, >0=force-size [default=0]) (int)probe_display:Probe display HW, otherwise it's left untouched [default=true]) (bool)force_execlist:Force Execlist submission (bool)svm_notifier_size:Set the svm notifier size in MiB, must be power of 2 [default=512] (uint)enable_dmc_wl:Enable DMC wakelock (-1=use per-chip default, 0=disabled, 1=enabled, 2=match any register, 3=always locked) Default: -1 (int)enable_psr2_sel_fetch:Enable PSR2 and Panel Replay selective fetch (0=disabled, 1=enabled) Default: 1 (bool)psr_safest_params:Replace PSR VBT parameters by the safest and not optimal ones. This is helpful to detect if PSR issues are related to bad values set in  VBT. (0=use VBT parameters, 1=use safest parameters)Default: 0 (bool)enable_panel_replay:Enable Panel Replay (0=disabled, 1=enabled). Default: -1 (use per-chip default) (int)enable_psr:Enable PSR (0=disabled, 1=enable up to PSR1, 2=enable up to PSR2) Default: -1 (use per-chip default) (int)enable_fbc:Enable frame buffer compression for power savings (default: -1 (use per-chip default)) (int)enable_dp_mst:Enable multi-stream transport (MST) for new DisplayPort sinks. (default: true) (bool)nuclear_pageflip:Force enable atomic functionality on platforms that don't have full support yet. (bool)verbose_state_checks:Enable verbose logs (ie. WARN_ON()) in case of unexpected hw state conditions. (bool)disable_display:Disable display (default: false) (bool)force_reset_modeset_test:Force a modeset during gpu reset for testing (default:false). For developers only. (bool)load_detect_test:Force-enable the VGA load detect code for testing (default:false). For developers only. (bool)enable_dpcd_backlight:Enable support for DPCD backlight control(-1=use per-VBT LFP backlight type setting [default], 0=disabled, 1=enable, 2=force VESA interface, 3=force Intel interface) (int)edp_vswing:Ignore/Override vswing pre-emph table selection from VBT (0=use value from vbt [default], 1=low power swing(200mV),2=default swing(400mV)) (int)invert_brightness:Invert backlight brightness (-1 force normal, 0 machine defaults, 1 force inversion), please report PCI device ID, subsystem vendor and subsystem device ID to dri-devel@lists.freedesktop.org, if your machine needs it. It will then be included in an upcoming module version. (int)enable_ips:Enable IPS (default: true) (bool)disable_power_well:Disable display power wells when possible (-1=auto [default], 0=power wells always on, 1=power wells disabled when possible) (int)enable_sagv:Enable system agent voltage/frequency scaling (SAGV) (default: true) (bool)enable_flipq:Enable DMC flip queue (default: false) (bool)enable_dsb:Enable display state buffer (DSB) (default: true) (bool)enable_dpt:Enable display page table (DPT) (default: true) (bool)enable_dc:Enable power-saving display C-states. (-1=auto [default]; 0=disable; 1=up to DC5; 2=up to DC6; 3=up to DC5 with DC3CO; 4=up to DC6 with DC3CO) (int)vbt_sdvo_panel_type:Override/Ignore selection of SDVO panel mode in the VBT (-2=ignore, -1=auto [default], index in VBT BIOS table) (int)panel_use_ssc:Use Spread Spectrum Clock with panels [LVDS/eDP] (default: auto from VBT) (int)lvds_channel_mode:Specify LVDS channel mode (0=probe BIOS [default], 1=single-channel, 2=dual-channel) (int)vbt_firmware:Load VBT from specified file under /lib/firmware (charp)dmc_firmware_path:DMC firmware path to use instead of the default one. Use /dev/null to disable DMC and runtime PM. (charp)
```

shared/macro.h
```diff
@@ -31,6 +31,8 @@
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]) + _array_size_chk(arr))
 
+#define ALIGN(A, B) (((A) + (B) - 1) & ~((B) - 1))
```
lucasdemarchi: this doesn't work if sizeof(B) < sizeof(A). E.g. typeof(A) == uint64_t and typeof(B) == uint32_t.

Safer to handle similarly to the kernel.... untested, but something like:

```
#define __ALIGN_MASK(x, mask) (((x) + (mask)) & ~(mask))
#define ALIGN(x, A) __ALIGN_MASK(x, (typeof(x))(a) - 1)
```

.... or simply casting B to typeof(A) so we guarantee they have the same width.

tools/modinfo.c
```diff
@@ -253,11 +247,11 @@ static int modinfo_do(struct kmod_module *mod)
 
 		strbuf_pushmem(&buf, p->name, p->namelen);
 		strbuf_pushchar(&buf, ':');
-		strbuf_pushmem(&buf, p->desc, p->desclen);
+		strbuf_pushchars(&buf, p->desc);
```
lucasdemarchi: p->desc can be NULL. `strbuf_pushmem(..., NULL, 0)` was safe, but `strbuf_pushchars(&buf, p->desc)` is not due to the strlen(p->desc) implicit call.

tools/modinfo.c
lucasdemarchi: or just check for NULL... just saw this comment after reviewing. See comments below.

lucasdemarchi:
Also getting this:

```
=================================================================
==195428==ERROR: AddressSanitizer: attempting double-free on 0x6e5cad5e03c0 in thread T0:
    #0 0x712caf32a3ff in free ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:51
    #1 0x64011a76b0f4 in modinfo_do ../tools/modinfo.c:109
```

Since it's now using DECLARE_STRBUF_WITH_STACK(), we need to remove the strbuf_release() at the end.

Added 2 fixup commits to handle the comments.

shared/macro.h
zhiwei-arbo: Done — dropped the block above `model_config` and left two lines on `type` itself. The `extra="ignore"` half went entirely; the docstring already frames the whole contract as tolerant, so restating it at the config was the redundant part.

tools/modinfo.c
evelikov: Sorry - completely forgot that my extra modinfo tests aren't upstream... So I got completely lazy with the manual testing.

libkmod/libkmod-builtin.c
evelikov: Looking at the PR that removed it:
 - there is one overflow bug (fixed in this PR)
 - the function comment/description didn't get updated
 - the above workaround was pre-existing

Can reintroduce and fix ^^ as a follow up PR.

Would you be OK with dropping "strbuf_init` or prefer to keep it - I don't mind either way.

evelikov:
> Added 2 fixup commits to handle the comments.

The two extra fixups look good to me. Thanks o/

shared/macro.h
evelikov: @zhiwei-arbo I'm suspecting the comment was meant for another project/PR ;-)

libkmod/libkmod-builtin.c
lucasdemarchi: > Can reintroduce and fix ^^ as a follow up PR.

sounds good

> Would you be OK with dropping "strbuf_init` or prefer to keep it - I don't mind either way.

we can drop the callers from the codebase, but let's keep it in the more-or-less-self-contained strbuf implementation. I'll push an update here before merging.

As a followup I think it's also reasonable to move strbuf_to_vector() to be in strbuf.c so the re-init happens inside the strbuf implementation. Then we document that sbuf becomes invalid afte the call.

lucasdemarchi:
double checking this commit:
- modinfo: omit param::desc == NULL case

> "printf("%.s", 0, NULL) is a perfectly valid use."

it actually isn't. Passing NULL here is undefined behavior regardless of the precision specifier. In the end this code was converted to strbuf so we don't need it. I will simply drop the commit and solve the conflicts.

evelikov:
> double checking this commit:
> 
>     * modinfo: omit param::desc == NULL case
> 
> 
> > "printf("%.s", 0, NULL) is a perfectly valid use."
> 
> it actually isn't. Passing NULL here is undefined behavior regardless of the precision specifier. In the end this code was converted to strbuf so we don't need it. I will simply drop the commit and solve the conflicts.

May I ask where did you spot that?

The man page suggests that up-to precision specifier characters of the string are accessed/printed. Plus the undefined sanitizer included in gcc (16.1.1 20260725) doesn't complain either.

Either way, thanks o/

lucasdemarchi:
There's some discussion here about that part of the standard: https://stackoverflow.com/questions/70499706/is-it-guaranteed-by-the-c-standard-to-be-safe-to-do-printf-s-0-null

lucasdemarchi:
Applied, thanks

evelikov:
It seems like v2 got merged, including the UB `printf("%.s", 0, NULL)` piece.

I'm assuming that's by mistake - will send an update PR tomorrow/day after, with all of your comments addressed.

lucasdemarchi:
ugh... sorry. That's what I get for working in 2 different computers :-/.

lucasdemarchi:
I removed the commits from the master branch and will re-add once I get my other computer back. Sorry about that.

lucasdemarchi:
Ok, now it's hopefully the right branch pushed here. @evelikov can you double check?

evelikov:
> I removed the commits from the master branch and will re-add once I get my other computer back. Sorry about that.

No problem - sh*t happens. Would prefer if we don't rewrite the branch history but revert/fixup instead. The former erodes the distro maintainers' trust, et al.

The updated branch looks great. Thanks

lucasdemarchi:
Applied, thanks. On the force push side, IMO both options were bad and I weighed one more than the other: most people (including distro maintainers) will not see the force push or be affected by it if the time span is short. Anyway, not something to do frequently or at all going forward.

PR #453, opened 2026-08-02
Various getopt nitpicks

evelikov:
A small series that started off with "hey, let me add a test for modinfo --invalid" and ended up homogenising the getopt usage in-tree.


evelikov:
v2:
 - clang-format tweaks and commit message polish in last commit

tools/kmod.c
```diff
@@ -15,7 +15,7 @@
 
 #include "kmod.h"
 
-static const char options_s[] = "+hV";
+static const char options_s[] = "hV";
```
lucasdemarchi: this is used for  commands designed as the `kmod` tool which includes subcommands. Think of it like `git -c foobar=true log --format=%h`, i.e. there are options for both the command and the subcommand.

Right now it doesn't make much difference because all the options we have,, "hV", exit the program when parsed.

lucasdemarchi: I dropped this and rebased to let CI hopefully succeed now on Alpine.

lucasdemarchi:
Applied, thanks

tools/kmod.c
evelikov: Makes sense. Thanks for the explanation.

PR #454, opened 2026-08-03
Try to fix/workaround build on alpine

lucasdemarchi:


lucasdemarchi:
Dropped the first commit... it could be further polished. Merged the fixes

.github/dev-container.py
```diff
@@ -0,0 +1,452 @@
...
+from typing import Any
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+DEFAULT_WORKFLOW = REPO_ROOT / ".github" / "workflows" / "main.yml"
+DEFAULT_SETUP_OS = REPO_ROOT / ".github" / "actions" / "setup-os"
+DEFAULT_PREFIX = "localhost/kmod-ci"
+DEFAULT_IMAGE_ENV = ["KDIR=any"]
+
+MATRIX_EXPR_RE = re.compile(r"^\s*\$\{\{\s*matrix\.([A-Za-z_][A-Za-z0-9_-]*)\s*\}\}\s*$")
```
evelikov, line 26: Slightly worried that (m)any yml changes might break the RE above. Perhaps we could wire some simple check to CI to catch issues?

.github/dev-container.py
```diff
@@ -0,0 +1,452 @@
...
+
+echo "Distro: ${DISTRO}"
+printf 'DISTRO=%s\n' "${DISTRO}" >> "${GITHUB_ENV}"
+
+chmod +x "${GITHUB_ACTION_PATH}"/setup-*.sh
+setup_script="${GITHUB_ACTION_PATH}/setup-${DISTRO}.sh"
+if [ ! -x "${setup_script}" ]; then
+    echo "Missing setup script: ${setup_script}" >&2
+    exit 1
+fi
```
evelikov, line 47: Most of the above bits are from `setup-os/action.yml`. Would be great if they can be fleshed out and reused.

.github/dev-container.py
```diff
@@ -0,0 +1,452 @@
...
+"""
+
+
+@dataclass(frozen=True)
+class ContainerImage:
+    source: str
+    local: str
+
+
+def unique(items: list[str]) -> list[str]:
```
evelikov, line 59: Admittedly, I didn't follow exactly how the script operates, but it feels like there should be no duplicates while parsing the yml.

.github/dev-container.py
```diff
@@ -0,0 +1,452 @@
...
+            continue
+        seen.add(item)
+        values.append(item)
+
+    return values
+
+
+def strip_yaml_scalar(value: str) -> str | None:
+    value = value.strip()
+    if not value or value in {"|", ">"}:
```
evelikov, line 74: Is there some yaml library that we can use? Or the worst case scenario we can go for a shell script with `yq` invocations.

For example: `yq .jobs.build.strategy.matrix.include .github/workflows/main.yml` gives us a much more readable command and expected output.

evelikov (commented):
Overall I like the idea to have simple script that lists the containers combos and pulls/sets one up for us.

I left a few ideas, for the future - feel free to borrow any that makes sense.

Also: thanks for sorting out Alpine. Will try to report this upstream, if you haven't already.

lucasdemarchi:
> Overall I like the idea to have simple script that lists the containers combos and pulls/sets one up for us.
> 
> I left a few ideas, for the future - feel free to borrow any that makes sense.

it will need some polish/rewrite. It was not something intended for review yet... since I needed to reproduce the alpine issue I just asked  gemini to code this so I could quickly iterate through the containers. In the end I think the interface ended up a little cumbersome as it would still run the setup as entrypoint, which was not exactly what I wanted.

PR #455, opened 2026-08-06
ci: Fix html files not being built

lucasdemarchi:
Also (hopefully) allow to generate docs for past releases.


.github/workflows/docs.yml
```diff
@@ -29,8 +29,8 @@ jobs:
 
       - name: Build docs
         run: |
-          meson setup -Ddocs=true builddir/
-          meson compile -C builddir/
+          meson setup --native-file build-dev.ini -Ddocs=true build-docs/
```
evelikov, line 53: OOC: I'm guessing the `--native-file` is for consistency sake?

.github/workflows/docs.yml
```diff
@@ -34,17 +42,24 @@ jobs:
 
       - name: Extract docs version
         shell: bash
+        env:
+          INPUT_TAG: ${{ inputs.tag }}
         run: |
-          if [[ "${GITHUB_REF}" == refs/heads/* ]]; then
-            echo DOC_VERSION=${GITHUB_REF#refs/heads/} >> $GITHUB_ENV
+          if [[ -n "${INPUT_TAG}" ]]; then
```
evelikov, line 61: I think there used to be (is) some Github actions vulnerability where, one can inject commands into the tag name combined with the `${...}` expansions.

Will see if I can find a link for posterity sake. I think we should be fine here, since limited people can push tags and they're prefixed... Maybe we can update the expression to`v[0-9.]+`?

.github/workflows/docs.yml
```diff
@@ -34,17 +42,24 @@ jobs:
...
+            echo DOC_VERSION_COMMIT="$(git rev-parse HEAD)" >> $GITHUB_ENV
           elif [[ "${GITHUB_REF}" == refs/tags/* ]]; then
-            echo DOC_VERSION=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
+            TAG="${GITHUB_REF#refs/tags/}"
+            echo DOC_VERSION="${TAG%%.*}" >> $GITHUB_ENV
+            echo DOC_VERSION_COMMIT="${GITHUB_SHA}" >> $GITHUB_ENV
+          elif [[ "${GITHUB_REF}" == refs/heads/* ]]; then
+            echo DOC_VERSION="${GITHUB_REF#refs/heads/}" >> $GITHUB_ENV
+            echo DOC_VERSION_COMMIT="${GITHUB_SHA}" >> $GITHUB_ENV
           else
```
evelikov, line 72: With the new input method, would it make sense it check, warn or ban if one is re-generating existing versions?

Say a bug creeps-in, documentation for random version(s) is malformed and these malformed files overwrite the existing good ones. If only a few versions are affected, the odds of spotting the problem are minuscule.

.github/workflows/docs.yml
```diff
@@ -34,17 +42,24 @@ jobs:
...
+          if [[ -n "${INPUT_TAG}" ]]; then
+            TAG="${INPUT_TAG#refs/tags/}"
+            echo DOC_VERSION="${TAG%%.*}" >> $GITHUB_ENV
+            echo DOC_VERSION_COMMIT="$(git rev-parse HEAD)" >> $GITHUB_ENV
           elif [[ "${GITHUB_REF}" == refs/tags/* ]]; then
-            echo DOC_VERSION=${GITHUB_REF#refs/tags/} >> $GITHUB_ENV
+            TAG="${GITHUB_REF#refs/tags/}"
+            echo DOC_VERSION="${TAG%%.*}" >> $GITHUB_ENV
+            echo DOC_VERSION_COMMIT="${GITHUB_SHA}" >> $GITHUB_ENV
+          elif [[ "${GITHUB_REF}" == refs/heads/* ]]; then
```
evelikov, line 69: Silly question: why do we need the heads instance here?

evelikov (commented):
A while ago, I was thinking of a slightly different approach:
 - trigger on tag push
 - iterate all tags, generate the docs for each - older versions might need different build command
 - for each missing -> push to staging
 - for each re-generated -> flag the diff, allow to continue manually (gitlab has manual job triggers, don't recall if GH does) or via special env. variable
 - once staging is full -> release

evelikov:
I am loving the `docs` alias, but not sure if we can use it (or meson in general) to push docs for pre-meson kmod tags/versions.

.github/workflows/docs.yml
lucasdemarchi: yes

.github/workflows/docs.yml
lucasdemarchi: humn... this is the INPUT_TAG, for manual runs. I think your comment is about the other one just below? Anyway, manual runs and tag creation is very limited... but yes, I can limit to `v[0-9.]+`

.github/workflows/docs.yml
lucasdemarchi: but I do want to override. Minor versions are basically fixing stuff that should had been on that release.
Right now we are missing v34 in https://kmod-project.github.io/. I´d like v34.1 / v34.2 to simply show at https://kmod-project.github.io/v34/index.html, not at the specific minor versions.

.github/workflows/docs.yml
lucasdemarchi: master

lucasdemarchi:
> I am loving the `docs` alias, but not sure if we can use it (or meson in general) to push docs for pre-meson kmod tags/versions.

right... but I started this only to generate v34, not anything older than that. We can go back and do that, but then maybe we don't need it this github workflow...

.github/workflows/docs.yml
lucasdemarchi: btw, the way the docs update work is that the html files get committed to the other repo. So it's even possible to inspect the log later and find issues

evelikov (approved):
Don't have the means to quickly test and overall it looks good.

lucasdemarchi:
Applied, thanks

PR #457, opened 2026-08-11
Change my email address

lucasdemarchi:
Now I understand why sometimes I get email @gmail.com rather than @kernel.org. Switch to the latter everywhere.

Also add entry in .mailmap for a wrong email configuration I had in one machine for a while.


evelikov (approved):
Moar emails, moar 😈 

Joking aside, correct me if I'm wrong - kernel.org is an alias, right?

I'm travelling the next few days - will get to the other PRs shortly.

PR #458, opened 2026-08-11
WIP: Compat with clang-built kernel

lucasdemarchi:
If kernel was built with clang, building our test modules with gcc may not really work. Try to match the compiler and add CachyOS to CI since it has a clang-built kernel.


evelikov (commented):
First and third patch look great - feel free to merge regardless of the rest.

Wrt the second commit - BTF aka pahole - can we apply this workaround in all cases? It will allow us to remove any pahole references, which are workaround for distro packaging bugs AFAICT.

For the last commit - seems like Cachy people have added extra sandboxing to their pacman fork. If re-running the job might fix it - another (simpler?) option is to use Arch and only pull the Cachy kernel itself?

PR #459, opened 2026-08-30
libkmod: fix NULL deref in module_options_concat(); merge multiple softdep/weakdep stanzas

### LINKED ISSUE #33: When using multiple softdep for a module, only one of them have effects
Comments in the code suggest this might be intentional. But this is a surprising behavior. In addition to that, when there is already `softdep` commands in system's `modules.softdep`, user configured `softdep`s  in `/etc/modprobe.d` will just be ignored, making it effectively non-user configurable. I think this is undesirable.

prabhakarpujeri:
Four small fixes:

**1. libkmod: fix NULL dereference in module_options_concat()**

The `malloc()` return value is dereferenced unguarded; on OOM this NULL-derefs
in the `kmod_module_probe_insert_module()` (modprobe) path. The function's
contract treats NULL as "no options" (the sole caller passes the result on to
`kmod_module_insert_module()` / `module_do_install_commands()` which both
accept NULL), so the fix returns NULL on OOM and documents this contract in a
comment rather than inventing a new error path.

**2. libkmod: merge all matching softdep and weakdep config stanzas (fixes #33)**

`kmod_module_get_softdeps()` / `kmod_module_get_weakdeps()` broke out of the
config iteration after the first matching stanza, silently dropping deps from
additional `softdep` / `weakdep` lines for the same module. Now every matching
stanza is accumulated, mirroring `lookup_dep()` itself. The `multi_softdep`
testsuite case was marked `expected_fail` with a FIXME referencing #33; with
this fix it becomes a normal passing test, verified with `meson test`
(`TESTSUITE: PASSED: multi_softdep`, full suite 7/7, plus an ASan+UBSan build).

**3. depmod: add missing --dry-run to --help output**

`-n` is accepted as both `--show` and `--dry-run` (and documented as such in
depmod(8)) but the help text only listed `--show`.

**4. modprobe: merge duplicated -i lines in --help output**

The help output printed `-i` twice (for `--ignore-install` and
`--ignore-remove`); merged into one line matching modprobe(8)'s joint wording.

All commits build warning-free (`-Dwerror=true`) and the full testsuite passes
(7/7 in both the regular and sanitizer builds; patches also round-trip cleanly
via `git am` on pristine master).


libkmod/libkmod-module.c
```diff
@@ -857,17 +857,22 @@ static int module_do_install_commands(struct kmod_module *mod, const char *optio
...
-static char *module_options_concat(const char *opt, const char *xopt)
+/* Join @opt and @xopt into a newly allocated string, separated by a space. */
+static int module_options_concat(const char *opt, const char *xopt, char **result)
 {
 	// TODO: we might need to check if xopt overrides options on opt
 	size_t optlen = opt == NULL ? 0 : strlen(opt);
 	size_t xoptlen = xopt == NULL ? 0 : strlen(xopt);
 	char *r;
 
+	*result = NULL;
```
evelikov: Thanks for spotting and fixing this. Minor nit: let's set the output only on success.

Please also include the following in the commit message
```
Fixes: ddbda022 ("Add implementation of modprobe's insertion")
```

libkmod/libkmod-module.c
```diff
@@ -1270,16 +1270,11 @@ KMOD_EXPORT int kmod_module_get_softdeps(const struct kmod_module *mod,
...
+		/* accumulate all the softdep stanzas matching this module */
 		array = kmod_softdep_get_pre(l, &count);
-		*pre = lookup_dep(mod->ctx, array, count);
+		*pre = kmod_list_append_list(*pre, lookup_dep(mod->ctx, array, count));
 		array = kmod_softdep_get_post(l, &count);
-		*post = lookup_dep(mod->ctx, array, count);
-
-		/*
-		 * find only the first command, as modprobe from
-		 * module-init-tools does
```
evelikov: Let's split this in two - softdep and weakdep patches.

A couple of words in the commit message about the "find only first..." comment would be great. Where it's copied from, why it make sense there and not here - or anything along those lines.

libkmod/libkmod-module.c
```diff
@@ -1270,16 +1270,11 @@ KMOD_EXPORT int kmod_module_get_softdeps(const struct kmod_module *mod,
 		if (fnmatch(modname, mod->name, 0) != 0)
 			continue;
 
+		/* accumulate all the softdep stanzas matching this module */
 		array = kmod_softdep_get_pre(l, &count);
-		*pre = lookup_dep(mod->ctx, array, count);
+		*pre = kmod_list_append_list(*pre, lookup_dep(mod->ctx, array, count));
```
evelikov: `kmod_list_append_list()` can fail - let's handle that appropriately.

testsuite/test-libkmod.c
```diff
@@ -565,12 +565,7 @@ static int multi_softdep(void)
 	return 0;
 }
 
-/*
- * FIXME: This test indicates a bug in kmod. Once fixed, the expected_fail should be removed.
- * See https://github.com/kmod-project/kmod/issues/33 for more details.
- */
 DEFINE_TEST(multi_softdep, .description = "check if multiple softdep is supported",
-	    .expected_fail = true,
```
evelikov: Note to self: the "expected_fail" here should probably be "todo", since it's something that should be addressed - massive thanks for which 👍

tools/depmod.c
```diff
@@ -85,7 +85,8 @@ static void help(void)
 	       "\t-a, --all            Probe all modules\n"
 	       "\t-A, --quick          Only does the work if there's a new module\n"
 	       "\t-e, --errsyms        Report not supplied symbols\n"
-	       "\t-n, --show           Write the dependency file on stdout only\n"
+	       "\t-n, --show, --dry-run\n"
```
evelikov: Do you mind if we instead deprecate `--show` - similar to how we've done for `modprobe`? Let's make that a separate PR to avoid (potentially) stalling the rest of this PR.

libkmod/libkmod-module.c
evelikov: Cannot comment on the commit message, so a random note here:

The `Link` trailer should be `Closes` as per the contribution guide.

tools/modprobe.c
```diff
@@ -106,8 +106,8 @@ static void help(void)
 	       "\t                            module's refcount to become 0 so it can be\n"
 	       "\t                            removed (use together with -r)\n"
 	       "\t    --first-time            Fail if module already inserted or removed\n"
-	       "\t-i, --ignore-install        Ignore install commands\n"
-	       "\t-i, --ignore-remove         Ignore remove commands\n"
+	       "\t-i, --ignore-install, --ignore-remove\n"
+	       "\t                            Ignore install and remove commands\n"
```
evelikov: Yes please. Let's move this to a separate PR, alongside the previous commit.

evelikov (changes requested):
Massive thanks for the fixes. Left a few comments - overall nothing serious.

Out of curiosity:
Did you use AI/LLM for this (and your other PRs)? We don't have any formal policy but it's great to know if/to what extend it was used.

What brought your attention to these two bugs?

PR #461, opened 2026-09-02
libkmod: honor install commands on modules with softdeps

### LINKED ISSUE #446: Can't blacklist ksmbd, even with install /bin/false
I am unable to blacklist the ksmbd module, even though I added the lines below to /etc/modprobe.d using blacklist.conf:

```# Disable ksmbd
blacklist ksmbd
install ksmbd /bin/false
```

When I install the module using a dry run, I receive the following output:

```
modprobe --dry-run --verbose ksmbd
insmod /lib/modules/5.15.0-177-generic/kernel/fs/smbfs_common/cifs_arc4.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/drivers/infiniband/core/ib_core.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/drivers/infiniband/core/ib_cm.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/drivers/infiniband/core/iw_cm.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/drivers/infiniband/core/rdma_cm.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/crypto/crc32_generic.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/fs/ksmbd/ksmbd.ko 
```

However, if any module blocks, then it has to have the following output(e.g. nfsv4 module):

```
modprobe --dry-run --verbose nfsv4
insmod /lib/modules/5.15.0-177-generic/kernel/fs/netfs/netfs.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/fs/fscache/fscache.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/net/sunrpc/sunrpc.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/fs/nfs_common/grace.ko 
insmod /lib/modules/5.15.0-177-generic/kernel/fs/lockd/lockd.ko 
install /bin/false 
install /bin/false 
```

Does anyone know why it's happening because of soft dependencies or anything else? I also found a similar issue here:https://forums.linuxmint.com/viewtopic.php?t=390505&sid=0aa590b8ebfea9ecdca5dc57d8225747he 

FYI: I am blocking the module due to CVEs: CVE-2025-68263,CVE-2026-43185

prabhakarpujeri:
Fixes #446.

Since 89e9248 ("modprobe: obey blacklist and softdep pre"), `__kmod_module_fill_softdep()` marks every module that has any softdep entry with `ignorecmd`, and `kmod_module_probe_insert_module()` then silently skips the module's configured `install` command and falls back to `insmod`.

Consequence: the documented block formula

```
blacklist ksmbd
install ksmbd /bin/false
```

has no effect for a module that ships a softdep (Gentoo confirmed `blacklist+install` is silently ignored) – a dry run inserts the real module file, and as the reporter notes, "even something as simple as a softdep kills any install command".

### Changes

- `libkmod`: only set `ignorecmd` when **no** install command is configured (`kmod_module_get_install_commands(mod) == NULL`). An explicit `install` directive is user configuration and must win over softdep metadata; modules without one keep the existing behavior, and `--ignore-install` is unaffected.
- `testsuite`: new fixture `test-modprobe/install-cmd-softdep` – module with a `pre` softdep plus `install mod-foo /bin/true`. The dry run must print `insmod …/mod-foo-a.ko` followed by `install /bin/true`; without the fix it prints `insmod …/mod-foo.ko` instead. Verified red on master, green with this change; full testsuite passes.

### Reproduction (on master)

```
# fake rootfs: mod-foo.ko + mod-foo-a.ko,
# modules.softdep: softdep mod-foo pre: mod-foo-a
# /run modprobe.d: blacklist mod-foo / install mod-foo /bin/false
$ modprobe -n -v mod-foo
insmod <rootfs>/lib/modules/$(uname -r)/kernel/mod-foo-a.ko
insmod <rootfs>/lib/modules/$(uname -r)/kernel/mod-foo.ko   ← installs despite /bin/false
```

after this change, the second line is `install /bin/false`.


libkmod/libkmod-module.c
```diff
@@ -912,7 +912,15 @@ static int __kmod_module_fill_softdep(struct kmod_module *mod, struct kmod_list
...
 	*list = l;
-	mod->ignorecmd = (pre != NULL || post != NULL);
+	/*
+	 * Softdep entries suppress this module's own install command in the
+	 * probe list.  Keep that only as long as no install command is
+	 * configured: explicit configuration has always had precedence and
+	 * losing it silently breaks "install foo /bin/false" blocks for
+	 * modules that happen to ship softdeps.
+	 */
+	mod->ignorecmd = (pre != NULL || post != NULL) &&
```
evelikov: Looking at the commit which introduced `ignorecmd` and it's not immediately obvious why we're setting it for softdeps. @lucasdemarchi realise it's been ages since you wrote 89e92487bc53cbb41c7d401d9a580b3009b2f711 but any context would be appreciated.

In addition, `ignorecmd` is being ignored (horrible pun intended) for weakdeps.

evelikov (changes requested):
When the issue got reported a more comprehensive solution came to mind:

 - make the kernel's `MODULE_SOFTDEP("pr3: non-module")` more robust
 - fix in-tree module issues

^^ are not (too) relevant here, for kmod so feel free to ignore.

 - emit kmod warning
 - fix the underlying kmod bug

Can we get something like this?

Small nits aside (Link -> Closes && Fixes tag), let's tweak the commit message to cover "why" instead of "how" and drop the less interesting "blacklist" & "Add a test-modprobe..." parts.

PR #463, opened 2026-09-09
move hash creation into kmod_module_new

### LINKED ISSUE #441: Move `kmod_module_new_from_alias` logic into `kmod_module_new`
I came across the following TODO:
https://github.com/kmod-project/kmod/blob/e3d6b397fdad041484b966669916d1f0c74eeda6/libkmod/libkmod-module.c#L236-L237

after inspecting the usages of this function, I think that it is possible to move the logic of `kmod_module_new_from_alias`
https://github.com/kmod-project/kmod/blob/e3d6b397fdad041484b966669916d1f0c74eeda6/libkmod/libkmod-module.c#L301-L302
into  `kmod_module_new` and updating the callers to call `kmod_module_new`

I am waiting for a response before starting to implement the changes.

ysfgalip:
Closes #441

I moved the hash creation into the `kmod_module_new` as mentioned in TODO.

The callers still call the `kmod_module_new_from_alias` `kmod_module_new_from_name` functions, I am not sure if it should stay this way. Callers can be updated to call the `kmod_module_new` instead.


libkmod/libkmod-module.c
```diff
@@ -234,44 +234,47 @@ bool kmod_module_is_builtin(struct kmod_module *mod)
  *
  * @key is "name\alias" or "name" (in which case alias == NULL)
```
evelikov: Lets drop this line.

libkmod/libkmod-module.c
```diff
@@ -234,44 +234,47 @@ bool kmod_module_is_builtin(struct kmod_module *mod)
...
-		return 0;
-	}
-
 	if (alias == NULL)
 		keylen = namelen;
 	else
 		keylen = namelen + aliaslen + 1;
 
+	if (keylen + 1 > PATH_MAX)
+		return -ENAMETOOLONG;
```
evelikov: Why do we need this check?

libkmod/libkmod-module.c
```diff
@@ -280,6 +283,7 @@ static int kmod_module_new(struct kmod_ctx *ctx, const char *key,
 		free(m);
 		return err;
 	}
+	m->ctx = kmod_ref(ctx);
```
evelikov: By moving the refcounting we avoid the memory leak in the error path just above - likely spotted.

Let's keep that as separate commit though.

libkmod/libkmod-module.c
```diff
@@ -296,24 +300,19 @@ KMOD_EXPORT int kmod_module_new_from_name(struct kmod_ctx *ctx, const char *name
...
 }
 
 int kmod_module_new_from_alias(struct kmod_ctx *ctx, const char *alias, const char *name,
 			       struct kmod_module **mod)
 {
-	char key[PATH_MAX];
 	size_t namelen = strlen(name);
 	size_t aliaslen = strlen(alias);
 
 	if (namelen + aliaslen + 2 > sizeof(key))
```
evelikov: The `key` is declaration got removed, yet `sizeof(key)` remained. Who did this even build?

evelikov (changes requested):
Thanks for the PR.

Please include the Closes trailed in the commit itself, alongside a brief commit message.

Yes - at a glance we can inline ` kmod_module_new_from_alias`. I don't think we can do much about `kmod_module_new_from_name` - it's public API that should not be broken.

libkmod/libkmod-module.c
ysfgalip: In `kmod_module_new_from_alias` the size of the buffer holding the key was `PATH_MAX`, since the key creation is moved to `kmod_module_new`, I thought it is a good idea to check the key length in this function

libkmod/libkmod-module.c
ysfgalip: You are right, I missed it and didn't test it after doing this change, It must be changed to `PATH_MAX` instead.

ysfgalip:
> Thanks for the PR.
> 
> Please include the Closes trailed in the commit itself, alongside a brief commit message.
> 
> Yes - at a glance we can inline ` kmod_module_new_from_alias`. I don't think we can do much about `kmod_module_new_from_name` - it's public API that should not be broken.

okay, then I will remove the `kmod_module_new_from_alias` and call the `kmod_module_new` instead

libkmod/libkmod-module.c
evelikov: `kmod_module_new_from_alias` had `PATH_MAX` sized variable, so we were checking for potential overflow prior to copying. I don't think we need such a check here.

libkmod/libkmod-module.c
evelikov: As per above - the check is meant to avoid overflow(s). Don't think we need it any more.

libkmod/libkmod-module.c
evelikov: There is a pre-existing corner-case, if you'd want to add an extra patch to this PR.

We don't check for overflows in `keylen` + associated `malloc`. Search for `uaddsz_overflow` et al for in-tree examples how to fix.

